PR #85, opened 2015-01-26
Update mod_ls.c

sylphid8:
don't read outside the options array

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -1572,7 +1572,7 @@ static void parse_list_opts(char **opt, int *glob_flags, int handle_plus_opts) {
       /* Options are found; skip past the leading whitespace. */
       *opt = ptr;
 
-    } else if (*(*opt + 1) == ' ') {
+    } else if (**opt && (*(*opt + 1) == ' ')) {
       /* If the next character is a blank space, advance just one character. */
       (*opt)++;
       break;
@@ -1686,7 +1686,7 @@ static void parse_list_opts(char **opt, int *glob_flags, int handle_plus_opts) {
       /* Options are found; skip past the leading whitespace. */
       *opt = ptr;
 
-    } else if (*(*opt + 1) == ' ') {
+    } else if (**opt && (*(*opt + 1) == ' ')) {
       /* If the next character is a blank space, advance just one character. */
       (*opt)++;
       break;


sylphid8:
@Castaglia anything holding this up from being merged in? This patch corrects for an occasional leak into a neighboring string resulting in erroring on a LIST with "No such file or directory"

Castaglia:
Could you provide an example LIST command which runs afoul of this bug, to be use in e.g. a regression test?  Thanks!

sylphid8:
issuing a "LIST -a" command twice can trigger this bug on the second occurrence, however, this does not happen for every user/directory. I do not have a way of recreating this bug, but it does seem to be reproducible once the right combination is stumbled upon

Castaglia:
I was able to reproduce a "450 no such file or directory" response by including 2 trailing spaces in the command, after the options, i.e. "LIST -a  ".  I think the bug might be in the handling of those spaces as real paths; does that sound right?

The mod_ls module doesn't maintain state across commands, so your comment about "issuing a LIST twice can trigger this bug" makes me wonder whether something in the client you're using might be somehow appending these trailing spaces?

Castaglia:
Hmm.  The trailing spaces might still be necessary; it should be possible to use LIST on a path that starts with whitespace; such files are allowed on Unix.  Given this, a "450 no such file or directory" response might be correct, albeit unexpected, in the case of trailing whitespace.

Given the changes you point out, though, I'm not sure that that is the issue you're encountering.  If you can reproduce this fairly easily, getting the FTP commands/responses verbatim (e.g. using "ftp -d ...") would be very, very useful.

sylphid8:
bash-3.2# ftp -d localhost
Connected to localhost (127.0.0.1).
220 ::ffff:127.0.0.1 FTP server ready
---> USER ccp0w
331 Password required for ccp0w
---> PASS XXXX
230 User ccp0w logged in
---> SYST
215 UNIX Type: L8
Remote system type is UNIX.
Using binary mode to transfer files.
ftp> ls -a
---> PASV
227 Entering Passive Mode (127,0,0,1,195,228).
---> LIST -a
drwxr-x--x  62 ccp0w    ccp0wg       8192 Jan 12 11:16 .
drwxr-x--x  62 ccp0w    ccp0wg       8192 Jan 12 11:16 ..
-rw-------   1 ccp0w    ccp0wg       5674 Jan 24 12:16 .bash_history
drwx------   2 ccp0w    ccp0wg       4096 Mar 10  2011 .gnome2
-rw-------   1 ccp0w    ccp0wg         63 Mar 20  2014 .lesshst
<snip>
226 Transfer complete
ftp> ls -a
---> PASV
227 Entering Passive Mode (127,0,0,1,205,21).
---> LIST -a
450      -a: No such file or directory

sylphid8:
ftp_debug

Castaglia:
Got it.  I was finally able to reproduce this on my Linux machine (but not on my Mac).  Interesting.

I've create a different branch, with your fix and with accompanying regression tests:

  https://github.com/proftpd/proftpd/tree/ls-option-parsing

Unless there're any objections, I'll get this merged to master (and backported to the 1.3.5 branch) later today.

Thanks!!

sylphid8:
Yea, triggering the bug kinda seems to be luck of the draw as to if the neighboring memory address is null or not. I don't believe that the regression test is 100% reliable, but then again, I can't come up with anything that would be. That being said, I have no objections.

Castaglia:
Merged to master, and backported to 1.3.5 branch.  Thanks!

PR #99, opened 2015-03-31
extend tracked file size only if current write request crossed assumed file end

bkmit:


Castaglia:
Can you provide a description of why this is useful?  What's your use case?  What's the desired functionality, and why?  Thanks!

bkmit:
1. Possible negative value indicating write error wrongly used for size calculating.
2. SFTP protocol allow writing in random region of existing file not always expanding it. This patch allow properly calculate file size in this case. This calculated value is used for enforcing file size limit.

mfuterko:
I've extended this pull request with additional flags for READ/WRITE request to indicate whether the dir_chack() functions were called on the file handle. There are more similar places for optimisation.

contrib/mod_sftp/fxp.c
```diff
@@ -11227,7 +11237,8 @@ static int fxp_handle_write(struct fxp_packet *fxp) {
    * want to track the on-disk size for enforcing limits such as
    * MaxStoreFileSize.
    */
-  fxh->fh_st->st_size += res;
+  if (res > 0 && (offset + res) > fxh->fh_st->st_size)
```
Castaglia: This logic has been added to master; thanks!

PR #113, opened 2015-04-08
Changes for support APPEND to a new non existing file

alfonso-escribano:
Changes for let clientes use append command to upload a new file.
I doesn't change xfer_type in patch because it is changed later in get_hidden_store_path function


Castaglia:
I believe that this functionality exists, as part of handling [Bug#4144](http://bugs.proftpd.org/show_bug.cgi?id=4144); see https://github.com/proftpd/proftpd/commit/cbd4a9d05863df0c0a73bf927d031c433d8a44a3.

PR #117, opened 2015-04-16
Update and rename README to README.md

gildasiojunior:
1. Use md extention because use markdown sintaxe;
2. Update in some sections with markdown "tags"


Castaglia:
Nice!  Very much appreciated!

PR #133, opened 2015-06-05
Fix small copy/paste leftower

bkmit:
Look like this error do not influence program functions.
But better to fix it until will do.


Castaglia:
Good catch; thanks!

PR #135, opened 2015-06-09
Fix setting of TZ variable before chroot, to fix bad timestamps in logs

jmaggard10:


src/auth.c
```diff
@@ -1716,13 +1716,16 @@ int pr_auth_chroot(const char *path) {
 
   tz = pr_env_get(session.pool, "TZ"); 
   if (tz == NULL) {
+    char tzbuf[32];
+    long off = (timezone / 60 / 60) - daylight;
+    snprintf(tzbuf, sizeof(tzbuf), "%s%ld", daylight ? tzname[1] : tzname[0], off);
```
Castaglia: This formatting is making assumptions about the format of the TZ environment variable which are platform-specific.  What is the problematic behavior you're seeing, that makes this patch necessary in your environment?

jmaggard10: I'm trying to address the issue you described at http://www.proftpd.org/docs/howto/Timestamps.html, but without requiring the user to do any special configuration.

I figured the existing GLIBC ifdef's would be enough to weed out platforms that wouldn't work with this code, but certainly that assumption could be invalid.

Castaglia: Ah, OK.  What do you think of having automatically setting the TZ value, if not found in the environment, to ":/etc/localtime"?  Per the tzset(3) and tzfile(5) man pages, that is the default assumption made by functions like ctime(3), if no TZ environment variable is found.

jmaggard10: Huh... that does appear to be working for me now.  I tried it a year or so ago, and it didn't work reliably, presumably because the chroot directory does not have a /etc/localtime file.  But I am running a newer glibc version now, so maybe something has changed on that side?

Castaglia: Well, since this code path would be executed before the chroot(2) call, the /etc/localtime file would need to be accessible the first time any ctime(3) or related functions were called.  Which happens, I think, because of the pr_log_debug() function call, right after the setting of the TZ environment variable.  The logging routines want to add a timestamp to the generated log message, so they call e.g. ctime(3), which in turn looks up TZ...and its loaded.

jmaggard10: Well, good enough for me. :-)

Castaglia: If you update this PR to use ":/etc/localtime", and maybe throw in a call to tzset(3) following the setting of the TZ environment variable, I'll merge it, and we'll go from there.  Sound good?

Castaglia:
I've drafted up a [different PR](https://github.com/proftpd/proftpd/pull/285), per this conversation, which I'll be merging.  Your patch/thoughts on this particular topic are much appreciated, and many other users, present and future, will enjoy this (I hope).  Thanks!

Castaglia:
Since I've merged the other PR, I'm going to close this one; please re-open if it doesn't work for you.

PR #152, opened 2015-08-26
Bug#4202 - MLSD lines not properly terminated with CRLF.

Castaglia:


modules/mod_facts.c
```diff
@@ -212,75 +215,70 @@ static size_t facts_mlinfo_fmt(struct mlinfo *info, char *buf, size_t bufsz) {
...
 
-  /* MLST entries are not sent via pr_data_xfer(), and thus we do not need
-   * to include an LF at the end; it is appended by pr_response_send_raw().
-   * But MLSD entries DO need the trailing LF, so that it can be converted
-   * into a CRLF sequence by pr_data_xfer().
-   */
-  if (strcmp(session.curr_cmd, C_MLSD) == 0) {
-    snprintf(ptr, bufsz - buflen, " %s\n", info->path);
+  if (flags & FACTS_MLINFO_FL_APPEND_CRLF) {
+    len = snprintf(ptr, bufsz - buflen, " %s\r\n", info->path);
```
Castaglia: For future reviewers: this change here is the key.  Notice how the previous code path was only using LF, instead of the necessary CRLF.  The previous code path worked, because it relied on code in pr_data_xfer() to convert that LF into a CRLF.  But later changes toggled **off** the ASCII flag, prior to calling pr_data_xfer(), which meant that pr_data_xfer() would **not** perform ASCII conversion.  And that is the cause of this regression.

PR #155, opened 2015-09-16
RNTO across mounts

lazy404:
Hi,

RNTO doesn't work across mounts,

I did as instructed in comments:
"This could be fixed to work later, e.g. using code from the mod_copy module."

After some basic testing it work's for me with current master head.


Castaglia:
One of the considerations here is that the directory being moved could be quite large.  This could mean that the requesting client could be left waiting for a while, waiting for the response -- and possibly eventually timing out, either due to client- or server-side timeouts (see [Bug#4263](http://bugs.proftpd.org/show_bug.cgi?id=4263).  A similar fix, as used for that bug, might be needed in this use case as well.

PR #157, opened 2015-09-24
OSX: Close a reasonable number of FDs on fork()

klamontagne:
nfiles ends up being 2^63-1 under OSX:
RLIMIT_OFILE is undefined
RLIMIT_NOFILE is 0
rlim.rlim_max is 2^63-1

However, by default, rlim_cur is 256.

### DIFF
--- a/contrib/mod_exec.c
+++ b/contrib/mod_exec.c
@@ -317,7 +317,10 @@ static void exec_prepare_fds(int stdin_fd, int stdout_fd, int stderr_fd) {
    * First, use getrlimit() to obtain the maximum number of open files
    * for this process -- then close that number.
    */
-#if defined(RLIMIT_NOFILE) || defined(RLIMIT_OFILE)
+#if defined(__APPLE__)
+   getrlimit(RLIMIT_NOFILE, &rlim);
+   nfiles = rlim.rlim_cur;
+#elif defined(RLIMIT_NOFILE) || defined(RLIMIT_OFILE)
 # if defined(RLIMIT_NOFILE)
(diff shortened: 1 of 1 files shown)


contrib/mod_exec.c
```diff
@@ -317,7 +317,10 @@ static void exec_prepare_fds(int stdin_fd, int stdout_fd, int stderr_fd) {
    * First, use getrlimit() to obtain the maximum number of open files
    * for this process -- then close that number.
    */
-#if defined(RLIMIT_NOFILE) || defined(RLIMIT_OFILE)
+#if defined(__APPLE__)
```
Castaglia: Rather than using a platform-specific macro like `__APPLE__`, I'd prefer to use either a) something feature-based, or b) something that the proftpd build system already provides/sets, e.g. `DARWIN9`, `DARWIN10`, `DARWIN11`, etc.

Also, is this change fixing a specific bug, or is it making something better?  It'd hard to tell from just the description.  I.e. should `RLIMIT_OFILE` be preferred over `RLIMIT_NOFILE`, but only on some platforms, or...?

klamontagne:
On my machine, the loop that closes files runs for so long that FTP commands time out. The "not Kosher" check at line 353 passes because the value _rlim_max_ does not overflow a long.

Castaglia:
Hmm.  What about this approach?  See:

  https://github.com/proftpd/proftpd/pull/158

klamontagne:
It's fine and fixes my issue just as well.

Castaglia:
OK, cool.  I've merged https://github.com/proftpd/proftpd/pull/158 now, and will thus close out this PR.  Thanks!

PR #163, opened 2015-10-10
Remove HAVE_TIMER_T macro and timer_t detection mechanism

bverschueren:
When compiling on AIX6.1, following error is raised:

```
In file included from /usr/include/sys/thread.h:41:0,
                 from /usr/include/sys/ptrace.h:28,
                 from /usr/include/sys/proc.h:42,
                 from /usr/include/sys/aacct.h:22,
                 from /usr/include/sys/vfs.h:34,
                 from fsio.c:39:
/usr/include/sys/timer.h:292:2: error: unknown type name 'p_timer_t'
  timer_t            timerid;       /* timer identifier               */
```

which can be tracked down to `include/conf.h`:

```
/* Solaris 2.5 seems to already have a typedef for 'timer_t', so
 * #define timer_t to something else as a workaround.  However, on AIX,
 * we do NOT want to do this.
 */
#if defined(HAVE_TIMER_T) && !defined(AIX7)
# define timer_t p_timer_t
#endif
```

As no `timer_t` is typedef'd in the ProFTPD source, I presume this workaround has become obsolete.
The last notion of `timer_t` has been dropped since bae6cd2330117f4bedeb1f07fe11bcc4619537ff.

### DIFF
--- a/acconfig.h
+++ b/acconfig.h
@@ -60,9 +60,6 @@
 /* Define if you have the <syslog.h> header file. */
 #undef HAVE_SYSLOG_H
 
-/* Define if you already have a typedef for timer_t */
-#undef HAVE_TIMER_T
-
 /* Define if your struct utmp has ut_host */
 #undef HAVE_UT_UT_HOST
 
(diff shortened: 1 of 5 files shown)


Castaglia:
Merged.  Thanks!

PR #173, opened 2015-12-01
Add EmulateWindows mode to mod_ls

asbaker:
We had an issue with some camera hardware not being able to successfully understand the linux based file and directory listings. This patch includes a directive, `EmulateWindows`, which sends back windows'ish file and directory listings to all clients.

Example:

```
ftp> ls
229 Entering Extended Passive Mode (|||57644|)
11-30-15  07:29PM       <DIR>      102 a_folder
11-30-15  07:29PM                    0 file
226 Transfer complete
```

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -633,6 +633,27 @@ static int listfile(cmd_rec *cmd, pool *p, const char *resp_code,
       t = pr_localtime(p, (time_t *) &sort_time);
     }
 
+    unsigned char *emulate_windows = NULL;
+    emulate_windows = get_param_ptr(TOPLEVEL_CONF, "EmulateWindows", FALSE);
+    if (emulate_windows != NULL && *emulate_windows == TRUE) {
+      char time_str[18], dir_str[13];
+      char nameline[(PR_TUNABLE_PATH_MAX * 2) + 128] = {'\0'};
+      memset(time_str, 0, sizeof(time_str));
+      memset(dir_str, 0, sizeof(dir_str));
+
+      strftime(time_str, sizeof(time_str), "%m-%d-%y  %I:%M%p", t);
+
+      if (S_ISDIR(st.st_mode)) {
+        strcpy(dir_str,  "<DIR>");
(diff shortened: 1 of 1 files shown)


Castaglia:
I like the idea of this patch, although I might want to rename the directive to be `ListStyle`, _e.g._:

    ListStyle Unix|MS-DOS

To be similar to the options available on IIS.  Sound good?

modules/mod_ls.c
```diff
@@ -633,6 +633,27 @@ static int listfile(cmd_rec *cmd, pool *p, const char *resp_code,
       t = pr_localtime(p, (time_t *) &sort_time);
     }
 
+    unsigned char *emulate_windows = NULL;
+    emulate_windows = get_param_ptr(TOPLEVEL_CONF, "EmulateWindows", FALSE);
```
Castaglia: FYI, I strive to ensure that the ProFTPD code compiles cleanly as ANSI code; this means (among other things) that all variable declarations must be at the start of the code block, rather than in the middle.  So I might adjust this particular coding for that; just a head's up.

asbaker:
@Castaglia No problem, thanks for the feedback! I like the ListStyle idea.

I can make these changes. I found an inconsistency with how we're displaying directory listings, so I can incorporate your changes in with that fix. I'll repush and drop a comment once that is complete, probably sometime Monday.

Castaglia:
@asbaker Any chance you did your mentioned update for this PR?  I think this would be a good addition for the 1.3.7 release cycle...

asbaker:
@Castaglia I do have something in process, but haven't had the opportunity to wrap it up. What's the timing look like on 1.3.7?

Castaglia:
At long last, I have implemented this a little differently; see https://github.com/proftpd/proftpd/pull/1110.

I needed this directive in order to support my work on https://github.com/Castaglia/proftpd-mod_proxy/issues/21.

Castaglia:
I'm going to close this PR, in favor the other one.  Thanks for your code for this, it really helped!

PR #180, opened 2015-12-11
Fix the 'make dist' target

Castaglia:


pghmcfc:
This change loses the "-Y" option from "makedepend", so doing "make dist" results in a bunch of system-specific include files getting added to the dependencies and then breaking builds on other downstream systems. Changing the default MAKEDEPEND value from "makedepend" to "makedepend -Y" may fix this. I had some success with 'make dist MAKEDEPEND="makedepend -Y"'.

Castaglia:
OK, I'll restore the `-Y`.  On the Mac OSX machine on which I've been developing, `gcc` (a la `clang`) doesn't understand the `-Y`.  I know that OpenSSL makes this work somehow; I'll see what they do.

PR #182, opened 2015-12-15
Fix log message on connection refused by MaxHostsPerUser limits

tkyf:
The log message on connection refused by MaxHostsPerUser limits is `Connection refused (MaxHostsPerHost ...`.
I presume for this message reads `Connection refused (MaxHostsPerUser ...` instead according to other messages.


Castaglia:
Thanks!

PR #183, opened 2015-12-16
Bug#4200 - Support TLS client configuration for SQL servers.

Castaglia:


Castaglia:
Note that this branch should be tested with `mod_sql_tds`; if necessary, a backward compatibility macro in `include/compat.h` should be added.

PR #226, opened 2016-03-08
Bug#4230: TLSDHParamFile directive appears ignored because unexpected…

Castaglia:
… DH is

chosen.

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -2755,7 +2755,7 @@ static int tls_ctrl_renegotiate_cb(CALLBACK_FRAME) {
 static DH *tls_dh_cb(SSL *ssl, int is_export, int keylen) {
   DH *dh = NULL;
   EVP_PKEY *pkey;
-  int pkeylen = 0;
+  int pkeylen = 0, use_pkeylen = FALSE;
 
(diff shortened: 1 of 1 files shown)


contrib/mod_tls.c
```diff
@@ -2844,6 +2882,10 @@ static DH *tls_dh_cb(SSL *ssl, int is_export, int keylen) {
     }
   }
 
+  if (use_pkeylen) {
+    keylen = pkeylen;
```
Castaglia: This addresses one part of the issue, where for a 2048-bit server cert, we _should_ have selected the 2048-bit DH, but instead we only used the unadjusted 1024-bit length provided by OpenSSL.

contrib/mod_tls.c
```diff
@@ -2818,6 +2820,22 @@ static DH *tls_dh_cb(SSL *ssl, int is_export, int keylen) {
           "found matching DH parameter for key length %d", keylen);
         return dhs[i];
       }
+
+      /* Try to find the next "best" DH to use, where "best" means
+       * the smallest DH that is larger than the necessary keylen.
+       */
```
Castaglia: This logic should hopefully address the bug, where the principle of least surprise was violated because a DH (4096 bits), larger than the configured server cert (of 2048 bits), was **not** selected.

Issue #230, opened 2016-03-13
Enforce minimum SSH key length


Castaglia:
Some security-conscious sites might wish to configure `mod_sftp` such that the keys used for authentication, _e.g._ user publickey authentication, are longer than some minimum key length.
See this [StackExchange question](http://serverfault.com/q/762652/253490) on the topic.

Castaglia:
OpenSSH currently enforces hardcoded minimum of 768 bits, but only for RSA keys; it does not impose any minimum length on DSA or ECDSA keys.

Castaglia:
I'm thinking to implement something like:

    SFTPKeyLimits RSAMinimumSize 768

Castaglia:
Also filed this in the project Bugzilla instance:

  http://bugs.proftpd.org/show_bug.cgi?id=4233

Castaglia:
Fixed in master.

PR #239, opened 2016-03-22
Bug#4233: Support enforcing minimum key lengths for SFTP/SCP

Castaglia:


Castaglia:
Note that this also addresses Issue #230 .

PR #272, opened 2016-04-23
add pr_log_stacktrace api function to trace in debug the stacktrace

mcarbonneaux:


Castaglia:
This shouldn't really be necessary; the existing stacktrace-handling code in the logging API kicks in whenever there's a segfault.  Why would you want this stacktrace handling code to be publicly available like this?  What's the use case?

mcarbonneaux:
is to abel to add tacktrace in the code when necesseary, to debug.

2 modules have this in her code (sftp/snmp)... seem to be util to have that has api callable.

i've used to anderstand what call my code to anderstand how proftpd work, and if somme code are called and by who...

include/log.h
```diff
@@ -99,6 +99,8 @@
 # endif
 #endif
 
+void pr_log_stacktrace(const char*moduleversion, const char*function);
```
Castaglia: I like the general idea of this PR.  However, I'm wondering if a slightly different interface might work better.  Consider, for example, a signature of:

    void pr_log_stacktrace(int fd, const char *label, const char *file, int lineno);

That way, the stacktrace can be logged to an arbitrary file descriptor, not just via ProFTPD's logging routines.  Also, instead of providing the function name where an issue occurred, you provide the actual name of the file, and line number using the `__FILE__` and `__LINE__` preprocessor macros, _e.g._:

    /* Log the stacktrace */
    pr_log_stacktrace(log_fd, MOD_EXAMPLE_VERSION, __FILE__, __LINE__);

Or, even better:

    void pr_log_stacktrace(int fd, module *m, const char *file, int lineno);

Given a `module` pointer, the function could automatically create the "mod_example/version" string.

src/log.c
```diff
@@ -56,6 +56,40 @@ static int fd_set_block(int fd) {
...
 
+void pr_log_stacktrace(const char*moduleversion, const char*function) {
+#if defined(HAVE_EXECINFO_H) && \
+    defined(HAVE_BACKTRACE) && \
+    defined(HAVE_BACKTRACE_SYMBOLS)
+  void *trace[PR_TUNABLE_CALLER_DEPTH];
+  char **strings;
+  int tracesz;
+
+  (void) pr_log_debug(DEBUG4, "%s: -----BEGIN STACK TRACE in %s -----",moduleversion, function);
```
Castaglia: I would probably change this implementation slightly, such that the debug level (if used/provided) could be configured by the caller.  Although one might argue that stacktraces are _always_ important, and thus perhaps `pr_log_pri(PR_LOG_NOTICE, ...)` might be a better fit.

include/log.h
mcarbonneaux: i modified in master the addition of module *m to auto generate mod/version string, file and lineno.

but fd is more complexe change, i'm not sure how to do properly and i think this type of debug backtrace are more needed in debug trace of proftpd than any other file...

src/log.c
mcarbonneaux: i've added an argument to send the debug level in the function.

include/log.h
```diff
@@ -83,6 +83,10 @@
 # include <login.h>
 #endif
 
+#ifdef HAVE_EXECINFO_H
+# include <execinfo.h>
+#endif
```
Castaglia: You probably don't need this `#include`, since the function signatures don't use any of the data types from this header.

src/log.c
```diff
@@ -56,6 +56,40 @@ static int fd_set_block(int fd) {
...
+  if (tracesz < 0) {
+    (void) pr_log_debug(debuglevel, "%s/%s: backtrace(3) error: %s", module->name, module->module_version, strerror(errno));
+  }
+
+  strings = backtrace_symbols(trace, tracesz);
+  if (strings != NULL) {
+    register unsigned int i;
+
+    for (i = 1; i < tracesz; i++) {
+      (void) pr_log_debug(debuglevel, "%s/%s: [%u] %s", module->name, module->module_version, i-1, strings[i]);
```
Castaglia: I'd suggest doing something like this, _before_ the `for` loop:

    const char *module_name = NULL, *module_version = NULL;

    if (module != NULL) {
      module_name = module->name;
      module_version = module->module_version;
    }

Because there _will_ be cases where the callers do not have a `module *` to provide (_.e.g._ in the core API functions).

Then, in the `for` loop, you might then use:

    if (module != NULL) {
      pr_log_debug(debuglevel, "%s/%s: [%u] %s", module_name, module_version, i-1, strings[i]);

    } else {
      pr_log_debug(debuglevel, "core: [%u] %s"", i-1, strings[i]);
    }

Without these defensive checks, this code could segfault on a null pointer dereference.

Castaglia:
I like the idea of this pull request.  However, the discussion/comments about configurable log level, and whether to use `pr_log_pri()` or to provide a specific log file descriptor, _etc_ have me wondering whether a slightly different approach might work better.

Consider, for example, a function like this:

    const array_header *pr_log_get_stacktrace(pool *p);

which would return an `array_header *`, containing an array of strings, one per frame.  The caller, then, could log that array of strings however it wished: `pr_log_pri()`, or to a specific descriptor, or whatever.  The _caller_ would be responsible for using `__FILE__` and `__LINE__` (or not).  Thoughts?

Castaglia:
I'm going to merge this in, then make some modifications.  Thanks!

Issue #279, opened 2016-05-06
Make PBSZ optional


gczuczy:
We've found a broken client, which doesn't do PBSZ before PROT, and by RFC2228 proftpd is refusing the client because of this. I would like to ask to make PBSZ optional, because even in the source it's forced to 0, and to the best of my knowledge, it doesn't raise any security issues as well.

As per email discussion with castaglia, I've created a patch, and will create a pull request for it.

The path introduces a new TLSOptions option named "RequirePBSZ", and PBSZ will only be required if that flag is present. If the flag is not specified, then PBSZ is optional.

Castaglia:
Thanks for the pull request!  I may change the `TLSOption` name to something else later, _e.g._ `TLSOption RFCConformance` or somesuch, some name that might be slightly more human-readable, since some users/admins may not be aware of the `PBSZ` and its role, specifically.  But the functionality will stay the same.

gczuczy:
You are most welcome, I've fixed the things you've pointed out in the pull request, please take a look at it again.

I've chosen "RequirePBSZ", because it's specifically about that command, and the error message also mentions this command in reply to PROT, when forgotten. There are a rather large number of relevant RFCs, and even RFC2228 has multiple requirements, that's one of the reasons why I chosen a name for the option with "RFC" in it. Personally I think, that might be misleading and confusing on the long run.

Castaglia:
In the case of FTP over SSL/TLS, there's only one particularly relevant RFC: [RFC 4127](https://tools.ietf.org/html/rfc4217).  However, your point is taken.

Another approach might be to not provide such a `TLSOption` at all, and simply always treat `PBSZ` as optional.

gczuczy:
Honestly, I don't really see any reason why couldn't be optional by default, except if some sysadmin is feeling to start a holy crusade against broken clients, and wants to teach the world a lesson. Apart from that, I don't think it would hurt anyone.

Would you like me to change the pull request to make this the default behaviour?

Castaglia:
Yes, please.  It's easy enough to add such an option in, later, if/when we find a use case.

gczuczy:
Done, please review. Instead of removing the error return, i've just disabled the codepath with a macro, if that's fine. Might come handy if there's a usecase.

Castaglia:
PR merged to master; thanks!

PR #280, opened 2016-05-06
[mod_tls] Introduced TLSOptions option RequirePBSZ, by default it's o…

gczuczy:
…ptional now

RFC2228 requires clients to do PBSZ before PROT, however some broken clients ignore
this requirement. Alone it's no harm, because it's forced to be 0, so PBSZ can be made
optional.

Issue: #279

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -485,6 +485,7 @@ static unsigned char *tls_authenticated = NULL;
 #define TLS_OPT_VERIFY_CERT_CN				0x0800
 #define TLS_OPT_NO_AUTO_ECDH				0x1000
 #define TLS_OPT_ALLOW_WEAK_DH				0x2000
+#define TLS_OPT_REQUIRE_PBSZ        0x4000
 
 /* mod_tls SSCN modes */
 #define TLS_SSCN_MODE_SERVER				0
@@ -10654,6 +10655,8 @@ MODRET tls_post_pass(cmd_rec *cmd) {
 }
 
 MODRET tls_prot(cmd_rec *cmd) {
+  config_rec *c = NULL;
+  unsigned long tls_opts = 0;
   char *prot;
 
   if (!tls_engine ||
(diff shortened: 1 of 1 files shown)


contrib/mod_tls.c
```diff
@@ -10674,13 +10677,28 @@ MODRET tls_prot(cmd_rec *cmd) {
...
+  c = find_config(main_server->conf, CONF_PARAM, "TLSOptions", FALSE);
+  while (c != NULL) {
+    unsigned long opts = 0;
+
+    pr_signals_handle();
+    opts = *((unsigned long *) c->argv[0]);
+    tls_opts |= opts;
+
+    c = find_config_next(c, c->next, CONF_PARAM, "TLSOptions", FALSE);
+  }
```
Castaglia: You probably don't need this `while` loop here; the `TLSOptions` configuration is looked up at session initialization time, and stored in the `tls_opts` file-global variable, which means that the above loop is probably redundant.

Unless you found it was needed for some other reason?

contrib/mod_tls.c
```diff
@@ -10674,13 +10677,28 @@ MODRET tls_prot(cmd_rec *cmd) {
...
+
   if (!(tls_flags & TLS_SESS_PBSZ_OK)) {
-    pr_response_add_err(R_503,
-      _("You must issue the PBSZ command prior to PROT"));
+    if ( tls_opts & TLS_OPT_REQUIRE_PBSZ ) {
```
Castaglia: Style nit: the ProFTPD coding style does not use spaces after an open parenthesis (`(`), or before the close parenthesis (`)`), thus:

    if (tls_opts & TLS_OPT_REQUIRE_PBSZ) {

would be more preferred.

contrib/mod_tls.c
gczuczy: Frankly, I've been looking at the source how to get that value, and I've copy-pasted this part from another function.

So, I should just remove the tls_opts local variable and simply use the global one?

contrib/mod_tls.c
gczuczy: fixed.

contrib/mod_tls.c
Castaglia: Yes, that would be best -- and should still work just as you desire/need.

PR #289, opened 2016-05-24
Use the statvfs(2)/statfs(2) f_bsize (block size) field, rather than …

Castaglia:
…f_frsize

(fragment size), as we want to calculate the "available file system size"
based on f_bavail ("available blocks").


MofX:
statfs is one piece of badly documented API... Actually the code was correct to use f_frsize. f_bsize is supposed to be the "preferred block size" and f_frsize is supposed to be the actual fragmentation aka block size. BUT if f_frsize is not existing or not plausible f_bsize should be used.... But after merging this pull request, the calculation could actually be wrong.

PR #313, opened 2016-07-29
Bug#4262: SITE CPTO returns 250 even when the copy operation fails due to the user's quota being exceeded

Castaglia:


dup2:
Glad this is taken care of!

If SITE CPTO is failing in the middle of an extensive copy operation with this error, we have an incomplete copy where the client has no way to detect the  effective state of the target. 

Can the server remove the incomplete copy? Is this something we should implement or leave to the client? For the latter, the client would need some logic or scripting to cope with a situation like this, the server could somehow do the operation in a pseudo-atomic fashion.

What do you think?

Castaglia:
That, I think, is a related but different issue/feature request; I've filed https://github.com/proftpd/proftpd/issues/314 to track that.

Issue #314, opened 2016-08-02
Support deleting of failed copies


Castaglia:
Much as the `DeleteAbortedStores` directive can be used to tell `mod_xfer` to delete any files that fail to upload successfully, the `mod_copy` module _should_ support the ability to delete failed _copies_ (_e.g._ due to filesystem quotas, `mod_quotatab` quotas, _etc_).

Castaglia:
Merged/fixed in master.

PR #315, opened 2016-08-03
Support deletion-on-failure, for a failed copy.

Castaglia:
Addresses Issue #314 .

### DIFF
--- a/contrib/mod_copy.c
+++ b/contrib/mod_copy.c
@@ -39,6 +39,8 @@ extern pr_response_t *resp_list, *resp_err_list;
 
 module copy_module;
 static int copy_engine = TRUE;
+static unsigned long copy_opts = 0UL;
+#define COPY_OPT_DELETE_ON_FAILURE	0x0001
 
 static const char *trace_channel = "copy";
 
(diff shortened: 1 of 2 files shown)


uschaufelberger:
Sweet, thanks a lot for this enhancement, which works fine on single files.
However, delete on failed SITE copy currently fails when the source / target is a directory, tracing giving me:
`2016-08-04 15:48:48,964 [22889] <response:7>: error response added to pending list: 552 CPTO: Disk quota exceeded
2016-08-04 15:48:48,964 [22889] <fsio:8>: using system unlink() for path '/dst'
2016-08-04 15:48:48,964 [22889] <copy:7>: error deleting failed copy '/dst': Is a directory
`

Castaglia:
Ah, yes -- excellent catch.  I'll adjust this, as part of working [Bug#4263](http://bugs.proftpd.org/show_bug.cgi?id=4263], as it will involve adjusting the signature of the core `pr_fs_file_copy()` function.

Castaglia:
@uschaufelberger to address this, I will have to change the added `CopyOptions DeleteOnFailure` option.  Turns out that the `pr_fs_copy_file()` function _already deleted the destination on failure_ -- just not *all* of the failure cases.  So that, actually, was the root cause.

In fixing that, I am thinking to change the directive to be `CopyOptions NoDeleteOnFailure`, for any sites who do *not* want the deletion-on-failure capability.  Make sense?

dup2:
Makes perfectly sense for me. Most sites will prefer the cleanup now done by default with the inverted option.

Castaglia:
This functionality (auto-delete-on-failure, with optional no-delete-on-failure configurable via `CopyOptions`) is now available in the master branch.

PR #336, opened 2016-09-15
Retrieving bytes of growing files

inremen:
It retrieves bytes while the file is growing. After retrieving a block
of bytes, it waits two seconds in order to check the new size of the
file.

### DIFF
--- a/modules/mod_xfer.c
+++ b/modules/mod_xfer.c
@@ -2513,6 +2513,15 @@ MODRET xfer_retr(cmd_rec *cmd) {
     }
 
     len = transmit_data(cmd->pool, nbytes_sent, &curr_pos, lbuf, bufsz);
+
+	/*waiting 2 secons to check if the file is growing*/
+	sleep(2);
+
+	/*check the new size of the file*/
+	if (stat(dir, &st2) == 0){
+		session.xfer.file_size=st2.st_size;		
+	}
+
     if (len == 0) {
       break;
     }


modules/mod_xfer.c
```diff
@@ -2513,6 +2513,15 @@ MODRET xfer_retr(cmd_rec *cmd) {
     }
 
     len = transmit_data(cmd->pool, nbytes_sent, &curr_pos, lbuf, bufsz);
+
+	/*waiting 2 secons to check if the file is growing*/
+	sleep(2);
```
Castaglia: Hardcoded `sleep(3)` calls are terrible; they should _at least_ be configurable.  In this particular case, why 2 seconds, rather than 1, or 5, or some other arbitrary number?

modules/mod_xfer.c
```diff
@@ -2513,6 +2513,15 @@ MODRET xfer_retr(cmd_rec *cmd) {
...
+
+	/*check the new size of the file*/
+	if (stat(dir, &st2) == 0){
+		session.xfer.file_size=st2.st_size;		
+	}
```
Castaglia: It would greatly help if you _first_ described what issue/problem/behavior you are encountering, and why you feel this change best addresses that behavior.  Otherwise, I don't have a way to evaluate what this change is trying to accomplish, or why.

inremen:
You are right. I t should be configurable.

We need these feature because we need to move videos that are recording. We choose two seconds after doing some test in our test server.

Castaglia:
OK.  This is an interesting use case, then.

Rather than having patching the core code, another approach to consider, for this, would be to have your own custom module which would inject the necessary delay when appropriate.  Maybe the delay is needed only in certain directories, or only for files whose names match a certain pattern/filter.  Either way, having a delay added, as done in your PR, into the core engine is not necessarily the best approach.

inremen:
So, are you going to add this feature?

Castaglia:
It's possible that I'll write a separate module.  However, there's another approach you might try _first_: using [`TransferRate`](http://www.proftpd.org/docs/modules/mod_xfer.html#TransferRate) to "slow down" the downloads, enough to give the writing process enough time to keep writing the video file.

Specifically, what if you used:

    <Directory /path/to/video/files/folder>
      # Slow down downloads from here to 1 KB/sec, to give enough time for the files to be written
      TransferRate RETR 1.0
    </Directory>

You may need to tweak the _KB/sec_ rate to one slow enough for your client speed and file-writing process.

Does that work?

PR #343, opened 2016-10-10
Address potential SQL Injection when using mod_sql_odbc

maniac777:
According to SQL99 (and later, and maybe earlier too) quotation sign (') should be escaped as double quotation ('') but module escaped it incorrectly (\').
Oracle doesn’t understand escape sequence used currently which leads to SQL injection.

### DIFF
--- a/contrib/mod_sql_odbc.c
+++ b/contrib/mod_sql_odbc.c
@@ -198,7 +198,7 @@ static void sqlodbc_escape_string(char *to, const char *from, size_t fromlen) {
         break;
 
       case '\'':
-        *to++ = '\\';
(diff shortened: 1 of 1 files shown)


Castaglia:
Which version of Oracle are you using?

maniac777:
I'm using 12c (exacly 12.1.0.2 + current PSU's) but i believe that the same
will be on 11g as it is 100% SQL99 compatible too.

2016-10-11 5:34 GMT+02:00 TJ Saunders <notifications@github.com>:

maniac777:
Is it something wrong with my pach? Do you need additional information to pull it?

PR #350, opened 2016-11-17
Bug#4275: Support OpenSSL 1.1.x API for 1.3.5 line.

Castaglia:


pghmcfc:
This PR introduces the use of X509_STORE_get1_crls for older OpenSSL (<1.1) versions, but that symbol itself appears to be a relatively recent introduction; it's available in openssl 1.0.0b (Fedora 12) but not openssl 0.9.8n (Fedora 11) and earlier. So this change is going to break builds on really old systems like CentOS/RHEL 5. There probably aren't many people trying to build current proftpd on those sorts of systems though...

Castaglia:
Argh.  Of course. ;)

I'll try to refactor this to handle the multiple versions more cleanly, when I get a chance, hopefully later tonight.  This issue affects the `master` branch as well.  Sigh.

PR #357, opened 2016-12-03
Address memory leaks in mod_facl, bad handling of FACLEngine

Castaglia:


modules/mod_facl.c
```diff
@@ -1092,27 +1102,36 @@ static int facl_fsio_access(pr_fs_t *fs, const char *path, int mode,
...
     }
+
+    errno = xerrno;
+    return -1;
   }
 
   pr_trace_msg(trace_channel, 10,
     "acl(2) returned %d ACL entries for path '%s'", nents, path);
 
-  acls = pcalloc(fs->fs_pool, nents * sizeof(aclent_t));
```
Castaglia: These allocations out of the `fs->fs_pool` are a *huge* resource leak; that pool is for the lifetime of the process, and is not cleared frequently.  Thus we instead create a temporary sub-pool, for this task, that we can release/destroy once we're done.

modules/mod_facl.c
```diff
@@ -1124,39 +1143,59 @@ static int facl_fsio_access(pr_fs_t *fs, const char *path, int mode,
...
   }
+
+  res = check_facl(tmp_pool, path, mode, acls, nents, &st, uid, gid,
+    suppl_gids);
+  xerrno = errno;
+
+# if defined(HAVE_BSD_POSIX_ACL) || \
+     defined(HAVE_LINUX_POSIX_ACL) || \
+     defined(HAVE_MACOSX_POSIX_ACL)
+  acl_free(acls);
```
Castaglia: And this call to `acl_free(3)` is the other main source of the memory leak here.

PR #373, opened 2016-12-24
Bug#4064: Upgrade to Test-Unit 0.25 for integration tests

Castaglia:
Make sure the ProFTPD integration testsuite runs using the `Test-Unit 0.25` Perl module.


Castaglia:
This has not actually needed for years.

Issue #361, opened 2016-12-06
Libressl support


tspicer:
### What I Did

Tried compiling Proftpd in Alpine:Edge which uses Libressl as a default
 
### What I Expected/Wanted

Proftpd compiles/operates in the same manner with Libressl as Openssl

### ProFTPD Version and Configuration

This is based on the latest source in Github

Castaglia:
This is going to require a fair amount of work; Libressl does not support the same functionality as OpenSSL, which, in turn, means that supporting libressl may result in *less* functionality in _e.g._ `mod_sftp`, `mod_tls`, `mod_sql_passwd`, _etc etc_.

For the record, I brought up an Alpine Edge (3.4) VM using Vagrant:

    Vagrant.configure("2") do |config|
      config.vm.box = "maier/alpine-3.4-x86_64"

      # Make sure the /etc/apk/repositories file contains:
      #
      # $ cat /etc/apk/repositories
      # http://dl-cdn.alpinelinux.org/alpine/v3.4/main
      # http://dl-cdn.alpinelinux.org/alpine/edge/main
      # http://dl-6.alpinelinux.org/alpine/edge/community
      #
      # After installing vim, then do `sudo apk update`.

      config.vm.provision "shell", inline: <<-SHELL
        apk update
        apk add git gdb make git vm build-base
        apk add libressl-dev
      SHELL
    end

And then, after cloning the `proftpd` GitHub repo and configuring it:

    $ ./configure --with-modules=mod_sftp:mod_tls

I encountered the following build errors:

```
gcc -DHAVE_CONFIG_H  -DLINUX  -I.. -I../include  -g2 -O2 -Wall -fno-omit-frame-pointer -c mod_tls.c
mod_tls.c: In function 'get_dh':
mod_tls.c:102:7: warning: implicit declaration of function 'DH_set0_pqg' [-Wimplicit-function-declaration]
   if (DH_set0_pqg(dh, p, NULL, g) != 1) {
       ^
mod_tls.c: In function 'read_cert':
mod_tls.c:120:34: warning: implicit declaration of function 'SSL_CTX_get_default_passwd_cb' [-Wimplicit-function-declaration]
   cert = PEM_read_X509(fh, NULL, SSL_CTX_get_default_passwd_cb(ssl_ctx),
                                  ^
mod_tls.c:121:5: warning: implicit declaration of function 'SSL_CTX_get_default_passwd_cb_userdata' [-Wimplicit-function-declaration]
[...]

PR #393, opened 2017-01-16
fix build with libressl

vakartel:


Castaglia:
How is this related to PR #414  ? And what's the driving concern/motivation?

Mostly I'm trying to get a feel for a) how many people want libressl support, b) how badly, and c) whether libressl can be used for all of the other crypto-wanting modules of ProFTPD, e.g. mod_auth_otp, mod_sql_passwd, mod_sftp, etc.

Castaglia:
This should now be fixed in master; see Issue #361.  Thanks!

PR #402, opened 2017-01-30
If we are not doing a snapshot build, then tweak the Version and Release

Castaglia:
tags so that the generated `.rpm` filenames match what you'd expect:

    proftpd-1.3.6rc5-1.el7.centos.x86_64.rpm

rather than what was being created:

    proftpd-1.3.6-rc5.el7.centos.x86_64.rpm

The `--with cvs` functionality, albeit old, from [Bug#2975](http://bugs.proftpd.org/show_bug.cgi?id=2975) is still kept (for now).

@pghmcfc this should address the RPM naming sufficiently, yes?

### DIFF
--- a/contrib/dist/rpm/proftpd.spec
+++ b/contrib/dist/rpm/proftpd.spec
@@ -254,7 +254,7 @@ ProFTPD server:
 %if %{usecvsversion}
 %setup -q -n %{name}-%{proftpd_cvs_version_main}
 %else
-%setup -q -n %{name}-%{version}%{?release_cand_version}
+%setup -q -n %{name}-%{proftpd_version}
 %endif
 
 # Avoid documentation name conflicts
--- a/contrib/dist/rpm/proftpd.spec
+++ b/contrib/dist/rpm/proftpd.spec
@@ -134,7 +134,7 @@ Source0:                ftp://ftp.proftpd.org/devel/source/proftpd-cvs-%{proftpd
 %else
 Version:                %{proftpd_version}
 Release:                %{release_version}%{?dist}
-Source0:                ftp://ftp.proftpd.org/distrib/source/proftpd-%{version}%{?release_cand_version}.tar%{srcext}
+Source0:                ftp://ftp.proftpd.org/distrib/source/proftpd-%{proftpd_version}.tar%{srcext}
 %endif
 BuildRoot:              %{_tmppath}/%{name}-%{version}-root
 Requires:               pam >= 0.99, /sbin/chkconfig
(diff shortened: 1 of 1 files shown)


pghmcfc:
Resulting RPM version looks OK to me in my testing, though unfortunately anyone installing a release candidate build will have trouble upgrading to a full release build since "1.3.6rc5" is a higher version number than "1.3.6" as far as rpm is concerned. For this reason, Fedora's guidelines say that the "rc5" part of the version string gets moved to a 0-prefixed release string, e.g. 1.3.6-0.1.rc5, which would then cleanly upgrade to 1.3.6-1.

Castaglia:
Hrm.  That makes it harder to get to the expected/preferred naming scheme then, doesn't it?

Would use of the `Epoch` tag help out in this sort of transition?

pghmcfc:
We tend to regard epoch bumps as a last resort, to be avoided wherever possible.

Supposing you made this change:
```
diff --git a/contrib/dist/rpm/proftpd.spec b/contrib/dist/rpm/proftpd.spec
index e04bf5e..de98463 100644
--- a/contrib/dist/rpm/proftpd.spec
+++ b/contrib/dist/rpm/proftpd.spec
@@ -54,6 +54,14 @@
 %global proftpd_version                        1.3.6rc5
 %global release_version                        1


+%if %(echo %{proftpd_version} | grep rc >/dev/null 2>&1 && echo 1 || echo 0)
+%global rpm_version %(echo %{proftpd_version} | sed -e 's/rc.*//')
+%global rpm_release 0.%{release_version}.%(echo %{proftpd_version} | sed -e 's/.*rc/rc/')
+%else
+%global rpm_version %{proftpd_version}
+%global rpm_release %{release_version}
+%endif
+
 %global usecvsversion                  0%{?_with_cvs:1}


 %global proftpd_cvs_version_main       1.3.6
@@ -159,8 +167,8 @@ Version:                %{proftpd_cvs_version_main}
 Release:                0.1.cvs%{proftpd_cvs_version_date}%{?dist}
 Source0:                ftp://ftp.proftpd.org/devel/source/proftpd-cvs-%{proftpd_cvs_version_date}.tar%{srcext}
 %else
-Version:                %{proftpd_version}
-Release:                %{release_version}%{?dist}
+Version:                %{rpm_version}
+Release:                %{rpm_release}%{?dist}
 Source0:                ftp://ftp.proftpd.org/distrib/source/proftpd-%{proftpd_version}.tar%{srcext}
 %endif
 BuildRoot:              %{_tmppath}/%{name}-%{version}-root
```
and you bumped release_version by 1 for each subsequent release candidate, then reset it to 1 for the final "gold" release. The rpms would then go in this sequence:

* proftpd-1.3.6-0.1.rc5
* proftpd-1.3.6-0.2.rc6
* proftpd-1.3.6-0.3.rc7
* proftpd-1.3.6-0.4.rc8
* proftpd-1.3.6-0.5.rc9
* proftpd-1.3.6-0.6.rc10
...
* proftpd-1.3.6-1

Which would satisfy the ordering (and Fedora's guidelines).

Issue #408, opened 2017-02-05
Updating to 1.3.5d-1.el7 on Centos 7 via YUM causes error (signal 11)


newmediaone:
### What I Did
I'm running Centos7 on Google Cloud.  Yum signaled an update from:
proftpd-1.3.5b
To 1.3.5d
(Yum shows:
The available updates are:

    proftpd.x86_64                         1.3.5d-1.el7                         epel)

I use proftpd for SFTP by adding this line to my `sshd_config`:

    Subsystem       sftp    /usr/libexec/openssh/sftp-server

*and* these lines to the end of `/etc/proftpd.conf`:

    LoadModule mod_sftp.c
    <IfModule mod_sftp.c>
        # Configure the server to listen on the normal SSH2 port, port 22
        Port 2222
        #PidFile /run/prosftpd/prosftpd.pid
        SFTPEngine on
        #SFTPLog /var/log/prosftpd/sftp.log
        #ScoreboardFile /run/prosftpd/prosftpd.scoreboard
        # Configure both the RSA and DSA host keys, using the same host key
        # files that OpenSSH uses.
        SFTPHostKey /etc/prosftpd/ssh_host_rsa_key
        #SFTPHostKey /etc/ssh/ssh_host_dsa_key
        # Enable compression
        SFTPCompression delayed
        # Allow the same number of authentication attempts as OpenSSH.
        #
        # It is recommended that you explicitly configure MaxLoginAttempts
        # for your SSH2/SFTP instance to be higher than the normal
        # MaxLoginAttempts value for FTP, as there are more ways to authenticate
        # using SSH2.
        MaxLoginAttempts 6
        SFTPClientMatch .*ClientSftp sftpProtocolVersion 3
    </IfModule>

After updating to 1.3.5d via YUM if any attempt to SFTP in failed with an login error, but checking the logs, it looks like it's crashing with signal 11.

Here's what the logs show on SFTP connection attempt (IP address part of logs removed):

    SSH2 session opened.
    ProFTPD terminating (signal 11)
    SSH2 session closed.

*After* reverting back, the logs show (IP address part of logs removed):

    SSH2 session opened.
    USER REALFTPUSENAMEHERE: Login successful

It appeared that _some_ SFTP clients (Coda on Mac) _still_ worked, but others (Yummy and Transmit on Mac) failed with signal 11.

I thought perhaps it was a issue with a certain encryption typeS, but after creating a valid DSA key and adding it to the configuration, I saw no difference.

[...]

PR #411, opened 2017-02-05
Bug#4287: Properly allocate (and clear) the UMAC contexts.  Failure t…

Castaglia:
…o do so leads to segfaults.


Castaglia:
Addresses Issue #408 .

Issue #415, opened 2017-02-13
unrecoverable backend error: (mod_sql_postgres/4.0.4) could not create SSL context: SSL error code 336236705


warpspeed6:
### What I Did
Compile latest version of ProFTPD with mod_proxy_protocol. 
rpmbuild -bb proftpd.spec --with ssl --with postgresql --with mysql

### What I Expected/Wanted

I was expecting proftpd to start fine but it runs fine, loads up all the modules fine but throws below error when a connection attempt is made. It does appear that mod_sql is expecting SSL connection and having trouble but I am not sure what else to check. 

Error is:
```
2017-02-10 04:45:48,753 mod_sql/4.3[4261]: exiting  postgres cmd_escapestring
2017-02-10 04:45:48,753 mod_sql/4.3[4261]: unrecoverable backend error
2017-02-10 04:45:48,753 mod_sql/4.3[4261]: error: 'mod_sql_postgres/4.0.4'
2017-02-10 04:45:48,753 mod_sql/4.3[4261]: message: 'could not create SSL context: SSL error code 336236705
'
2017-02-10 04:45:48,753 mod_sql/4.3[4261]: entering     postgres cmd_exit
2017-02-10 04:45:48,753 mod_sql/4.3[4261]: exiting  postgres cmd_exit

```
### ProFTPD Version and Configuration

```
# proftpd -V
Compile-time Settings:
  Version: 1.3.6rc5 (git)
  Platform: LINUX [Linux 4.4.19-29.55.amzn1.x86_64 x86_64]
  Built: Mon Feb 13 2017 02:29:18 UTC
  Built With:
[...]

PR #417, opened 2017-02-14
Properly inform Postgres of how to initialize OpenSSL, depending on w…

Castaglia:
…hich

modules are already loaded.  And also explicitly provide the `sslmode`
connection parameter when *any* SSL-related settings are used.

Addresses Issue #415 .

### DIFF
--- a/contrib/mod_sql_postgres.c
+++ b/contrib/mod_sql_postgres.c
@@ -2,7 +2,7 @@
  * ProFTPD: mod_sql_postgres -- Support for connecting to Postgres databases.
  * Time-stamp: <1999-10-04 03:21:21 root>
  * Copyright (c) 2001 Andrew Houghton
- * Copyright (c) 2004-2016 TJ Saunders
+ * Copyright (c) 2004-2017 TJ Saunders
  *  
  * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


contrib/mod_sql_postgres.c
```diff
@@ -1626,27 +1643,41 @@ static void sql_postgres_mod_unload_ev(const void *event_data,
...
+
+  if (pr_module_exists("mod_tls.c") == TRUE) {
+    init_ssl = FALSE;
+    init_crypto = FALSE;
   }
+
+  pr_trace_msg(trace_channel, 18,
+    "telling Postgres about OpenSSL initialization: ssl = %s, crypto = %s",
+    init_ssl ? "yes" : "no", init_crypto ? "yes" : "no");
+  PQinitOpenSSL(init_ssl, init_crypto);
```
Castaglia: The root cause of:

    could not create SSL context: SSL error code 336236705

is that `mod_sql_postgres` was not informing the Postgres client library of what needed to be initialized properly, OpenSSL-wise.  And since OpenSSL was not properly initialized for creating SSL contexts (needed for SSL sessions), the "could not create SSL context" error was happening.

The above fixes this.

PR #418, opened 2017-02-15
Modify the proftpd.spec such that the naming of the generated RPMs mo…

Castaglia:
…re closely

conforms to Fedora guidelines.  Graciously provided/suggested by Paul Howarth (@pghmcfc).

### DIFF
--- a/contrib/dist/rpm/proftpd.spec
+++ b/contrib/dist/rpm/proftpd.spec
@@ -52,7 +52,22 @@
 # so use --with rhel5 to inhibit those features when using --with everything
 
 %global proftpd_version			1.3.6rc5
+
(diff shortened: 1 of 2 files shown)


contrib/dist/rpm/proftpd.spec
```diff
@@ -52,8 +52,23 @@
 # so use --with rhel5 to inhibit those features when using --with everything
 
 %global proftpd_version			1.3.6rc5
+
+# rc_version should be incremented for each RC release, and reset back to 1
+# AFTER each stable release.
+%global rc_version			5
```
Castaglia: Hopefully starting with at 5, so that it matches the "rc5" number, is OK.

Issue #420, opened 2017-02-16
Support building against LibreSSL, rather than OpenSSL


Castaglia:
There are now two different pull requests (PR #393 and PR #414) for compiling `mod_tls` successfully when LibreSSL, rather than OpenSSL, are used.

Given this, this ticket/task is to ensure that *all* of the crypto-wanting modules of ProFTPD will built successfully when LibreSSL is used.  They may not be able to provide all of the same functionality, but the above PRs indicate there's enough admins out there who want this.

Castaglia:
I think this ticket duplicate Issue #361.  Oops.

PR #421, opened 2017-02-16
Issue #420: Support building against LibreSSL instead of OpenSSL.

Castaglia:


Castaglia:
And, of course, this looks to duplicate Issue #361.  Oops.

Castaglia:
This should address Issue #361 .

PR #455, opened 2017-03-18
Bug#4299: TimeoutLogin no longer working for SFTP connections.

Castaglia:
The root cause was the blocking of alarms when reading/writing SFTP packets;
this was done to avoid corrupting packets during rekeying.  But it had the
unwanted side effect of blocking alarms like TimeoutLogin.  To balance the
TimeoutLogin need, we block alarms for packet IO only after the client has
authenticated.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -69,7 +69,7 @@
   now supports a SortedNLST flag for such use cases.
 - Bug 4260 - Restarting server fails when using password-protected SSL
   certificates and no TLSPassPhraseProvider.
-- Bug 4126 - SSH rekeying causes invalid packet due to interrupting timer.
+- Bug 4216 - SSH rekeying causes invalid packet due to interrupting timer.
 - Bug 4278 - Memory leak when mod_facl is used.
 
 1.3.6rc2 - Released 10-Mar-2016
--- a/contrib/mod_sftp/mod_sftp.c
+++ b/contrib/mod_sftp/mod_sftp.c
@@ -1829,6 +1829,12 @@ static void sftp_shutdown_ev(const void *event_data, void *user_data) {
   }
 }
 
+static void sftp_timeoutlogin_ev(const void *event_data, void *user_data) {
+  if (sftp_sess_state & SFTP_SESS_STATE_HAVE_KEX) {
+    SFTP_DISCONNECT_CONN(SFTP_SSH2_DISCONNECT_BY_APPLICATION, NULL);
+  }
+}
+
 #ifdef PR_USE_DEVEL
 static void pool_printf(const char *fmt, ...) {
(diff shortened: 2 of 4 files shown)


jschvartz:
Hi,
Yes, it works, thank you.

While I'm on it: would it be possible to link this condition with "**mod_ban**", as I suspect this is a kind of threat (ddos ?); if timeout don't work, we promptly have tons of pending connections (chinese ...)

Also, just to inform, **reload** does not update TimeoutLogin, you have to stop & start proftpd.

Thanks for being so reactive,

Joel

Castaglia:
@jschvartz I'm not sure what you mean by "reload does not update TimeoutLogin" -- sounds like "reload" is something defined by your init scripts/system?  I think that any mod_ban-related issues, etc should be handled using separate issue tickets, lest this one become too cluttered for clear understanding the multiple factors.

jschvartz:
**@Castaglia** yes, sorry for not being precise enought: my 'reload' sends a SIGHUP signal to proftpd process, which then should reload the configuration file.

Castaglia:
@jschvartz could you provide/describe the before *and* after config files you use, for changing the `TimeoutLogin`, and how you tested whether the change was picked up properly?  I'd like to reproduce that behavior locally, if I can.

jschvartz:
before: TimeoutLogin 90
after: TimeoutLogin 60
(global section)

send a SIGHUP to proftpd process: kill -HUP (process)
try the new configuration by opening a session and not responding to password request
--> times out after 90 seconds
stop and restart proftpd
--> same action times out after 60 seconds.

Castaglia:
Hmm. I just tried to use the above reproduction recipe, locally, and it worked as I would expect.  In particular, I was looking at the ProFTPD debug logging for the "Login timeout exceeded, disconnected" log message (and its timestamp) in both cases.  As opposed to looking at the client UI, which may or not show receive the disconnect or display it...

jschvartz:
Yes, i made another try right now, and it worked ok.
Maybe it was a pb with pid file ?
Sorry for the false error …

Many thanks again.
Joel








De : TJ Saunders [mailto:notifications@github.com]
Envoyé : mardi 21 mars 2017 16:59
À : proftpd/proftpd
Cc : Joel Schvartz; Mention
Objet : Re: [proftpd/proftpd] Bug#4299: TimeoutLogin no longer working for SFTP connections. (#455)


Hmm. I just tried to use the above reproduction recipe, locally, and it worked as I would expect. In particular, I was looking at the ProFTPD debug logging for the "Login timeout exceeded, disconnected" log message (and its timestamp) in both cases. As opposed to looking at the client UI, which may or not show receive the disconnect or display it...

Issue #461, opened 2017-03-25
Handle more daemon-terminating signals by terminating child processes as well


Castaglia:
See PR #459.  The use case is that sometimes the `proftpd` daemon will be terminated by the OS using _e.g._ `SIGINT`.  The handling of that signal, unlike `SIGTERM`, causes the daemon process to send but *not to terminate the child processes*.  In such cases, where the _most likely_ culprit/cause of the `SIGINT` is a contented resource policy enforcement, the child processes should _also_ be terminated.

Castaglia:
I consider this addressed for now.

PR #462, opened 2017-03-25
Modify the handling of signals, affecting SIGILL and SIGINT, such tha…

Castaglia:
…t they,

like SIGTERM, cause the child processes to be terminated as well.  Addresses
Issue #461.  @ton31337 could you try this PR, see if its behavior is as you need?

### DIFF
--- a/src/signals.c
+++ b/src/signals.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD - FTP server daemon
- * Copyright (c) 2014-2016 The ProFTPD Project team
+ * Copyright (c) 2014-2017 The ProFTPD Project team
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -141,9 +141,9 @@ static void handle_evnt(void) {
   pr_event_generate("core.signal.USR2", NULL);
 }
 
-static void handle_terminate(void) {
+static void handle_terminate_with_kids(void) {
   /* Do not log if we are a child that has been terminated. */
-  if (is_master) {
+  if (is_master == TRUE) {
 
     /* Send a SIGTERM to all our children */
     if (child_count()) {
@@ -158,7 +158,7 @@ static void handle_terminate(void) {
   finish_terminate(term_signo);
 }
 
-static void handle_terminate_other(void) {
+static void handle_terminate_without_kids(void) {
   pr_log_pri(PR_LOG_WARNING, "ProFTPD terminating (signal %d)", term_signo);
   finish_terminate(term_signo);
 }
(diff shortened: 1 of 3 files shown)


src/signals.c
```diff
@@ -302,39 +302,58 @@ static RETSIGTYPE sig_terminate(int signo) {
...
+        signame = "SIGSEGV";
+        break;
+
+      case SIGXCPU:
+        recvd_signal_flags |= RECEIVED_SIG_XCPU;
+        signame = "SIGXCPU";
+        break;
+
+      default:
+        recvd_signal_flags |= RECEIVED_SIG_TERMINATE;
```
ton31337: with this you are ignoring SIGTERM :) this means, that childs will never be killed, because of first bad condition.

ton31337: I meant here https://github.com/proftpd/proftpd/pull/462/files#diff-75dcdde0b93b7f3e0f4444cdfac71befR304

ton31337: So we have to add `signo == SIGTERM ||` condition too

Castaglia: Hrm.  I'll restructure this a little differently, I think -- the legibility of this function, for making this change, is making things difficult.

src/signals.c
```diff
@@ -298,43 +298,69 @@ static RETSIGTYPE sig_abort(int signo) {
...
-    } else {
-      recvd_signal_flags |= RECEIVED_SIG_SEGV;
-    }
+    case SIGINT:
+      recvd_signal_flags |= RECEIVED_SIG_TERMINATE;
+      signame = "SIGINT";
+      break;
 
+    default:
+      break;
```
ton31337: why it would be bad here to set `recvd_signal_flags |= RECEIVED_SIG_TERMINATE;` ? because I see, that some other signals like `SIGTRAP` is leaving children hanging up too

Castaglia: There are some signals for which ProFTPD does not currently have any explicit handling, _e.g._ `SIGTRAP`, `SIGXFSZ`, others.  For those, the response would be the OS default for them.

Thus once this PR is merged, there'll be a follow-on one to add handling for those signals.  I'm thinking `SIGTRAP`, `SIGXFSZ` -- any others?

ton31337: It's hard to say this in advance ;-) If I will notice something that would be handled like this, will open PR or issue 👍

ton31337: For now, SIGINT is handled OK as expected

Castaglia: So `SIGTRAP` is usually emitted by debuggers; if one was debugging the daemon process, I can't see why that should cause the child processes to be killed/terminated...

ton31337: right, this isn't actual signal

Issue #485, opened 2017-04-18
Writes to wtmpx are incorrect when compiled in 64bit mode on x86 Solaris


jiri-k:
I've compiled proftpd as 64bit program on Solaris 11 (LDFLAGS=-m64;CFLAGS=-m64) and I've found out that the "last" command shows a user is still logged in (using ftp) after the user logs out:
\# last
john      ftp                           Tue Apr 18 16:31   still logged in
...

The problem is in function log_wtmp. Proftpd uses "struct utmpx" when writing to wtmpx, but the entries in the file are stores as "struct futmpx".
These two structures are the same when compiling a program in 32bit mode but differ in 64bit mode.
The problem is already corrected when building on SPARC, where futmpx is used.

jiri-k:
I've created pull request #486 to address this issue.

Castaglia:
Fix merged to master, and [backported to 1.3.6 branch](https://github.com/proftpd/proftpd/commit/105f7e33d9e9d695261ecfd0a7083eed50b99add).  Thanks!

PR #486, opened 2017-04-18
Fix writes to wtmpx when compiled in 64bit mode on Solaris.

jiri-k:
Addresses issue #485

### DIFF
--- a/src/wtmp.c
+++ b/src/wtmp.c
@@ -40,7 +40,7 @@ int log_wtmp(const char *line, const char *name, const char *host,
     !(defined(LINUX) || defined(__hpux) || defined (_AIX))
   /* This "auxilliary" utmp doesn't exist under linux. */
 
-#if defined(__sun)
+#if (defined(__sparcv9) || defined(__sun)) && !defined(__NetBSD__) && !defined(__FreeBSD__)
   struct futmpx utx;
   time_t t;
 
@@ -102,7 +102,7 @@ int log_wtmp(const char *line, const char *name, const char *host,
 #else /* SVR4 */
     utx.ut_syslen = strlen(utx.ut_host)+1;
 
-#  if defined(__sun)
+#  if (defined(__sparcv9) || defined(__sun)) && !defined(__FreeBSD__)
     time(&t);
     utx.ut_tv.tv_sec = (time32_t)t;
 #  else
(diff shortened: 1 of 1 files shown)


src/wtmp.c
```diff
@@ -40,7 +40,7 @@ int log_wtmp(const char *line, const char *name, const char *host,
     !(defined(LINUX) || defined(__hpux) || defined (_AIX))
   /* This "auxilliary" utmp doesn't exist under linux. */
 
-#if defined(__sparcv9) && !defined(__NetBSD__) && !defined(__FreeBSD__)
+#if defined(__sun)
```
Castaglia: Hmm.  These platform-specific macros are from before I'd assumed maintainership of the project.  This issue makes me wonder whether now might be the time to move these into the `configure` script, so that this becomes more of a feature macro...

jiri-k: Could you explain more how you mean it?
I can see how we can move definition of WTMPX_FILE into the `configure` script.
Do you mean there would be a specific macro that would tell us whether to use `struct utmpx` or `struct futmpx`?
I think that such a move should be done as a separate commit anyway, and this pull request should just fix this one bug.

Castaglia: Fair enough.  To be a little more conservative, in terms of not potentially changing behavior for the existing macros, would the following also work for your needs?

    #if (defined(__sparcv9) || defined(_sun)) && !defined(__NetBSD__) && !defined(__FreeBSD__)

jiri-k: Yes, it would.
But I still think using just `defined(__sun)` is better as `struct futmpx` is only used on Solaris, and this should limit us exactly to Solaris.

Castaglia: So then there are _two_ goals for this change: 1) make code compile properly on Solaris, and 2) restrict this code path to just Solaris?  I'm OK with the first purpose/goal, but would rather spend the time to rework the Autoconf detection if the second goal is required.

jiri-k: Ok. What I meant to say is that I think this code path was never meant for any other OS than Solaris, because `struct futmpx` is not defined anywhere else.
My main goal is to make code correct, but my secondary goal is also to make nicer/simpler code, if it works the same.
I'll update the pull request.

PR #497, opened 2017-05-03
Add --enable-tests=local-only option

pghmcfc:
This disables API tests that involve resolving/connecting to external
network services such as Google, which may not be possible in some
build environments (such as Fedora's koji build service).

Tested using systemd-nspawn --private-network

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -1068,6 +1068,9 @@
 /* Define if ncursesw support, if available, should be used.  */
 #undef PR_USE_NCURSESW
 
+/* Define if non-local network tests are enabled. */
+#undef PR_USE_NETWORK_TESTS
+
 /* Define if using nonblocking open of log files.  */
 #undef PR_USE_NONBLOCKING_LOG_OPEN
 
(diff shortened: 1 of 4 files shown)


Castaglia:
Nice!  The only change I might suggest is making the configure option be `--enable-tests=nonetwork`; this makes it more consistent with _e.g._ `--enable-devel=nodaemon` or `--enable-devel=nofork`, _i.e._ using a `no` prefix for the functionality to disable.  Thanks!

pghmcfc:
That's fine by me. Do you want to make that change or shall I do it?

I'm about to cycle home from work...

Castaglia:
If you make the change to your branch, I can then pull that branch in as-is.

pghmcfc:
I'd appreciate it if this could go in the 1.3.6 branch too, otherwise I'll need to carry it downstream for the lifetime of the 1.3.6 branch so that I can run the API tests as part of the build.

Castaglia:
I've cherry-picked the changes to the 1.3.6 branch.

pghmcfc:
Thanks!

PR #499, opened 2017-05-05
Don't touch TLSCipherSuite when using system profiles

pghmcfc:
Fedora and possibly other Linux distributions support system-wide
crypto policies to enable sane defaults to be specified in an ever
changing world of different cipher recommendations. In order to use
such a policy, OpenSSL users just set their cipher selection to
"PROFILE=SYSTEM", and the system-wide policy will be selected
(which can itself be set to various values, for best compatibility,
best strength, a compromise of the two, etc.).

See:
https://fedoraproject.org/wiki/Packaging:CryptoPolicies
https://fedoraproject.org/wiki/Changes/CryptoPolicy

The "PROFILE=SYSTEM" string cannot be used in conjunction with other
cipher selections, so prepending it with "!EXPORT:" results in:

mod_tls/2.7[xxxxx]: unable to accept TLS connection: client does not support
any cipher from 'TLSCipherSuite !EXPORT:PROFILE=SYSTEM' (see `openssl ciphers
!EXPORT:PROFILE=SYSTEM` for full list)

Hence, do not touch the supplied TLSCipherSuite if it starts with "PROFILE=".

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -11985,7 +11985,12 @@ MODRET set_tlsciphersuite(cmd_rec *cmd) {
   c = add_config_param(cmd->argv[0], 1, NULL);
 
   /* Make sure that EXPORT ciphers cannot be used, per Bug#4163. */
-  ciphersuite = pstrcat(c->pool, "!EXPORT:", ciphersuite, NULL);
+  /* This breaks system profiles though, so don't change them.   */
+  if (strncmp(ciphersuite, "PROFILE=", 8) == 0) {
+    ciphersuite = pstrdup(c->pool, ciphersuite);
+  } else {
+    ciphersuite = pstrcat(c->pool, "!EXPORT:", ciphersuite, NULL);
+  }
 
   /* Check that our construct ciphersuite is acceptable. */
   ctx = SSL_CTX_new(SSLv23_server_method());


contrib/mod_tls.c
```diff
@@ -11984,7 +11984,10 @@ MODRET set_tlsciphersuite(cmd_rec *cmd) {
   c = add_config_param(cmd->argv[0], 1, NULL);
 
   /* Make sure that EXPORT ciphers cannot be used, per Bug#4163. */
-  c->argv[0] = pstrcat(c->pool, "!EXPORT:", ciphersuite, NULL);
+  /* This breaks system profiles though, so don't change them.   */
+  if (strncmp(ciphersuite, "PROFILE=", 8)) {
+    c->argv[0] = pstrcat(c->pool, "!EXPORT:", ciphersuite, NULL);
+  }
```
Castaglia: Hmm.  I expect that OpenSSL's support/handling of `PROFILE` depends on the _version_ of OpenSSL used, and/or possibly _how_ that OpenSSL installation is built/compiled?  I'd like to restrict the scope of this change to only the OpenSSL installations for which it is pertinent...

Castaglia: On the other hand, only admins who explicitly request `PROFILE` would know to do so; using `PROFILE` for an older OpenSSL version would lead to an error.  So this change might be fine as is.

A nice addition (not for this PR) would be to use `SSL_CTX_set_cipher_list` on the constructed string, in the configuration directive handler, to detect any issues at startup time.  I'll work on that.

Castaglia: See https://github.com/proftpd/proftpd/pull/500; may merge the `PROFILE` handling PR is after that one?

pghmcfc:
Added on top of (and including) #500 but it's not working for me.
The config check works: the server fails to start if I set a TLSCipherSuite of "profile=system" (lower case) and I get the expected error message.
However, if I use upper case "PROFILE=SYSTEM", the server starts but an attempted connection results in this:
```
2017-05-05 22:59:16,346 mod_tls/2.7[27668]: SSL/TLS-P requested, starting TLS handshake
2017-05-05 22:59:16,457 mod_tls/2.7[27668]: unable to accept TLS connection: protocol error: 
  (1) error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher
2017-05-05 22:59:16,457 mod_tls/2.7[27668]: unable to accept TLS connection: client does not support any cipher from 'TLSCipherSuite ' (see `openssl ciphers ` for full list)
2017-05-05 22:59:16,457 mod_tls/2.7[27668]: SSL/TLS-P negotiation failed on control channel
```
It was working OK before #500 though.

Castaglia:
Odd.  What does `openssl version -a` show, for the system on which you're testing your changes?

contrib/mod_tls.c
```diff
@@ -11984,8 +11985,29 @@ MODRET set_tlsciphersuite(cmd_rec *cmd) {
   c = add_config_param(cmd->argv[0], 1, NULL);
 
   /* Make sure that EXPORT ciphers cannot be used, per Bug#4163. */
-  c->argv[0] = pstrcat(c->pool, "!EXPORT:", ciphersuite, NULL);
+  /* This breaks system profiles though, so don't change them.   */
+  if (strncmp(ciphersuite, "PROFILE=", 8)) {
+    ciphersuite = pstrcat(c->pool, "!EXPORT:", ciphersuite, NULL);
+  }
```
Castaglia: Ah.  If you are *not* using `pstrcat()` on the string, then the memory buffer for that string might be reused.  Thus you might try:

    if (strncmp(ciphersuite, "PROFILE=", 8) == 0) {
      ciphersuite = pstrdup(c->pool, ciphersuite);

    } else {
      ciphersuite = pstrcat(c->pool, "!EXPORT:", ciphersuite, NULL);
    }

We want to have that `ciphersuite` string allocated out of the `c->pool` in either situation.  Does that make sense?

pghmcfc:
That's better!

Curious log entry though:
```
2017-05-06 19:04:19,983 mod_tls/2.7[1970]: SSL/TLS-P requested, starting TLS handshake
2017-05-06 19:04:20,106 mod_tls/2.7[1970]: client supports secure renegotiations
2017-05-06 19:04:20,106 mod_tls/2.7[1970]: TLSv1.2 connection accepted, using cipher ECDHE-RSA-AES128-GCM-SHA256 (128 bits)
2017-05-06 18:04:20,133 mod_tls/2.7[1970]: Protection set to Private
2017-05-06 18:04:20,137 mod_tls/2.7[1970]: starting TLS negotiation on data connection
2017-05-06 18:04:20,142 mod_tls/2.7[1970]: TLSv1.2 data connection accepted, using cipher ECDHE-RSA-AES128-GCM-SHA256 (128 bits)
```
The logging for the data connection is timestamped an hour earlier than the control connection.
Possibly relevant is that my local timezone is currently UTC+0100.

pghmcfc:
Incidentally, here is the openssl downstream patch in Fedora to support system-wide cipher lists:

http://pkgs.fedoraproject.org/cgit/rpms/openssl.git/tree/openssl-1.1.0-system-cipherlist.patch

Castaglia:
Are your sessions chrooted?  The TLS handshake for control connections typically happens _before_ authentication, and thus before any possible `chroot(2)` (which has ramifications for timestamps, as documented [here](http://www.proftpd.org/docs/howto/Timestamps.html)); data connection TLS handshakes occur after any possible `chroot(2)`.

pghmcfc:
I use mod_vroot (which you suggested to me many years ago to work around some pam session messages that would turn up in logs), so that's probably it.

Castaglia:
Did you want to update this PR for the current (changed) code in master?  It looks like a worthwhile change to merge...

pghmcfc:
I believe it's up to date now.

Castaglia:
I've also backported this change to the 1.3.6 branch.

PR #503, opened 2017-05-09
Explain the function of the NoRootPrivs parameter to CreateHome

natefoo:
The documentation mentions this parameter and shows an example of its usage but doesn't explain what it does, so here's a short explanation for the documentation, drawing from the explanation in [bug #3813](http://bugs.proftpd.org/show_bug.cgi?id=3813).


Castaglia:
Nice!  Thank you!

PR #506, opened 2017-05-10
Switch to Type = simple and add configuration test

pghmcfc:
Upstream recommends Type = simple if possible rather than Type = forking:
http://0pointer.de/public/systemd-man/daemon.html#Integration%20with%20Systemd

Also add configuration test prior to starting the daemon, to help diagnose
start-up problems.


Castaglia:
Cherry-picked this to the 1.3.6 branch, too.

PR #513, opened 2017-05-18
Fix pr_str_get_similars

pghmcfc:
The comparison function used with qsort(3) for pr_str_get_similars
was missing a level of pointer dereferencing, and hence did not
actually work as intended.

The accompanying test, which had workarounds for some of the
resulting flakiness, can now be more deterministic.


Castaglia:
Thanks!

PR #514, opened 2017-05-18
Use /etc/hosts rather than /etc/resolv.conf in fsio unit tests

pghmcfc:
The fsio unit tests require a read-only system file to test that
files can be read, can't be written or deleted etc. The file
/etc/resolv.conf is currently used for this, but does not exist
in the minimum build environment used on Fedora's koji build
servers, resulting in test failures. Using /etc/hosts, which does
exist there and should be equally ubiquitous, fixes this issue.

### DIFF
--- a/tests/api/fsio.c
+++ b/tests/api/fsio.c
@@ -119,8 +119,8 @@ START_TEST (fsio_sys_open_test) {
 
   mark_point();
   flags = O_RDONLY;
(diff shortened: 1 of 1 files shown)


Castaglia:
Thanks!

Castaglia:
Cherry-picked this change to the 1.3.6 branch as well.

PR #532, opened 2017-07-10
Fix RFC 2389 compliance regression in FEAT output.

jmaggard10:
This is a fix for a regression of the original bug: http://bugs.proftpd.org/show_bug.cgi?id=2752.
One symptom is that FileZilla won't be able to function on files with multi-byte UTF-8 characters in the name.

### DIFF
--- a/modules/mod_core.c
+++ b/modules/mod_core.c
@@ -6136,10 +6136,11 @@ MODRET core_feat(cmd_rec *cmd) {
   /* Sort the features, for a prettier output. */
   qsort(feats->elts, feats->nelts, sizeof(char *), feat_cmp);
 
-  pr_response_add(R_211, "%s", _("Features:"));
+  feat = pstrcat(cmd->tmp_pool, "Features:", NULL);
   for (i = 0; i < feats->nelts; i++) {
-    pr_response_add(R_DUP, "%s", ((const char **) feats->elts)[i]);
+    feat = pstrcat(cmd->tmp_pool, feat, "\r\n ", ((const char **)feats->elts)[i], NULL);
   }
+  pr_response_add(R_211, "%s", feat);
   pr_response_add(R_DUP, _("End"));
 
   return PR_HANDLED(cmd);


Castaglia:
With no description of what the issue being fixed, how to verify it, _etc_, this PR is not likely to be merged...

jmaggard10:
I updated the description.

modules/mod_core.c
```diff
@@ -6136,10 +6136,11 @@ MODRET core_feat(cmd_rec *cmd) {
   /* Sort the features, for a prettier output. */
   qsort(feats->elts, feats->nelts, sizeof(char *), feat_cmp);
 
-  pr_response_add(R_211, "%s", _("Features:"));
+  feat = pstrcat(cmd->tmp_pool, "Features:\n ", NULL);
   for (i = 0; i < feats->nelts; i++) {
-    pr_response_add(R_DUP, "%s", ((const char **) feats->elts)[i]);
+    feat = pstrcat(cmd->tmp_pool, feat, ((const char **) feats->elts)[i], "\n ", NULL);
```
Castaglia: `FEAT` lines should be terminated using `\r\n`, not just `\n`'; the adding of CRLF characters should be done properly by `pr_response_add()`.  So this patch appears to be _removing_ the use of CRLF, in favor of just LF...?

Castaglia:
This PR deals with `FEAT` output; how is that related to the emitting of multibyte filenames?

jmaggard10:
Some clients depend on "UTF8" being included in the features list to support multi-byte filenames.  If the client is unable to parse the FEAT response, it will not know that UTF8 is supported by the server.

I have updated the patch again to fix the remaining issues.

Castaglia:
Could you also provide a test demonstrating the regression, to prevent future mistakes/changes affecting this?

Castaglia:
Here's the [regression test I added/wrote](https://github.com/proftpd/proftpd/pull/639), for asserting the `FEAT` response lines are properly CRLF-terminated.

rot13:
I found this pull request after hours of searching why utf8 broke after upgrading an old ftp server to Debian 10.

While this patch fixes the problem, so does changing MultilineRFC2228 option to off. I wonder why this option was added 20 years ago, and why Debian set it to on in their default configuration. Maybe the reason behind it is no longer relevant nowadays. So far I had no complains from users after setting it to off.

The option's name is misleading. It sounds as turning it on makes ProFTPD RFC2228 compliant, while actually it does the opposite.

Since the whole point of the option is to change multiline response mode for ALL responses, not just 6xy, I'm not sure making an exception for FEAT output is the proper way to go. Some clients for which this option was made may rely on there being no exceptions.

I would definitely like a documentation change though, to mention this problem in the option description and maybe in mod_lang faq section too. The problem is documented, but it's at the end of mod_core documentation, which makes it difficult to find. I found the option responsible for my problems faster by just reading the code.

Castaglia:
Closing this, per https://github.com/proftpd/proftpd/issues/1089.

Issue #445, opened 2017-03-07
Remove non-functional TransferPriority directive


Castaglia:
Listing directories (_e.g._ via `LIST`, `NLST`, and/or `MLSD`) can be quite intensive for a server.  To help mitigate any "overload" situations, it would be nice if the `mod_ls` and `mod_facts` modules implemented/supported a `ListPriority` directive, akin to the existing [`TransferPriority`](http://www.proftpd.org/docs/modules/mod_xfer.html#TransferPriority) directive.

Castaglia:
Actually, the existing `TransferPriority` directive, being defined for "all data transfers", _should_ also apply to `LIST`, `NLST`, `MLSD` -- but it does not.  The `mod_facts` and `mod_ls` modules would need to be change to look for/honor this directive as well.

I like that idea.  And it will lead to refactoring the priority-manipulating code into a core API.

Castaglia:
I started on working on this, and then realized that the existing `TransferPriority` implementation _does not work at all_.  For two reasons.

First, the configuration directive has a _flags_ member, a bitset, to indicate which commands are to be affected.  That bitset is _never checked_, and thus the priority manipulation never happens at present.

Second, most of the sessions will have root privs revoked entirely.  And root privileges are necessary for _raising_ the priority via `setpriority(2)`, even if it is to _restore_ the previous priority.  Not great.  Especially since we want to move more and more away from using root privileges.

So now, I'm thinking to repurpose this ticket to be removing the `TransferPriority` directive entirely.

Castaglia:
Fixed in master.

PR #549, opened 2017-07-31
Issue #445: Remove the TransferPriority directive; it never functioned as

Castaglia:
intended.


Castaglia:
Addresses Issue #445.

PR #563, opened 2017-08-14
Enable mod ldap extensions

ws0w:
Related to issue: #546

The purpose of this modification is to enable extensions to mod_ldap, without having to re-implement LDAP authentication. 

By exposing two variables and one function in mod_ldap.h, new modules can leverage the existing LDAP authentication and focus only on the new feature(s) to be added.

This mod exposes the following through a new mod_ldap.h file:

extern LDAP *ld;
extern char *ldap_authbind_dn;
extern int pr_ldap_connect(LDAP **, int);

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -1,0 +1,32 @@
+Contributor Code of Conduct
+===========================
+
+As contributors and maintainers of this project, we pledge to respect all
+people who contribute through reporting issues, posting feature requests,
+updating documentation, submitting pull requests or patches, and other
+activities.
+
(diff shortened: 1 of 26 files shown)


Castaglia:
Shouldn't `contrib/mod_ldap.c` now `#include` the new header, to keep the function signatures and data types in sync, as well?

ws0w:
Yes, it should.

ws0w:
My requirements have been changing, and the code change in this pull request is no longer sufficient. I've copied mod_ldap into a new module, mod_ldapauth, and am now making my changes there.

So, I'm not sure if this pull request is worth spending any more time on.

ws0w:
No longer relevant.

Issue #568, opened 2017-09-05
CreateHome NoRootPrivs only works partially


natefoo:
### What I Did

With `CreateHome ... NoRootPrivs` set, the directory presence is *checked* as the logging-in user, but privs are switched back to the configured `User` and `Group` in order to *create* the directory (and parents), meaning that the `NoRootPrivs` option does not work as I expect it to.

For example. I have an root-squashed NFS filesystem at `/galaxy-repl` with subdirectories `/galaxy-repl/test`, `/galaxy-repl/test/ftp`:

Relevant directives:
 ```apache
User nobody
Group nobody
DefaultRoot ~
CreateHome on dirmode 700 NoRootPrivs
AuthOrder mod_sql.c
SQLDefaultUID       819456
SQLDefaultGID       803372
SQLUserInfo         custom:/LookupGalaxyUser
SQLPasswordUserSalt sql:/GetUserSalt
SQLNamedQuery       LookupGalaxyUser  SELECT "email, (CASE WHEN substring(password from 1 for 6) = 'PBKDF2' THEN substring(password from 38) ELSE encode(decode(password, 'hex'), 'base64') END),null,null,'/galaxy-repl/test/ftp/%U','/bin/bash' FROM galaxy_user WHERE email='%U'"
SQLNamedQuery       GetUserSalt SELECT "(CASE WHEN SUBSTRING (password from 1 for 6) = 'PBKDF2' THEN SUBSTRING (password from 21 for 16) END) FROM galaxy_user WHERE email='%U'"
```

The home directory parent is accessible as the `nobody` user:

```console
# su - nobody -s /bin/sh -c 'ls -al /galaxy-repl/test/ftp'
total 0
drwxrwxr-x  2 g2test G-803372 4096 Sep  5 12:13 .
drwxrwsr-x 28 g2test G-803372 4096 Sep 15  2016 ..
```

When I connect, authentication succeeds but home creation fails:

```
2017-09-05 11:36:45,189 galaxy07 proftpd[20555] :: (128.118.200.123[128.118.200.123]): retrieved UID 819456 for user 'nate@bx.psu.edu'
2017-09-05 11:36:45,189 galaxy07 proftpd[20555] :: (128.118.200.123[128.118.200.123]): no supplemental groups found for user 'nate@bx.psu.edu'
[...]

PR #570, opened 2017-09-06
Issue #568: Make sure that if CreateHome NoRootPrivs is used, we DO use

Castaglia:
the identity of the logging-in user, not the enclosing context's User/Group
identity.


Castaglia:
Yeah, that stack trace looks unrelated to this change.  That _said_, would you mind filing a different ticket for that stack trace, so that I don't forget to investigate, please?

PR #575, opened 2017-09-15
Using mask for supported attributes when building ATTRS payload

mligor:
There is a problem when creating ATTRS payload, because  the flags are just copied (including not supported flags by mod_sftp). This modification use predefined mask that is applied when creating new payload.


Castaglia:
Thanks!  I ended up addressing this differently (see Issue #578), so that I could include trace logging of the attributes that the client had requested, for better logging/triage.

PR #577, opened 2017-09-16
Reduce the amount of extra space we assume (and allocate) for xattr data

Castaglia:
in SFTP `READDIR` requests; it causes slower directory listings due to more
`READDIR` requests needed, since we send fewer entries back per request.


Castaglia:
This change has been cherry-picked/backported to the 1.3.6 branch as well.

Issue #578, opened 2017-09-16
SFTP OPEN response includes attribute flags that are not actually provided


Castaglia:
When the `mod_sftp` module receives an SFTP `OPEN` request, it reads the requested file attributes (as a bitset) from the client via the `fxp_attrs_read()` function.  In many cases, some of the attributes that a client _may_ request are ignored (as not supported on Unix and/or by the module).

However, that read-in attribute bitset is returned by that function, and then used in the `OPEN` response.  And therein lies the issue; `mod_sftp` does not provide the values in the response, for attributes that are requested (and still enabled) in the response bitset.

Instead, `mod_sftp` should include, in the `OPEN` response attributes bitset, only those attributes that are actually provided.

Castaglia:
Merged to master, and backported to 1.3.6 branch.

Castaglia:
+@mligor

mligor:
Thanks for fixing this.
I still have a problem with SSH2_FX_ATTR_MODIFYTIME flag. It is set but no valid time is found.

Castaglia:
@mligor what is the specific error, and/or unexpected behavior, that you're seeing with the `SSH2_FX_ATTR_MODIFYTIME` flag?  Can you open a new ticket for that issue, please?

mligor:
Sorry, but I can't really remember anymore. I had this issue during implementation some feature in bitvise SSH client. Unfortunately I'm not working for bitvise anymore and I have no access to the source code where I can try to find what kind of issue I had at that time.

PR #579, opened 2017-09-16
Issue #587: Before serializing out our supported SFTP attribute flags…

Castaglia:
…, make

sure we clear those flags (read in from the request) for any unsupported
attributes.


Castaglia:
Address Issue #578 .

PR #591, opened 2017-09-23
Bug #4318: If sendfile(2) returns zero, e.g. due to EOF (file truncat…

Castaglia:
…ion?),

break out of the loop, lest we loop infinitely.


Castaglia:
@yurashek this appears to address the infinite-loop-on-file-truncation behavior for me; does it work in your environment?

yurashek:
Yes exactly. Thank you for retaking it. I am not 100% familiar with signal/timers handling when transfer (sendfile()) returns.

PR #613, opened 2017-10-17
Bug #4319: Treat EINTR like EAGAIN

jmaggard10:
This bug described a situation where an ongoing transfer would be
prematurely aborted when one of our timers fired.  The timer could have
fired for an unrelated reason, but if we were in the process of reading
or writing with pr_netio_read() or pr_netio_write(), those calls would
be interrupted with errno set to EINTR, and an error would be returned.
Then pr_data_xfer() would abort the transfer.

EAGAIN was already being handled properly, and we can just use the same
treatment for EINTR so that we only respond to the timers we should
actually care about.

### DIFF
--- a/src/data.c
+++ b/src/data.c
@@ -1143,7 +1143,7 @@ int pr_data_xfer(char *cl_buf, size_t cl_size) {
         while (len < 0) {
           int xerrno = errno;
  
-          if (xerrno == EAGAIN) {
+          if (xerrno == EAGAIN || xerrno == EINTR) {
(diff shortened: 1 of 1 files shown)


Castaglia:
Do you think that `EINTR` should be handled _within_ `pr_netio_read()`, or do you think that it's better to have that sort of policy handled by the caller?

jmaggard10:
From my perspective, I think it's reasonable to let the caller handle EINTR since pr_netio_read() seems analogous to read().

Castaglia:
Makes sense to me.  Merging this!

PR #614, opened 2017-10-17
Rename the timer struct's 'count' member to 'remaining' to make its p…

jmaggard10:
…urpose more obvious.

### DIFF
--- a/src/timers.c
+++ b/src/timers.c
@@ -33,7 +33,7 @@ extern volatile unsigned int recvd_signal_flags;
 struct timer {
   struct timer *next, *prev;
 
-  long count;                   /* Amount of time remaining */
+  long remaining;               /* Amount of time remaining */
   long interval;                /* Original length of timer */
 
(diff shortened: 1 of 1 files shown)


src/timers.c
```diff
@@ -33,7 +33,7 @@ extern volatile unsigned int recvd_signal_flags;
 struct timer {
   struct timer *next, *prev;
 
-  long count;                   /* Amount of time remaining */
+  long remaining;               /* Amount of time remaining */
```
Castaglia: To be even more descriptive, would you be open to making this be `remaining_secs`?

jmaggard10: I'm not totally opposed to that.  Then would you want to also rename `interval` to `interval_secs`, and the `elapsed` variable in process_timers() to `elapsed_secs` for consistency?

Castaglia: Yes, that would be great.  Thanks!

PR #617, opened 2017-10-27
systemd: use network-online.target

pghmcfc:
With systemd, wait for network-online.target (rather than network.target)
before starting.

If your network is slow coming up (say you're using DHCP and the server is
unavailable or rebooting after a power outage, etc.) then proftpd might
come up before the network is sufficiently configured.

Refs:
https://bugzilla.redhat.com/show_bug.cgi?id=1119787
https://bugzilla.redhat.com/show_bug.cgi?id=1506805
https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

### DIFF
--- a/contrib/dist/rpm/proftpd.service
+++ b/contrib/dist/rpm/proftpd.service
@@ -1,6 +1,7 @@
 [Unit]
 Description = ProFTPD FTP Server
-After = network.target nss-lookup.target local-fs.target remote-fs.target
+Wants=network-online.target
+After=network-online.target nss-lookup.target local-fs.target remote-fs.target
 
 [Service]
 Type = simple


Castaglia:
@pghmcfc do you feel this change should be backported to the 1.3.6 branch as well?  I don't have a good feel for how much of an issue this might be, for proftpd + systemd users...

pghmcfc:
It'd make my life easier if you did backport it to the 1.3.6 branch as I'd be able to drop the downstream patch when 1.3.6a comes out. Looking at the documentation, it looks like the right thing to do for proftpd, and that's true for 1.3.6 as much as master. There's no back-compatibility issue either, as RHEL7 (oldest supported systemd-based OS I can think of) does have network-online.target.

Castaglia:
Done and done (cherry-picked to the 1.3.6 branch).  Thanks!

pprindeville:
Thanks!

PR #629, opened 2017-11-07
Bug #4322: Can't see symlinks using any FTP client when using MLSD.

Castaglia:
The bug also affects LIST and other commands; it is fallout from Bug#4219.


JordyZomer:
Thank you a lot @Castaglia  👍  🏆

PR #630, opened 2017-11-07
fix(group): remove dups when using option --add-member and sort users alphabetically

ianhuynh:
The --add-member option does not remove any duplicates (and so does option --member).

This PR handles removing any duplicates and also re-sort the users alphabetically before adding the list to the file.   I also updated the --help documentation

### DIFF
--- a/contrib/ftpasswd
+++ b/contrib/ftpasswd
@@ -130,12 +130,12 @@ if ((exists($opts{'passwd'}) && exists($opts{'group'})) ||
 
   # determine to which file to write the passwd entry
   if (defined($opts{'file'})) {
-    $output_file = $opts{'file'}; 
+    $output_file = $opts{'file'};
     print STDOUT "$program: using alternate file: $output_file\n"
 
   } else {
     $output_file = $default_passwd_file;
-  } 
+  }
 
   # make sure that the required arguments are present
   die "$program: --passwd: missing required argument: --name\n"
@@ -363,7 +363,7 @@ sub check_shell {
     if ($line eq $shell) {
       $result = 1;
       last;
-    } 
+    }
   }
 
   close(SHELLS);
@@ -646,7 +646,7 @@ sub get_passwd {
       my @string = split('', $hash);
       my $prefix = $string[0] . $string[1] . $string[2];
 
-      if ($prefix ne '$1$') { 
+      if ($prefix ne '$1$') {
         print STDOUT "You requested MD5 passwords but your system does not support it.  Defaulting to DES passwords.\n\n";
       }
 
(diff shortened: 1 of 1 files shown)


contrib/ftpasswd
```diff
@@ -130,12 +130,12 @@ if ((exists($opts{'passwd'}) && exists($opts{'group'})) ||
...
   # determine to which file to write the passwd entry
   if (defined($opts{'file'})) {
-    $output_file = $opts{'file'}; 
+    $output_file = $opts{'file'};
     print STDOUT "$program: using alternate file: $output_file\n"
 
   } else {
     $output_file = $default_passwd_file;
-  } 
+  }
```
Castaglia: I appreciate these whitespace nits, but it does make a PR like this harder to review.  In the future, please separate style changes like this from the actual functional changes.

contrib/ftpasswd
```diff
@@ -808,6 +808,15 @@ sub handle_group_entry {
     } else {
       $members = $add_user;
     }
+
+    # Remove any duplicates and resort
```
Castaglia: Is this solving a functional problem, or is this more for esthetics?

contrib/ftpasswd
ianhuynh: agree+ and sorry about the whitespace issue, not sure why my editor only picked up a few whitespace issue but not all.

contrib/ftpasswd
ianhuynh: 1. The sort:  is prob. more esthetics than functional but makes it easier to read.  
2. The dups:  For us it is just easier to audits if there are no duplicates and users are less prone to make mistakes as the list gets long.
3. the documentation: functional

ianhuynh: Actually, the PR fixed another issue.  If dups are allowed, the --delete-member left a trailing comma.   See below 


```
ian:~/git-repo/proftpd/contrib$ ./ftpasswd --group --name foo --gid 101 --add-member foo --member bar --member foo --member foo
ftpasswd2: creating group entry for group foo
ftpasswd2: entry created

ian:~/git-repo/proftpd/contrib$ cat ftpd.group
foo:x:101:bar,foo,foo,foo

ian:~/git-repo/proftpd/contrib$ ./ftpasswd --group --name foo --gid 101 --delete-member foo
ftpasswd2: updating group entry for group foo
ftpasswd2: entry updated

ian:~/git-repo/proftpd/contrib$ cat ftpd.group
foo:x:101:bar,
ian:~/git-repo/proftpd/contrib$
```

Castaglia:
Thanks!  I've opened [this PR](https://github.com/proftpd/proftpd/pull/640) which uses your code, and includes a regression test for the bug.

Castaglia:
Fixed via [other PR](https://github.com/proftpd/proftpd/pull/640).  Thanks!

PR #632, opened 2017-11-10
Fix SQLNamedConn and multiple backends

AdamJacobMuller:
When trying to use `SQLNamedConn` with a backend type different from `SQLBackend`, the backend connection was always being created with the `SQLBackend` type.

This fix calls `sql_set_backend` inside the `SQLNamedConn` loop (with the 2nd argument to `SQLNamedConn`) before we call `sql_define_conn` and later calls `sql_set_backend` to restore it to the original value of `SQLBackend`

I'm not sure this is a perfect fix, or even a good one (I'm not even really a C programmer) but this fixes my issue. I can generate and provide configurations that demonstrate the issue if you'd like.

### DIFF
--- a/contrib/mod_sql.c
+++ b/contrib/mod_sql.c
@@ -6333,6 +6333,8 @@ static int sql_sess_init(void) {
       pr_sql_conn_policy = SQL_CONN_POLICY_PERCALL;
     }
 
+    sql_set_backend(c->argv[1]);
+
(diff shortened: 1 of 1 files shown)


Castaglia:
Yes, would you mind opening a issue on this project, describing the behavior and the configurations you're using?  That way we can track the issue number in our release notes -- and I can try to use your configuration to reproduce the behavior locally.  Thanks!

Castaglia:
Thanks for the pull request!  This has now been fixed in master (with some logging code added for future debugging), and backported to the 1.3.6 branch, per Issue #642.

PR #636, opened 2017-11-18
David Ordal reported a bug in the popular Rebex SFTP library.  It is …

Castaglia:
…a client

side bug, but popular enough to warrant an SFTPOption for working around it.


lukaaash:
FYI, this bug has been fixed in Rebex SFTP 2017 R6.1 (build 6534) released on 2017-11-20.
Sorry for the inconvenience!

PR #638, opened 2017-11-19
Fixed mistake in configure localstatedir example

ChickenPeanut:
The documentation incorrectly stated that in order to place the configuration file in /etc/ and the runtime state files in /var/proftpd/, you would use:
$ ./configure --sysconfdir=/etc --localstatedir=/var
However the correct command would be
$ ./configure --sysconfdir=/etc --localstatedir=/var/proftpd


Castaglia:
Thanks!

PR #680, opened 2018-03-28
Bug #4341: scp broken on FreeBSD 11

Anthrazz:
This addition does fix scp transfers under FreeBSD 11.


Castaglia:
Thanks for the PR!  I'll merge this right now.

PR #690, opened 2018-04-16
Extend mod_exec to support additional backends and add IPC message queue module

offsides:
mod_exec doesn't work in a chroot environment, for various good reasons.  This patch extends mod_exec to support submodules (similar to mod_sql or mod_wrap), which can provide alternate methods of sending/executing commands.  It also add the mod_exec_mqueue module, which uses SysV IPC message queues to send messages to an external program and thus bypasses the chroot issue.  This was inspired by how NcFTPd handles it's events (as I am looking to replace NcFTPd with ProFTPd now that the former is no longer supported), and is configurable to send messages in the exact same format or raw without the NcFTPd-specific header.  Supports separate message queues per server instance, as well as separate message types based on both context and the command definition.

This was written against mod_exec 0.9.15 and ProFTPd 1.3.5e (running on CentOS 7), but the patch is clean against the latest mod_exec so it should work fine without any changes.

### DIFF
--- a/contrib/mod_exec.c
+++ b/contrib/mod_exec.c
@@ -41,7 +41,7 @@
 module exec_module;
 
 static pool *exec_pool = NULL;
-static int exec_engine = FALSE;
+int exec_engine = FALSE;
 static unsigned int exec_nexecs = 0;
 
 static int exec_logfd = -1;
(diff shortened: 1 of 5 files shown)


Castaglia:
Thanks!  Very nice.  Would you be up for providing some regression tests in Perl?  If not, I'd be happy to work on some...

deba12:
@offsides can you provide example program?

offsides:
@Castaglia - I'll take a look and see what I can do.  I'm not that familiar with Perl testing, so it'll depend on both time and the learning curve.

@deba12 - the best example I can give is the NcFTPd Event Daemon SDK, which is freely available at http://www.ncftp.com/ncftpd/doc/misc/eventd.html.  That's what my current NcFTPd-based system that I'm looking to move to ProFTPd is based on, and thus what I wrote this module to replicate.  Assuming  it's OK, I'll add a link to that page to the module docs.  Alternately since it's freely reproducible I could add the entire SDK, but that feels like a poor choice.

deba12:
@offsides thanks,
I think it will be OK to link NcFTPd sdk, but let @Castaglia decides

PR #704, opened 2018-05-25
- fs_is_nfs_test must be disabled on Solaris (and other OSes),

Sashan:
which struct statfs does not have f_fstypename member.

  Another option to fix this issue is to fix fs_is_nfs_test test,
  so it will take ENOSYS error code into account.


Castaglia:
Thank you for this patch; much appreciated!

PR #712, opened 2018-05-29
sendfile(fd, -1, ...) on solaris

Sashan:
returns EINVAL instead of EBADF

### DIFF
--- a/tests/api/data.c
+++ b/tests/api/data.c
@@ -291,7 +291,8 @@ START_TEST (data_sendfile_test) {
   mark_point();
   res = pr_data_sendfile(fd, &offset, 1);
   fail_unless(res < 0, "Failed to handle bad file descriptor");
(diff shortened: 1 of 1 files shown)


Castaglia:
Since this is for a test, I would address this differently.  That is, the test assertion could check for _either_ `EBADF` or `EINVAL`, rather than being _that_ platform-specific.  Thus:
```
  fail_unless(errno == EBADF || errno == EINVAL, "Expected EBADF (%d) or EINVAL (%d), got %s (%d)", EBADF, EINVAL, strerror(errno), errno);
```

Sashan:
got it. I've just force pushed the change you've suggested.

BTW travis clang configuration seems broken....

thanks and
regards
sasha

Castaglia:
Looks like Travis had a transient package installation issue; now fixed (by restarting those jobs).

PR #713, opened 2018-05-29
symlink("/tmp/prt-realink.lnk, "") creates 'empty' symlink

Sashan:
we must unlink() the '/tmp/prt-readlink.lnk' (empty symlink),
before proceeding further. Not doing so makes following test
to always fail with 'EEXIST'.

### DIFF
--- a/tests/api/misc.c
+++ b/tests/api/misc.c
@@ -309,6 +309,7 @@ START_TEST (dir_readlink_test) {
     res = dir_readlink(p, path, buf, bufsz, flags);
     fail_unless(res == 0, "Failed to handle empty symlink");
   }
+  (void) unlink(path);
 
   /* Not chrooted, absolute dst path */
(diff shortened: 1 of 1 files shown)


Sashan:
I did minor tweak to address suggestion proposed during the code review:
     we should check `unlink()` return code.

Sashan:
the earlier change segfaulted, format string asks for 2 ASCIIZ and 1 int, but it got just one ASCIIZ and one int.

Sashan:
let's close the loop and stick with original version. the only outcome here is a lesson I've just learned:
```
symlink("/dst/path/to", "");
```
creates different results on different OSes. Solaris creates `/dst/path/to`, which points
to nowhere. On the other hand linux, which runs on travis creates nothing. so having extra `(void) unlink("/dst/path/to");` should not hurt...

world is colourful...

PR #715, opened 2018-05-29
the hardwired DNS in tests/api/inet.c is bad

Sashan:
our network is behind the proxy, so test systems can't
talk to google DNS at all. This change introduces TEST_HOST
env. variable, so we can hint unit test with other DNS it can use.
if TEST_HOST env. variable is not defined the hardwired google
DNS (current 8.8.8.8) will be used instead

### DIFF
--- a/tests/api/inet.c
+++ b/tests/api/inet.c
@@ -24,7 +24,12 @@
 
 /* Inet API tests */
 
+#include <stdlib.h>
+
 #include "tests.h"
+
+#define	PR_TEST_DNS_RESOLVER	((getenv("PR_TEST_DNS_RESOLVER") == NULL) ? \
+			    "8.8.8.8" : getenv("PR_TEST_DNS_RESOLVER"))
 
 static pool *p = NULL;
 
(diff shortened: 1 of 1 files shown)


Castaglia:
Hmm.  Is there a different resolver that you could provide, in your test environment?  That is, rather than overriding the resolved host, could you provide an address, other than 8.8.8.8, via an environment variable like `PR_TEST_DNS_RESOLVER`?

Sashan:
I see. I've just force pushed an update, which does s/TEST_HOST/PR_TEST_DNS_RESOLVER.
is that what is your mind?

thanks and
regards
sasha

Castaglia:
I was thinking something more like [this](https://github.com/proftpd/proftpd/pull/719).  With that change, you can then tell the unit tests to use your internal resolver, if needed, _e.g._:

    $ PR_USE_DNS_RESOLVER=192.168.1.2 make api-tests

Sashan:
I've seen your change, it's more elegant than my 'ternary' hack. I'm going to close this pull request.
thanks a lot.

Issue #718, opened 2018-05-31
When MaxLoginAttempts is reached, the FTP session is closed before POST_CMD_ERR and LOG_CMD_ERR events can be executed


offsides:
### What I Did

When MaxLoginAttempts is set (in this case at 1), the POST_CMD_ERR and LOG_CMD_ERR commands are not executed on a failed login that reaches MaxLoginAttempts, and therefore the ExecOnError handler from mod_exec does not fire either.  This also means that failed passwords are not logged when disconnected for exceeding MaxLoginAttempts!

### What I Expected/Wanted
[proftpd.conf.txt](https://github.com/proftpd/proftpd/files/2059621/proftpd.conf.txt)

I use ExecOnError PASS commands to track failed logins.  I expect to have the ExecOnError PASS command run every time a login fails, even if the client is disconnected as a result.  Realistically, POST_CMD_ERR and LOG_CMD_ERR commands should execute before the FTP session is closed.

### ProFTPD Version and Configuration

RHEL7 ProFTPD rpm:
```
Compile-time Settings:
  Version: 1.3.5e (maint)
  Platform: LINUX [Linux 3.10.0-862.3.2.el7.x86_64 x86_64]
  Built: Wed Jan 31 2018 16:32:29 UTC
  Built With:
    configure  '--build=x86_64-redhat-linux-gnu' '--host=x86_64-redhat-linux-gnu' '--program-prefix=' '--disable-dependency-tracking' '--prefix=/usr' '--exec-prefix=/usr' '--bindir=/usr/bin' '--sbindir=/usr/sbin' '--sysconfdir=/etc' '--datadir=/usr/share' '--includedir=/usr/include' '--libdir=/usr/lib64' '--libexecdir=/usr/libexec' '--localstatedir=/var' '--sharedstatedir=/var/lib' '--mandir=/usr/share/man' '--infodir=/usr/share/info' '--libexecdir=/usr/libexec/proftpd' '--localstatedir=/run/proftpd' '--disable-strip' '--enable-ctrls' '--enable-dso' '--enable-facl' '--enable-ipv6' '--enable-memcache' '--enable-nls' '--enable-openssl' '--enable-pcre' '--enable-shadow' '--enable-tests' '--with-libraries=/usr/lib64/mysql' '--with-includes=/usr/include/mysql' '--with-modules=mod_readme:mod_auth_pam:mod_tls' '
[...]

PR #720, opened 2018-06-01
Offsides maxloginfix

offsides:
After a failed login exceeds MaxLoginAttempts, don't disconnect the session until after the POST_CMD_ERR and LOG_CMD_ERR events have been processed.  This fixes a security issue where failed logins won't be logged properly if they exceed MaxLoginAttempts, and also allows things like ExecOnError to work as they are called by POST_CMD_ERR.

This is a slightly cleaner patch than the one I proposed in issue #718.


Castaglia:
I like this approach!  I'll need a few days to write up some regression tests for this, just for my comfort, but so far, this approach looks very nice and appropriate.

PR #728, opened 2018-07-04
Always truncate before writing the file

yorickdewid:
Call truncate *before* writing the data to file so that the underlying storage driver can reserve the space. This is useful for filesystems that want to know the file size in advance, which can be the case in usermode filesystems like FUSE. Also, it is a good idea to truncate before writing anything to know if there is enough space left on the storage device.

### DIFF
--- a/contrib/mod_sftp/scp.c
+++ b/contrib/mod_sftp/scp.c
@@ -1447,51 +1447,51 @@ static int recv_path(pool *p, uint32_t channel_id, struct scp_path *sp,
     return recv_finfo(p, channel_id, sp, data, datalen);
   }
 
-  if (!sp->recvd_data &&
-      sp->recvlen != sp->filesz) {
-    if (cmd == NULL) {
-      cmd = scp_cmd_alloc(p, C_STOR, sp->best_path);
+  /* Always truncate before writing the contents to file to ensure proper
+   * file length, and inform the storage subsystem of the data size.
+   */
+  if (S_ISREG(sp->st_mode)) {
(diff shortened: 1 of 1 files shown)


Castaglia:
Would you object to having this be optional behavior, defaulting to opt-in initially?  A change like this can cause issues with other systems/setups, so ideally we'd mitigate that risk by having the feature be configurable.

Castaglia:
I'm thinking to re-implement this feature as a [`TransferOption`](http://www.proftpd.org/docs/modules/mod_xfer.html#TransferOptions), _e.g._:
```
# Truncate files before writing them on upload
TransferOption TruncateBeforeUpload
```
This way, it could be done for _all_ uploads, not just SCP uploads.

What would be interesting to know whether these filesystems in question, like FUSE, differentiate between a `truncate(2)` system call, or an `open(2)` system call with the `O_TRUNC` flag?

Castaglia:
I guess this is complicated by the fact that the intended size of the file to be uploaded is not always communicated.  FTP/FTPS uploads certainly do not know the size of the file being uploaded in advance.  SCP does.  SFTP _might_, depending on the protocol version and flags used.   So the desired goal, of having the filesystem be informed of the target size to be reserved, may not always be achievable.

Castaglia:
Unfortunately, without any additional information about the use cases (and thus how they apply to the multiple ways of file uploads supported by ProFTPD), I'll be unable to accept this PR.  We can re-examine and discuss this feature separately, via ticket, if you'd like.

PR #730, opened 2018-07-07
Guard against ENOATTR being undefined in API tests

pghmcfc:
Since updating libattr in Fedora Rawhide last week, ENOATTR was undefined and caused a compilation error in tests/api/fsio.c.


Castaglia:
Thanks!

PR #733, opened 2018-07-27
provide a default shell, when pwentry is empty

Sashan:
The idea for this bugfix comes from Oracle ZFSSA team. The have pointed out the empty `shell` field in pwentry should not be interpreted as invalid (not valid) shell. I could not find any mention in POSIX standard how applications should handle missing shell in pwentry. It seems to me the standard does not define any behaviour here.

Oracle Solaris ships ftp account with empty shell. It makes anonymous ftp logons to fail unless system administrator does either of two:

1. add `RequireValidShell off` anonymous server section in proftpd.conf
or
2. leave /etc/proftpd.conf intact and add empty line to /etc/shells 

I just did a quick research on how various unix-like OSes interpret empty shell field in pwentry:

[linux](https://www.systutorials.com/docs/linux/man/5-passwd/), [MacOS](http://www.manpagez.com/man/5/passwd/), [OpenBSD](https://man.openbsd.org/passwd.5), [NetBSD](http://netbsd.gw.com/cgi-bin/man-cgi?passwd+5), [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=passwd&sektion=5), all those seem to agree on something as follows:
```
The shell field is the command interpreter the user prefers.  If there
is nothing in the shell field, the Bourne shell (/bin/sh) is assumed.
```

[Oracle Solaris](https://docs.oracle.com/cd/E26502_01/html/E29042/passwd-4.html) stands outside of the crowd again:
```
is the user's initial shell program. If  this  field  is empty, the default shell is /usr/bin/sh.
```

The proposed pull request introduces `PR_DEFAULT_SHELL`, which can be assumed in case `shell` argument to `pr_auth_is_valid_shell()` is empty.


krytarowski:
NetBSD ships with `_PATH_BSHELL` in /usr/include/path.h.

Sashan:
> NetBSD ships with _PATH_BSHELL in /usr/include/path.h

I assume it's a [/usr/include/paths.h](http://cvsweb.netbsd.org/bsdweb.cgi/src/include/paths.h?rev=1.43&content-type=text/x-cvsweb-markup&only_with_tag=MAIN), looks like OpenBSD has the [same file](https://github.com/openbsd/src/blob/master/include/paths.h) with same definition of [_PATH_BSHELL[(https://github.com/openbsd/src/blob/master/include/paths.h#L43). FeeBSD is the same [here](https://github.com/freebsd/freebsd/blob/master/include/paths.h). I'll try to teach proftpd to handle *BSDs on my notebook over the weekend and will resubmit a better patch.
thanks for pointing that out.

Sashan:
So it has turned out suggestion by Mr. @krytarowski is more universal. The `include/paths.h` with definition of `_PATH_BSHELL` is available on many platforms (MacOS, Linux, *BSD). For all platforms, which define `_PATH_BSHELL` in `<paths.h>` we define PR_DEFAULT_SHELL as alias of _PATH_BSHELL, for platforms, which don't provide `_PATH_BSHELL`, we let proftpd to use an empty string `""`.

PR #761, opened 2018-11-15
bcp: update documentation to enable-buffer-size

jmbents:
Update documentation to use --enable-buffer-size instead of
the older --enable-tunable-buffer-size

Signed-off-by: Jared Bents <jared.bents@rockwellcollins.com>


Castaglia:
Thank you!

PR #764, opened 2018-11-20
let proftpd to drop core, when it dies because of segfault

Sashan:
The Stack Trace we get from customers is not that helpful sometimes.
We would like our customers to send us also a core file from dying
process. Unfortunately the proftpd is not allwed to drop the core.
Not sure if it is a bug or intention, hence I'm sending a patch,
which prevents proftpd dying on SigSegv to call exit(3C).

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -113,7 +113,7 @@ void pr_session_disconnect(module *m, int reason_code,
   }
 
   if (reason_code == PR_SESS_DISCONNECT_SEGFAULT) {
-    flags |= PR_SESS_END_FL_ERROR;
+    flags |= PR_SESS_END_FL_NOEXIT;
   }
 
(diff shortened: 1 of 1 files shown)


Castaglia:
The behavior of _not_ writing a core file is explicit/by design.  ProFTPD can contain, in memory, many secrets/credentials (private keys, passwords, _etc_) that would be written out to disk, as part of the core file, in such cases -- and we try to ensure that those secrets remain secret.

It should be possible to have your code, to write a core file, be written as a module.  That way, you can have your customers use your module to get what you need, without having to relax the behavior of ProFTPD for _all_ users.

Sashan:
OK I see, thanks for clarification. I'm closing the pull request.

PR #768, opened 2018-11-30
Support for masqueraded passive ports (MasqueradePassivePortsOrigin)

Benjamin-Dobell:
_This is not ready to be merged._

Presumably this PR will need some work surrounding tests and docs before this would be ready for merging. However, I'm opening this PR to get a conversation going around the feature itself. Additionally, support is presently only implemented with `PASV`, support for `EPSV` is missing entirely.

Basically the idea is we can now tell passive mode clients to connect on ports that are different from the ones ProFTPD itself is listening on. This allows ProFTPD to sit behind a L4 proxy and/or load balancer that is open on a different port range than what ProFTPD itself is available on.

L4 proxies have no idea about `PASV` (or any FTP semantics) so can't massage the port number for clients. So instead ProFTPD can be configured with `MasqueradePassivePortsOrigin <port>` where `<port>` is the first port of the proxy's / load balancer's port range. Then when responding to clients, ProFTPD provides ports relative to `MasqueradePassivePortsOrigin`, rather than the start of its own port range.

### DIFF
--- a/modules/mod_core.c
+++ b/modules/mod_core.c
@@ -959,6 +959,32 @@ MODRET set_masqueradeaddress(cmd_rec *cmd) {
   return PR_HANDLED(cmd);
 }
 
+MODRET set_masqueradepassiveportsorigin(cmd_rec *cmd) {
+    int port;
+    config_rec *c = NULL;
+
+    CHECK_ARGS(cmd, 1);
+    CHECK_CONF(cmd, CONF_ROOT|CONF_VIRTUAL|CONF_GLOBAL);
+
+    port = atoi(cmd->argv[1]);
+
+    /* Sanity check */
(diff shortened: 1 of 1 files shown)


Benjamin-Dobell:
Sadly, the fact RFC 2428 was so poorly designed as to include and IP address field, but specifically state it must be left blank, means we can't rely on masquerading in IPv6 environments.

The alternative is to ensure your port ranges match between your load balancer / proxy and ProFTPD, or use a L7 FTP proxy. Neither are ideal, but RFC 2428 makes ProFTPD's existing masquerading functionality unreliable/confusing, so it's probably best not to build additional masquerading functionality.

Thus, I'm abandoning this PR.

PR #778, opened 2019-01-21
Generate an event when clients disconnect explicitly

### LINKED ISSUE #777: Emit an event when a user disconnect explicitly (QUIT)
Currently, it does not seems possible for a module to hook into mod_core in order to differentiate explicit user disconnections from arbitrary disconnections.

This is a use case I have with one custom fsio module, which should only trigger some actions when the client disconnects explicitly (ie. full file upload) and not when the session is interrupted for external reasons.

Malizor:
This allows to distinguish clients exiting normally from all other
disconnections. This can be useful for external modules.

Fixes #777

### DIFF
--- a/modules/mod_core.c
+++ b/modules/mod_core.c
@@ -3634,6 +3634,11 @@ MODRET core_quit(cmd_rec *cmd) {
    * the session.
    */
 
+  /* This event allows modules to distinguish between a client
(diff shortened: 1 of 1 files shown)


Castaglia:
While I like the idea of this PR, it can be better implemented another way; see:
* https://github.com/proftpd/devel-guide/blob/master/devel/disconnects.md

Thanks for the PR!

PR #785, opened 2019-02-12
Synchronize with en_US.po, update translation

gdsotirov:
I have forgot about this translation, but ahead of it's 10th anniversary I hereby provide an update. :-)


Castaglia:
Thank you!

PR #787, opened 2019-02-21
- proftpd should not rely fsio I/O is instant

Sashan:
one of our customers hit this glitch on back up server,
which serves files from SAM-QFS [1] volume. The I/O on
such FS can take up to several seconds. This typically
happens when file must be retrieved from tape.

In this case the FSIO read(2) operation fails with
EINTR error. The EINTR failure is mishandled further:
instead of reporting error to remote client, server
sends file with 0 size.

The patch also makes the write(2) more robust. No one
his hit such issue so far, but I believe similar failures
might happen when received files are written to
NFS file systems.

the issue got uncovered during upgrade of 1.3.5b
to 1.3.6. Unlike 1.3.5b the new 1.3.6 guarantees
a minimum alarm(2) interval to be a 5sec.

[1] https://en.wikipedia.org/wiki/QFS


Castaglia:
I believe that now, with https://github.com/proftpd/proftpd/pull/1359, these occurrences should be addressed.  Thanks!

PR #798, opened 2019-04-20
Issue#797: Use the Autoconf `--libdir` (and `--includedir`) in the ge…

Castaglia:
…nerated

pkgconfig file.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -127,8 +127,8 @@ install-headers: $(DESTDIR)$(includedir)/proftpd
 install-pkgconfig: $(DESTDIR)$(pkgconfigdir)
 	@echo 'prefix=$(prefix)' > $(top_builddir)/proftpd.pc
 	@echo 'exec_prefix=$${prefix}' >> $(top_builddir)/proftpd.pc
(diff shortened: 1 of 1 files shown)


orbea:
This doesn't seem quite right, I applied the PR as a patch against master and found it made this file.
```
prefix=/usr
exec_prefix=${prefix}
libdir=${libdir}/proftpd
includedir=${includedir}/proftpd

Name: ProFTPD
Description: Professional FTP Daemon
Version: 1.3.7rc1
Requires: 
Libs: -L${libdir}
Cflags: -I${includedir}
```
I think `libdir=${libdir}/proftpd` which references itself will end up as an empty variable?

Castaglia:
Yes, you're right.  Let me refine this a bit.

orbea:
I tested the latest changes.
```
prefix=/usr
exec_prefix=${prefix}
libdir=/usr/lib64/proftpd
includedir=/usr/include/proftpd

Name: ProFTPD
Description: Professional FTP Daemon
Version: 1.3.7rc1
Requires: 
Libs: -L${libdir}
Cflags: -I${includedir}
```
This seems correct, the only thing to note is that it leaves the `prefix` variable unused, but in practice this is not a problem.

Thanks for the quick response!

PR #812, opened 2019-07-04
004.proftpd-retry.patch

Sashan:
FTP server in active mode may hit EADDRINUSE, when
connecting back to FTP client, when client port
range is limited. There is no easy way to quantify
'limited port range'. In some cases 100 ports on NAT
pool may be well enough (when files are long and
transmission lines are slow) in some cases 100 ports
in pool is not sufficient, when lines are fast and
files transferred are short.

As soon as file is transferred the connection is
closed (via shutdown(2)). This transfers socket/session
to TIMEWAIT state, which has typically 30sec duration.
However this seems to vary system to system [1].

With short files and fast lines 100 ports may get
wasted in couple of seconds (when doing mget), preventing
server to establish new connections. When server
is connecting to port number, which is still kept
in timewait state, the connect(2) operation fails
with EADDRINUSE.

Besides using this patch there are two ways
to fix/workaround the situation:
   - either let server to use any source port
     not just port 20

   - or set REUSEADDR/REUSEPORT socket option,
     so server will be connect to destination
     although the previous session is still
     in TIMEWAIT state.

The proposed patch works reasonably well, but
I can imagine it deserves some love, like introducing
yet another configuration to option to make
number of retry attempts tunable.

[1] https://en.wikipedia.org/wiki/Maximum_segment_lifetime

### DIFF
--- a/src/data.c
+++ b/src/data.c
@@ -214,6 +214,7 @@ static int data_passive_open(const char *reason, off_t size) {
 static int data_active_open(const char *reason, off_t size) {
   conn_t *c;
   int bind_port, rev, *root_revoke = NULL;
+  int retries = 0;
   const pr_netaddr_t *bind_addr = NULL;
 
   if (session.c->remote_addr == NULL) {
@@ -275,6 +276,7 @@ static int data_active_open(const char *reason, off_t size) {
     }
   }
 
+  for(;;) { /* begin of endless loop */
   session.d = pr_inet_create_conn(session.pool, -1, bind_addr, bind_port, TRUE);
   if (session.d == NULL) {
     int xerrno = errno;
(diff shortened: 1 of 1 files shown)


Castaglia:
ProFTPD _does_ set the `SO_REUSEADDR` and `SO_REUSEPORT` socket options; see [here](https://github.com/proftpd/proftpd/blob/master/src/inet.c#L321).

Sashan:
> ProFTPD _does_ set the `SO_REUSEADDR` and `SO_REUSEPORT` socket options; see [here](https://github.com/proftpd/proftpd/blob/master/src/inet.c#L321).

however `SO_REUSEADDR` does not count for implicit bind in `connect()`. At least on Solaris/[illumos](https://github.com/illumos/illumos-gate/blob/955eb5e1fb25fc64c2c34501c3d9426057e3b1f8/usr/src/uts/common/inet/tcp/tcp_bind.c#L858). Also seems to be case for OpenBSD as well, at least by quick look at [in_pcbconnect()](https://github.com/openbsd/src/blob/b764efda0543b1b3bc5df86127045ddbd161641c/sys/netinet/in_pcb.c#L476)

Issue #693, opened 2018-04-26
mod_sftp unsuccessful login count issues on AIX


ws0w:
### What I Did

I do not use AIX by choice.

On AIX 7.1 using gcc I compiled proftpd (latest master branch from github) with mod_sftp, and configured proftpd to allow SSH keys or passwords for authentication. I did not set up the ssh keys for the id tested.

Problem 1) When I sftp to proftpd, the unsuccessful login count gets incremented twice before the user is prompted for a password. From the trace log, I can see that it actually increments the count twice before checking for the SSH key. The unsuccessful login count should not be incremented until the login has actually failed (like OpenSSH' sftp). 

Problem 2) After successful login with a password, proftpd fails to rest the unsuccessful login count because root privileges have already been revoked before the login completed and the call to AIX's loginsuccess() fails.

### What I Expected/Wanted

I expected that when I sftped to the server I would be prompted for a password. After successfully logging in using a password, I expected the user's unsuccessful login count to be 0 and other  details maintained by the OS in /etc/securoty/lastlog to be updated appropriately. This is all accomplished by a call to AIX's loginsuccess() system call.

### ProFTPD Version and Configuration
```
Compile-time Settings:
  Version: 1.3.7rc1 (git)
  Platform: AIX7 (AIX7_1_4_0) [AIX 1 00C21E024C00]
  Built: Thu Apr 26 2018 14:43:33 MDT
  Built With:
    configure  '--prefix=/opt/proftpd' '--with-modules=mod_sftp' 'CFLAGS=-DPR_TUNABLE_XFER_LOG_MODE=0664 -DAIX7 -I/home/dan/proftpd-git -I/home/dan/proftpd-git/include -maix64' 'LDFLAGS=-dn -L/opt/freeware/lib -L/opt/freeware/lib64 -maix64 -Wl,-bbigtoc'

  CFLAGS: -g2 -DPR_TUNABLE_XFER_LOG_MODE=0664 -DAIX7 -I/home/dan/proftpd-git -I/home/dan/proftpd-git/include -maix64 -Wall -fno-omit-frame-pointer
  LDFLAGS: -L$(top_srcdir)/lib -L$(top_builddir)/lib -dn -L/opt/freeware/lib -L/opt/freeware/lib64 -maix64 -Wl,-bbigtoc
  LIBS:  -lssl -lcrypto -lpam  -lcrypto -lsupp

  Files:
    Configuration File:
      /opt/proftpd/etc/proftpd.conf
[...]

PR #813, opened 2019-07-08
Fix -- AIX mod_sftp unsuccessful login count problems #693

ws0w:
This is a proposed fix for  [Issue #693](https://github.com/proftpd/proftpd/issues/693)   
mod_auth_unix.c is modified to call login_succeeded() or login_failed() right after the call to authenticate(), before PRIVS_ROOT.

mod_sftp/auth.c is modified to make call of pr_cmd_dispatch_phase(pass_cmd, LOG_CMD_ERR, 0) conditional, so that it is not called for each attempt to find a client identity file.


ws0w:
The code changes create other undesired side effects.

PR #814, opened 2019-07-10
Fix -- AIX mod_sftp unsuccessful login count problems #693

ws0w:
This is a proposed fix for [Issue #693](https://github.com/proftpd/proftpd/issues/693)
mod_auth_unix.c is modified to call login_succeeded() right after the call to authenticate(), before PRIVS_ROOT.

mod_sftp/auth.c is modified to make call of pr_cmd_dispatch_phase(pass_cmd, LOG_CMD_ERR, 0) conditional, so that it is not called for each attempt to find a client identity file.

mod_sftp/auth-passwd.c is modified to add calls when the password is bad to:

      pr_response_add_err(R_501, "Login incorrect.");
      pr_cmd_dispatch_phase(pass_cmd, POST_CMD_ERR, 0);
      pr_cmd_dispatch_phase(pass_cmd, LOG_CMD_ERR, 0);

Now, the AIX unsuccessful_login_count is correctly incremented for failed logins using ftps or sftp, and correctly reset to 0 on a successful login using ftps and sftp.

### DIFF
--- a/modules/mod_auth_unix.c
+++ b/modules/mod_auth_unix.c
@@ -35,6 +35,10 @@
 # include <crypt.h>
 #endif
 
+#ifdef HAVE_SYS_AUDIT_H
+# include <sys/audit.h>
+#endif
+
 #ifdef PR_USE_SHADOW
 # ifdef HAVE_SHADOW_H
 #   include <shadow.h>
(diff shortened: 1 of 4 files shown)


modules/mod_auth_unix.c
```diff
@@ -1002,6 +1004,9 @@ MODRET pw_check(cmd_rec *cmd) {
         "AIX authenticate result: %d (msg '%.100s')", res, msg);
 
     } while (reenter != 0);
+    if (res == 0) {
+      login_succeeded(cmd->tmp_pool, user);
```
Castaglia, line 1008: Rather than reaching naughtily into the `mod_auth` symbols, I've instead used a little code duplication in https://github.com/proftpd/proftpd/pull/966, to achieve the same functionality here.

Castaglia:
Thanks for this, it helped point my brain in the right directions, for getting the issue fixed!

PR #837, opened 2019-10-04
- mod_gss does not compile with IPv6 support. The issue has been

Sashan:
introduced in 1.3.6 release.
another option to fix it is to change mod_gss module, so it will use PR_USE_IPV6
instead of its USE_INET6 guard.

### DIFF
--- a/include/compat.h
+++ b/include/compat.h
@@ -50,4 +50,11 @@
 /* The following macros first appeared in 1.3.6rc2. */
 #define _sql_make_cmd			sql_make_cmd
 
+/*
+ * USE_IPV6 is required by mod_gss.
+ */
+#ifdef PR_USE_IPV6
+#define USE_IPV6			1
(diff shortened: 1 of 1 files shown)


Castaglia:
Have you also filed an issue for this, with [`mod_gss`](https://github.com/huaraz/gssmod), so it can be corrected there?  Third-party modules should be fixed to deal with the core ProFTPD code, not the other way around.

Sashan:
> Have you also filed an issue for this, with [`mod_gss`](https://github.com/huaraz/gssmod), so it can be corrected there? Third-party modules should be fixed to deal with the core ProFTPD code, not the other way around.

I've just sent pull request to modgss:
https://github.com/huaraz/gssmod/pull/3

Sashan:
I want to close this pull request. I agree this should rather fixed in gssmod

PR #839, opened 2019-10-15
configure.in: fix --disable-tests handling

trofi:
Toralf found out that `./configure --disable-tests` fails configure
phase in an attempt to find `libcheck`:

```
...
checking for check.h... no
checking for tcase_create in -lcheck... no
configure: error: libcheck support, required for tests, not present -- aborting
```

This happens because the condition lost a `x` guard at test site:
    `if test "$enableval" != xno ; then`

The fix is to add missing guard:
    `if test x"$enableval" != xno ; then`

Reported-by: Toralf Förster
Bug: https://bugs.gentoo.org/697782

### DIFF
--- a/configure
+++ b/configure
@@ -20860,7 +20860,7 @@ ENABLE_TESTS="\"\""
 # Check whether --enable-tests was given.
 if test "${enable_tests+set}" = set; then
   enableval=$enable_tests;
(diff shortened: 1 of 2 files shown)


Castaglia:
Thanks!

Issue #846, opened 2019-10-19
Remote denial-of-service due to issue in network IO handling


stze:
Dear @Castaglia —

As discussed, please find below the description of the ProFTPD 1.3.6 denial-of-service issue. 

#### Issue
Remote unauthenticated denial-of-service in ProFTPD 1.3.6, triggered by a malformed network packet due to an incorrect handling of too long commands.

#### Version
ProFTPD 1.3.6 - ftp://ftp.proftpd.org/distrib/source/proftpd-1.3.6.tar.gz

#### Description
An unauthenticated attacker can trigger an infinite loop by sending a malformed network packet due to an incorrect handling of too long commands. The issue in the network IO causes a child process that handles the client connection to consume 100% CPU. 

If errno is 7 (E2BIG) due to a too long command and the EOF is reached, read() function will always return 0 and will be called indefinitely due to the continue statement:

```c
src/main.c:
...
    if (cmd_buflen < 0) {
      if (errno == E2BIG) {
        /* The client sent a too-long command which was ignored; give
         * them another chance?
         */
        continue;
      }
...
```
#### Impact
Remote unauthenticated denial-of-service 

Cheers
    -Stephan Zeisberg

Castaglia:
Note that this bug is also present in ProFTPD 1.3.7rc1.

Castaglia:
Thanks!

stze:
CVE-2019-18217

PR #847, opened 2019-10-19
Issue #846: Handle the case where a client tries to send too-large co…

Castaglia:
…mmands

in an effort to DoS the server.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -471,6 +471,7 @@ int pr_cmd_read(cmd_rec **res) {
   static long cmd_bufsz = -1;
   static char *cmd_buf = NULL;
   int cmd_buflen;
+  unsigned int too_large_count = 0;
   char *ptr;
 
   if (res == NULL) {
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -496,8 +497,15 @@ int pr_cmd_read(cmd_rec **res) {
     if (cmd_buflen < 0) {
       if (errno == E2BIG) {
         /* The client sent a too-long command which was ignored; give
-         * them another chance?
+         * them a few more chances, with minor delays?
          */
+        too_large_count++;
+        pr_timer_usleep(250 * 1000);
```
Castaglia: This "sleep" helps mitigate the tightly-spinning loop, which is the core of the denial-of-service behavior.  The `too_large_count` limit limits the number of loops before the client is disconnected.

src/netio.c
```diff
@@ -1452,6 +1452,7 @@ int pr_netio_read(pr_netio_stream_t *nstrm, char *buf, size_t buflen,
       }
 
       nstrm->strm_errno = 0;
+      errno = EOF;
```
Castaglia: In the reproduction recipe, the tight loop _should_ have ended sooner, but the lack of properly setting this `errno` value prevented that, further allowing the denial-of-service to occur.  So we properly set `errno` to EOF, to catch another possible avenue for exploitation.

PR #874, opened 2019-12-12
- change makes proftpd to comply with RFC 640.

Sashan:
RFC 640 defines format of multi-line responses. All the foundation
to produce proper multi-line error codes exists already in proftpd
code. The change allows proftpd the right thing.

The RFC 640 defines multi-line responses as follows
'''
       Thus the format for multi-line replies is that the first line
       will begin with the exact required reply code, followed
       immediately by a Hyphen, "-" (also known as Minus), followed
       by text.  The last line will begin with the same code,
       followed immediately by Space <SP>, optionally some text, and
       TELNET <eol>.                                                     10a
    
          For example:
                               123-First line
                               Second line
                                 234 A line beginning with numbers
                               123 The last line                        10a1
    
       The user-process then simply needs to search for the second
       occurrence of the same reply code, followed by <SP> (Space),
       at the beginning of a line, and ignore all intermediary lines.
       If an intermediary line begins with a 3-digit number, the
       Server must pad the front to avoid confusion.                     10b
'''

`proftpd` server deviates from standard, when it fails to open
data connection back to client, when handling `LIST` command.
In this case the command gets dispatched to `mod_ls` module,
which calls `pr_data_open()`. If function fails it puts `425` error
code to the list and returning back to caller. The caller handles
the same error, however with different error code (`450`). The
mod_ls module puts its own error message to the list. Since
list contains two different error messages the code dispatches
back to client on control connection. The legacy ftp client bundled
with `Solaris`/`Illumos` fails to recover from such situation (for
example `lftp` seems to be immune).

The fix is straightforward make sure the caller of `pr_data_open()`
adds it's own message with `425` error code, so proftpd server
can construct multi-line message properly.

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -1960,7 +1960,7 @@ static int dolist(cmd_rec *cmd, const char *opt, const char *resp_code,
       if (pr_data_open(NULL, "file list", PR_NETIO_IO_WR, 0) < 0) {
         int xerrno = errno;
 
-        pr_response_add_err(R_450, "%s: %s", (char *) cmd->argv[0],
+        pr_response_add_err(R_425, "%s: %s", (char *) cmd->argv[0],
           strerror(xerrno));
 
         pr_cmd_set_errno(cmd, xerrno);
@@ -2203,7 +2203,7 @@ static int dolist(cmd_rec *cmd, const char *opt, const char *resp_code,
       if (pr_data_open(NULL, "file list", PR_NETIO_IO_WR, 0) < 0) {
         int xerrno = errno;
 
-        pr_response_add_err(R_450, "%s: %s", (char *) cmd->argv[0], 
+        pr_response_add_err(R_425, "%s: %s", (char *) cmd->argv[0],
           strerror(xerrno));
 
         pr_cmd_set_errno(cmd, xerrno);
@@ -3104,7 +3104,7 @@ MODRET ls_nlst(cmd_rec *cmd) {
             if (pr_data_open(NULL, "file list", PR_NETIO_IO_WR, 0) < 0) {
               int xerrno = errno;
 
-              pr_response_add_err(R_450, "%s: %s", (char *) cmd->argv[0], 
+              pr_response_add_err(R_425, "%s: %s", (char *) cmd->argv[0],
                 strerror(xerrno));
 
               pr_cmd_set_errno(cmd, xerrno);
@@ -3131,7 +3131,7 @@ MODRET ls_nlst(cmd_rec *cmd) {
           if (pr_data_open(NULL, "file list", PR_NETIO_IO_WR, 0) < 0) {
             int xerrno = errno;
 
-            pr_response_add_err(R_450, "%s: %s", (char *) cmd->argv[0],
+            pr_response_add_err(R_425, "%s: %s", (char *) cmd->argv[0],
               strerror(xerrno));
 
             pr_cmd_set_errno(cmd, xerrno);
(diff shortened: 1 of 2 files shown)


Castaglia:
What problem is actually being addressed here?  RFC 640 is _very very old_; RFC 959 is the relevant RFC for FTP implementations.

Given that most modern clients/browsers have no issues with ProFTPD, I am not inclined to change its response format just for some legacy bundled FTP client.

modules/mod_ls.c
```diff
@@ -3368,6 +3368,9 @@ MODRET ls_nlst(cmd_rec *cmd) {
       if (pr_data_open(NULL, "file list", PR_NETIO_IO_WR, 0) < 0) {
         int xerrno = errno;
 
+        pr_response_add_err(R_425, "%s: %s", (char *) cmd->argv[0],
```
Castaglia: This addition makes sense -- but why not keep the existing response codes as they are?  My understanding is that the discrepancy in response codes is what's confusing your legacy FTP client.

Sashan:
> What problem is actually being addressed here? RFC 640 is very very old;
> RFC 959 is the relevant RFC for FTP implementations.

RFC 640 specifies format of multi-line responses. The RFC 959 does not change that.
May be I've just overlooked specification of multi-line responses in RFC 959.
The issue got indeed uncovered in corner case. As I've said `lftp` has no issue
with it certainly. I still would like to check other clients just for curiosity.

> This addition makes sense -- but why not keep the existing response
> codes as they are? My understanding is that the discrepancy in response
> codes is what's confusing your legacy FTP client.

In all those cases the `pr_data_open()` put error message with `425` code to error list.
The caller just adds yet another message to error list here.

The suggested change is based on my brief tour through `pr_response_add_err()` (and
its related code). `pr_response_add()` adds a single error line to `resp_err_list`, which
is managed by `src/response.c` module. According to my understanding each proftpd
module uses `pr_response_add_err()` to error list. As soon as all modules are done with
command processing, the `src/response.c` uses `pr_response_flush()` to send back all
accumulated error messages to client. And there is a logic there, which makes `pr_response_flush()`
to assemble proper multi-line response:
```
193     } else {
194       /* Look for start of multiline */
195       if (resp->next &&
196           (resp->next->num == NULL ||
197            strcmp(resp->num, resp->next->num) == 0)) {
198         RESPONSE_WRITE_NUM_STR(session.c->outstrm, "%s-%s\r\n", resp->num,
199           resp->msg)
200         ml = TRUE;
201         last_numeric = resp->num;
202
203       } else {
204         RESPONSE_WRITE_NUM_STR(session.c->outstrm, "%s %s\r\n", resp->num,
205           resp->msg)
206       }
207     }
```
there is a `strcmp()` at line 197, which checks error code of the first message with error
code in the second message. If codes do match, then function assembles proper multi-line
message. The proper multi-line message as received by client looks as follows:
```
ftp> dir
200 PORT command successful
      
```
note the dash (`-`) at the first response line shown by client. The dash indicates it's a multi-line response. 

While without proposed change we receive two error responses for single `LIST` command.
The data sent by proftpd look as follows:
```
425 Unable to build data connection: Address already in use
450 LIST: Address already in use
```
In this case ftp client just displays the first line and second line stays in buffer. As soon
as client sends yet another command (typically `TYPE I`), then it reads the previously
received `450 LIST...` and deems it's a response to `TYPE I` command, which switches
data channel back to binary mode. The client stays confused by cached error messages.

Another option to fix it would be to remove the `strcmp()` check completely and let 
`pr_response_flush()` to always assemble multi-line message as follows:
```
425-Unable to build data connection: Address already in use
450 LIST: Address already in use
```
The most critical stuff is to put the `-` there. And leaving it out on the last line.

Castaglia:
I think [this PR](https://github.com/proftpd/proftpd/pull/944) addresses this issue more cleanly, by removing the use of the separate/different error codes entirely.

Sashan:
> I think [this PR](https://github.com/proftpd/proftpd/pull/944) addresses this issue more cleanly, by removing the use of the separate/different error codes entirely.

I agree, removing extra error messages will also fix this glitch. thanks.

PR #883, opened 2020-01-15
Also output xml when running tests

dstathis:
This would be nice to have for people who want to ingest test result programmatically.

### DIFF
--- a/tests/api/tests.c
+++ b/tests/api/tests.c
@@ -102,7 +102,9 @@ int main(int argc, char *argv[]) {
    * variable or command-line option.
    */
   srunner_set_log(runner, log_file);
-  srunner_set_xml(runner, xml_file);
+  if (getenv("XML_TEST_OUTPUT")) {
+    srunner_set_xml(runner, xml_file);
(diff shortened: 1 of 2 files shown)


dstathis:
It would definitely be more convenient if it always outputted the xml and I don't think it would cause too much clutter.

I can easily make it require an environment variable, though, if that's what you decide.

Castaglia:
Yes, please do make it require an environment variable, if you wouldn't mind.  Thanks!

tests/api/tests.c
```diff
@@ -102,7 +102,9 @@ int main(int argc, char *argv[]) {
    * variable or command-line option.
    */
   srunner_set_log(runner, log_file);
-  srunner_set_xml(runner, xml_file);
+  if (getenv("XML_TEST_OUTPUT")) {
```
Castaglia: Thanks!  For consistency, would you object to having this be `PR_XML_TEST_OUTPUT`?

Castaglia:
Thanks!

PR #886, opened 2020-01-22
Move definition of recvd_signal_flags for API tests

pghmcfc:
Defining it in tests/api/tests.h results in compile failures with gcc 10 due to there being multiple definitions of the object (see https://www.gnu.org/software/gcc/gcc-10/porting_to.html).

Moving it to the stubs.c file along with all of the other stub objects/functions resolves this issue.


Castaglia:
Thanks!

Issue #882, opened 2020-01-08
mod_sql "named connection already exists" error when using name-based virtual hosts


akunz:
### What I Did

VirtualHost configuration with name-based virtual hosts and MySQL/SQLite authentication. Connecting to default server works fine, but name-based virtual host failed with following errors:

```
2020-01-08 11:22:17,716 proftpd[8] 0.0.0.0 (ftp.example.com[***.***.***.***]): FTP session opened.
2020-01-08 11:22:17,735 proftpd[8] 0.0.0.0 (ftp.example.com[***.***.***.***]): Changing to server 'ProFTPD' (ServerAlias ftp.example.com) due to HOST command
2020-01-08 11:22:17,738 proftpd[8] 0.0.0.0 (ftp.example.com[***.***.***.***]): mod_sql/4.4: unrecoverable backend error: (mod_sql_sqlite/0.4) named connection already exists
2020-01-08 11:22:17,739 proftpd[8] 0.0.0.0 (ftp.example.com[***.***.***.***]): mod_sql/4.4: check the SQLLogFile for more details
2020-01-08 11:22:17,740 proftpd[8] 0.0.0.0 (ftp.example.com[***.***.***.***]): FTP session closed.
```

```
2020-01-08 11:22:17,713 mod_sql/4.4[8]: defaulting to 'sqlite3' backend
2020-01-08 11:22:17,713 mod_sql/4.4[8]: backend module 'mod_sql_sqlite/0.4'
2020-01-08 11:22:17,713 mod_sql/4.4[8]: backend api    'mod_sql_api_v1'
2020-01-08 11:22:17,713 mod_sql/4.4[8]: >>> sql_sess_init
2020-01-08 11:22:17,713 mod_sql/4.4[8]: entering    sqlite cmd_defineconnection
2020-01-08 11:22:17,713 mod_sql/4.4[8]:  name: 'default'
2020-01-08 11:22:17,713 mod_sql/4.4[8]:   dsn: '/usr/local/etc/proftpd.sqlite3'
2020-01-08 11:22:17,713 mod_sql/4.4[8]:   ttl: '0'
2020-01-08 11:22:17,713 mod_sql/4.4[8]: exiting     sqlite cmd_defineconnection
2020-01-08 11:22:17,713 mod_sql/4.4[8]: connection 'default' successfully established
2020-01-08 11:22:17,713 mod_sql/4.4[8]: defaulting to 'sqlite3' backend
[...]

PR #891, opened 2020-01-31
Issue #882: When using name-based virtual hosts, make sure the SQL ba…

Castaglia:
…ckend

modules properly handle already-defined named connections.


Castaglia:
Note that this patch _would_ allow for multiple `SQLNamedConnectInfo` directives, with the same name but different configurations; the first defined `SQLNamedConnectInfo` would be used, and the rest of the "duplicates" silently ignored.

I'll update the PR to add logging for the "named connection already exists" case, for helping detect such configurations -- but without raising an error, such as happens now.

Issue #932, opened 2020-03-14
mod_tls "no matching server found" for first SNI / name-based virtual host


akunz:
### What I Did

When using SNI, first name-based virtual host fails with the following error while other virtual hosts work like expected.

```
2020-03-14 14:57:22,250 proftpd[6]: FTP session opened.
2020-03-14 14:57:22,300 proftpd[6]: mod_tls/2.9: no matching server found for client-sent SNI 'ftp1.example.com', rejecting SSL/TLS connection
2020-03-14 14:57:22,303 proftpd[6]: FTP session closed.
2020-03-14 14:57:22,520 proftpd[7]: FTP session opened.
2020-03-14 14:57:22,536 proftpd[7]: Unknown host 'ftp1.example.com' requested on ***.***.***.***#21, refusing HOST command
2020-03-14 14:57:22,548 proftpd[7]: FTP session closed.
```

### What I Expected/Wanted

 SNI for all name-based virtual hosts.

### ProFTPD Version and Configuration

```
$ proftpd -V
root@c41b5ebffdd5:/# proftpd -V
Compile-time Settings:
  Version: 1.3.7rc4 (git)
  Platform: LINUX [Linux 4.15.0-88-generic x86_64]
  Built: Sat Mar 14 2020 14:48:17 UTC
  Built With:
    configure  '--enable-openssl' '--with-modules=mod_tls'

  CFLAGS: -g2 -O2 -Wall -fno-omit-frame-pointer
  LDFLAGS: -L$(top_srcdir)/lib -L$(top_builddir)/lib  -rdynamic
  LIBS:  -lssl -lcrypto -lssl -lcrypto -lsupp -lcrypt -ldl  -pthread

  Files:
    Configuration File:
      /usr/local/etc/proftpd.conf
    Pid File:
      /usr/local/var/proftpd.pid
    Scoreboard File:
      /usr/local/var/proftpd.scoreboard

  Info:
    + Max supported UID: 4294967295
    + Max supported GID: 4294967295

  Features:
    - Autoshadow support
    - Controls support
    - curses support
    - Developer support
    - DSO support
    + IPv6 support
    + Largefile support
    - Lastlog support
    - Memcache support
    - ncurses support
    - NLS support
    - Redis support
    - Sodium support
    + OpenSSL support (OpenSSL 1.1.1d  10 Sep 2019)
    - PCRE support
    - POSIX ACL support
    + Shadow file support
    + Sendfile support
    + Trace support
    + xattr support

[...]

PR #934, opened 2020-03-14
Issue #932: When handling `ServerAlias` directives, we need to ensure…

Castaglia:
… that

the internal IP-based bindings are created _before_ handling `ServerAlias`.

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD - FTP server daemon
- * Copyright (c) 2001-2016 The ProFTPD Project team
+ * Copyright (c) 2001-2020 The ProFTPD Project team
  *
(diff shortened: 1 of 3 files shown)


Castaglia:
Hopefully https://github.com/proftpd/proftpd/pull/934 addresses this.  The issue looks to be related to the startup sequence and processing of `<VirtualHost>` sections that contain the `ServerAlias` directive.  That processing handles variations like `DefaultServer` being on/off, `Port` being zero (or not), `SocketBindTight` being on/off, _etc_.

Since _your_ configuration does not use `DefaultServer on`, but _does_ have `Port 0`, it encountered this issue (which is not strictly related to TLS SNI), which has been lurking for a while.  Thanks for reporting this!

PR #935, opened 2020-03-15
Issue #932: Make sure to include the target port number for marking a…

Castaglia:
… named

binding as "active", just like when creating the named binding.

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -799,7 +799,8 @@ int pr_ipbind_open(const pr_netaddr_t *addr, unsigned int port,
     for (i = 0; i < ipbind->ib_namebinds->nelts; i++) {
       pr_namebind_t *nb = namebinds[i];
 
-      res = pr_namebind_open(nb->nb_name, nb->nb_server->addr);
+      res = pr_namebind_open(nb->nb_name, nb->nb_server->addr,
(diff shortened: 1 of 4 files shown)


src/bindings.c
```diff
@@ -1131,6 +1134,9 @@ int pr_namebind_open(const char *name, const pr_netaddr_t *addr) {
   }
 
   port = ntohs(pr_netaddr_get_port(addr));
+  if (port == 0) {
+    port = server_port;
+  }
```
Castaglia: This is the key change; this `port == 0` check occurs in `pr_namebind_create()` as well, and probably _should_ have always been done here.

tests/t/lib/ProFTPD/Tests/Modules/mod_tls.pm
```diff
@@ -491,7 +496,10 @@ sub list_tests {
     }
   }
 
-  return testsuite_get_runnable_tests($TESTS);
+#  return testsuite_get_runnable_tests($TESTS);
+  return qw(
+    tls_login_with_sni_port_zero_multi_ports_issue932
+  );
```
Castaglia: Hrm.  Incomplete attempts for local reproduction; I'll have to clean these up before merging this PR.

Issue #940, opened 2020-03-19
Second LIST of the same symlink shows different results


dimsua:
### What I Did
Hello
When logging and make first ls symlink list ok test2 -> test
```
ftp> ls data
227 Entering Passive Mode (95,183,14,53,234,127).
150 Opening ASCII mode data connection for file list
drwxr-xr-x   4 p284174  284174       4096 Mar 19 09:15 .
drwxr-x---  14 99       284174       4096 Apr 26  2019 ..
drwxrwxr-x   3 p284174  284174       4096 Mar 19 09:15 old
-rw-rw-r--   1 p284174  284174          0 Oct 18  2017 .pearrc
drwxrwxr-x   2 p284174  284174       4096 Mar 16 10:25 test
lrwxrwxrwx   1 p284174  284174          4 Mar 16 10:25 test2 -> test
226 Transfer complete
```
But when make second list after first and other listing after test2 listing as dir and not symlink
```
ftp> ls data
227 Entering Passive Mode (95,183,14,53,198,170).
150 Opening ASCII mode data connection for file list
drwxr-xr-x   4 p284174  284174       4096 Mar 19 09:15 .
drwxr-x---  14 99       284174       4096 Apr 26  2019 ..
drwxrwxr-x   3 p284174  284174       4096 Mar 19 09:15 old
-rw-rw-r--   1 p284174  284174          0 Oct 18  2017 .pearrc
drwxrwxr-x   2 p284174  284174       4096 Mar 16 10:25 test
drwxrwxr-x   2 p284174  284174       4096 Mar 16 10:25 test2
226 Transfer complete
```

### What I Expected/Wanted

We want that symlink always listed as symlink, not a dir

### ProFTPD Version and Configuration

```
proftpd -V
Compile-time Settings:
  Version: 1.3.6c (maint)
  Platform: LINUX [Linux 3.10.0-962.3.2.lve1.5.26.1.el7.x86_64 x86_64]
  Built: Thu Mar 19 2020 12:07:42 MSK
  Built With:
[...]

PR #941, opened 2020-03-21
Issue #940: The stashing of the ShowSymlinks value, when pushing/popp…

Castaglia:
…ing the

stack of directories during directory listings, was flawed.

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -2,7 +2,7 @@
  * ProFTPD - FTP server daemon
  * Copyright (c) 1997, 1998 Public Flood Software
  * Copyright (c) 1999, 2000 MacGyver aka Habeeb J. Dihu <macgyver@tos.net>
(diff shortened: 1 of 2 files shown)


modules/mod_ls.c
```diff
@@ -205,9 +205,8 @@ static void push_cwd(char *_cwd, unsigned char *symhold) {
   if (!_cwd)
     _cwd = cwd;
 
-  *symhold = show_symlinks_hold;
   sstrncpy(_cwd, pr_fs_getcwd(), PR_TUNABLE_PATH_MAX + 1);
-  *symhold = list_show_symlinks;
+  *symhold = show_symlinks_hold = list_show_symlinks;
```
Castaglia: This is the key fix: properly stash the current `list_host_symlinks` value in the `show_symlinks_hold` variable.  This has not been being done -- and thus broken -- for quite a while.

PR #942, opened 2020-03-21
Bug #4394: Make sure we use the remote address for the %a LogFormat v…

Castaglia:
…alue, not

the local address.

### DIFF
--- a/src/jot.c
+++ b/src/jot.c
@@ -1251,7 +1251,7 @@ static int resolve_logfmt_id(pool *p, unsigned char logfmt_id,
     case LOGFMT_META_REMOTE_IP: {
       const char *ipstr;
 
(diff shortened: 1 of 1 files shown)


rafasc:
`epel-8` has `proftpd` at version `1.3.6`,  can we have this backported?
cc: @pghmcfc

pghmcfc:
@rafasc, the code in 1.3.6 is a bit different and the handling of the %a format is done in two places, mod_log.c and mod_redis.c. The code in mod_log.c looks OK but mod_redis.c looks to have this issue. Are you using mod_redis?

rafasc:
Please disregard my comment. I was experimenting with multiple versions of proftpd due to another [logging bug](https://github.com/Castaglia/proftpd-mod_vroot/pull/25) and got things mixed up.

Apologies for the noise.

Issue #964, opened 2020-04-10
Unable to load mod_sftp, mod_sql_passwd as shared modules on Alpine


tspicer:
Proftpd is failing to start for new commits. For example, using the latest commit:
```
APP_COMMIT_HASH="8b152e5b3d4e7f069f83fdfc8a60c0cf2d17dbe5" \
```
Will result in proftpd failing to start:
```
bash-5.0# /usr/sbin/proftpd -c /etc/proftpd/proftpd.conf
2020-04-10 05:07:39,396 dc7a538ee417 proftpd[769]: fatal: LoadModule: error loading module 'mod_sql_passwd.c': Invalid argument on line 9 of '/etc/proftpd/prod/modules.conf'
2020-04-10 05:07:39,397 dc7a538ee417 proftpd[769]: warning: unable to include '/etc/proftpd/prod/modules.conf': Operation not permitted
2020-04-10 05:07:39,397 dc7a538ee417 proftpd[769]: fatal: LoadModule: error loading module 'mod_sftp.c': Invalid argument on line 5 of '/etc/proftpd/proftpd.conf'
```
However, setting to an older commit:
```
APP_COMMIT_HASH="1e75710d43edb07aea34a200c839d67de698bba1" \
```
It will work OK, no issues. I was doing some random update tests and came across this. 

```
proftpd -V
Compile-time Settings:
  Version: 1.3.7rc4 (git)
  Platform: LINUX [Linux 4.19.76-linuxkit x86_64]
  Built: Fri Apr 10 2020 05:00:51 UTC
  Built With:
[...]

PR #965, opened 2020-04-10
Issue #964: Modules that use the bcrypt code in the `libsupp` library…

Castaglia:
… need

to properly declare that usage for the linker.

Otherwise, strict platforms like Alpine will not be able to properly
resolve all of the symbols, and will fail to load such shared modules.

### DIFF
--- a/src/auth.c
+++ b/src/auth.c
@@ -2,7 +2,7 @@
  * ProFTPD - FTP server daemon
  * Copyright (c) 1997, 1998 Public Flood Software
  * Copyright (c) 1999, 2000 MacGyver aka Habeeb J. Dihu <macgyver@tos.net>
- * Copyright (c) 2001-2017 The ProFTPD Project team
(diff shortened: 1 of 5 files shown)


src/auth.c
```diff
@@ -2433,6 +2434,29 @@ size_t pr_auth_set_max_password_len(pool *p, size_t len) {
...
+
+  if (p == NULL ||
+      key == NULL ||
+      salt == NULL ||
+      hashed_len == NULL) {
+    errno = EINVAL;
+    return NULL;
+  }
+
+  if (bcrypt_hashpass(key, salt, hashed, sizeof(hashed)) != 0) {
```
Castaglia, line 2449: By using the `bcrypt_hashpass` symbol in the core code, we pull that symbol (and all of its underlying Blowfish functions) into the core code by the linker.  And that means that shared modules can find those symbols when they are loaded.  Sneaky.

PR #1008, opened 2020-05-02
Address a use-after-free bug which only manifests when the

Castaglia:
`--enable-devel=nodaemon:nofork` configure/compile options are used.


Castaglia:
This bug occurs because `free_pools()` is called (but only when `PR_DEVEL_NO_DAEMON` is defined), and then `PRIVS_ROOT` is called (which fails, triggering logging, which triggers an event), but only when `PR_DEVEL_NO_DAEMON` is defined.

PR #1009, opened 2020-05-04
Bug #4185: Create directive for tuning the presence/absence of fields…

Castaglia:
… in the

default ProFTPD syslog/module logging.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -55,6 +55,7 @@
   signal "ext-info-c".
 - Issue 983 - Use re-entrant versions of time functions where available.
 - Bug 4398 - Handle zero-length SFTP WRITE requests without error.
+- Bug 4185 - Implement options for tuning fields used in syslog/module logging.
 
 1.3.7rc3 - Released 20-Feb-2020
 --------------------------------
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -30,6 +30,12 @@ ChangeLog files.
       mechanisms to use, when using the LDAPBindDN to bind to the LDAP
       server.  See doc/contrib/mod_ldap.html#LDAPUseSASL for details.
 
+    LogOptions
+
+      The LogOptions directive is used to modify the default logging format
+      for ProFTPD syslog, debug, and module logging.  See
+      doc/modules/mod_log.html#LogOptions for more information.
+
     SQLKeepAlive
 
       The SQLKeepAlive directive configures a periodic "keepalive" query
@@ -43,6 +49,10 @@ ChangeLog files.
       file, client certificate and key files, SSL ciphers, and verification
       policies for LDAP connections.  See doc/contrib/mod_ldap.html#LDAPServer
       for more details.
+
+    TraceOptions
+      The TraceOptions directive now supports a "Timestamp" option, for
+      disabling inclusion of timestamps in Trace logs.
 
   + Developer notes
 
--- a/doc/modules/mod_core.html
+++ b/doc/modules/mod_core.html
@@ -2988,9 +2988,11 @@ adding/remove certain fields of data.
 The options supported by the <code>TraceOptions</code> directive are:
 <ul>
   <li>ConnIPs
+  <li>Timestamp
   <li>TimestampMillis
 </ul>
-These options are all <em>disabled</em> by default.
+The <code>Timestamp</code> option is <em>enabled</em> by default; the others
+are <em>disabled</em> by default.
 
 <p>
 To enable an option, preface the option name with a '+' (plus) character;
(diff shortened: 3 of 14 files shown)


pghmcfc:
I was about to request an option for dropping the `proftpd[pid] ` too...

This is with systemd and `LogOptions -Timestamp -Hostname`
```
May 04 14:04:32 ftpserver.city-fan.org systemd[1]: Starting ProFTPD FTP Server...
May 04 14:04:32 ftpserver.city-fan.org proftpd[2187247]: Checking syntax of configuration file
May 04 14:04:32 ftpserver.city-fan.org systemd[1]: Started ProFTPD FTP Server.
May 04 14:04:32 ftpserver.city-fan.org proftpd[2187248]: proftpd[2187248] ftpserver.city-fan.org: ProFTPD 1.3.7rc3 (devel) (built Mon May 4 2020 13:59:22 BST) standalone mode STARTUP
May 04 14:05:17 ftpserver.city-fan.org proftpd[2187248]: proftpd[2187261] ftpserver.city-fan.org (ftpclient.city-fan.org[2001:...::24]): FTP session opened.
May 04 14:05:24 ftpserver.city-fan.org proftpd[2187248]: proftpd[2187261] ftpserver.city-fan.org (ftpclient.city-fan.org[2001:...::24]): USER paul: Login successful.
May 04 14:05:28 ftpserver.city-fan.org proftpd[2187248]: proftpd[2187261] ftpserver.city-fan.org (ftpclient.city-fan.org[2001:...::24]): FTP session closed.
```
... but I noticed that `proftpd` is logging the PID of the process serving the client and `systemd` is just logging the pid of the master process so they're both useful values, albeit slightly confusing at first glance.

Castaglia:
Thanks for trying out the PR, @pghmcfc .  So in your opinion, will this meet your current `systemd` needs/use case?

pghmcfc:
It's OK for me but I asked the reporter of my original issue (https://bugzilla.redhat.com/show_bug.cgi?id=1808989) to comment too.

pghmcfc:
I think it would be a good idea to add a `-Pid` option to drop the "proftpd[PID]: "  part of the log message. Whilst it does contain information that `systemd `doesn't provide, I can certainly envisage some people not wanting to see it given that `systemd` does provide the main process pid.

Castaglia:
@pghmcfc unfortunately, if the session-specific PID is omitted from the log message, there will not be anything to differentiate the multiple sessions from each other in the `systemd` logs, which I feel will create more confusion and work later, for folks trying to provide `systemd` logs for debugging issues -- and we won't be able to tell one session from another.  The master PID is not sufficient, IMHO.

pghmcfc:
@Castaglia I see where you're coming from there. Maybe if that PID could be made to look less like a duplicate it might look better? Something like changing the "proftpd" to "master" or "child" depending on which process logged the message?

Castaglia:
That approach would work - but only because `systemd` is using a _different_ label, _i.e._ "proftpd", for its logging of that master PID.  This is getting into the area where `systemd`-specific details are imposing constraints on ProFTPD logging even more -- and there _are_ other process managers than `systemd`.

Castaglia:
One possibility is to change the "proftpd[$PID]" permanently, in the ProFTPD logging, to use _e.g._ "daemon[$PID]" and "session[$PID]", I suppose.  I expect that this would cause compatibility issues for custom ProFTPD log parsers out there.

pghmcfc:
Unless there was a `LogOption` to leave it as "proftpd[$PID]"...

Castaglia:
True.  Hmm.  Maybe something like this?
```
LogOptions -Timestamp -Hostname +RoleBasedProcessLabels
```
where the "role-based" process labels would be "daemon" and "session", _vs_ "proftpd" for both.  Think that'd be understandable, clear?

pghmcfc:
Yes, that looks OK to me.

Castaglia:
@pghmcfc PR now updated with the suggested `RoleBasedProcessLabels` option, for your testing.

pghmcfc:
Logging now looks like this:
```
May  9 13:15:39 ftpserver proftpd[2544603]: Checking syntax of configuration file
May  9 13:15:39 ftpserver proftpd[2544604]: daemon[2544604] ftpserver.city-fan.org: ProFTPD 1.3.7rc3 (devel) (built Sat May 9 2020 13:10:04 BST) standalone mode STARTUP
May  9 13:16:33 ftpserver proftpd[2544604]: session[2544699] ftpserver.city-fan.org (ftpclient.city-fan.org[2001:...::24]): FTP session opened.
May  9 13:16:41 ftpserver proftpd[2544604]: session[2544699] ftpserver.city-fan.org (ftpclient.city-fan.org[2001:...::24]): USER paul: Login successful.
May  9 13:16:50 ftpserver proftpd[2544604]: session[2544699] ftpserver.city-fan.org (ftpclient.city-fan.org[2001:...::24]): FTP session closed.
```
I don't think there's much can be done to improve that.

PR #1015, opened 2020-05-14
PID file should never be world-writable

d-schramm:
The PID file should be created with restrictive permissions 0664, not 0666 & umask (set via the Umask directive). Since version 1.19.3 start-stop-daemon refuses to start/stop the daemon if the PID file has world-write permission.

### DIFF
--- a/src/pidfile.c
+++ b/src/pidfile.c
@@ -52,10 +52,13 @@ int pr_pidfile_set(const char *path) {
 int pr_pidfile_write(void) {
   int xerrno;
   FILE *fh = NULL;
+  mode_t prev_mask;
 
   PRIVS_ROOT
+  prev_mask = umask(0113);
   fh = fopen(pidfile_path, "w");
   xerrno = errno;
+  umask(prev_mask);
   PRIVS_RELINQUISH
 
   if (fh == NULL) {


Castaglia:
Thanks for the PR!  I've filed a ticket for this issue; see https://github.com/proftpd/proftpd/issues/1018.

I may try to address this a slightly different way; will see how it goes.

Castaglia:
@d-schramm when you encounter this error, what does this show for you, in your shell:
```
umask
```
I'm curious to see what your umask was, such that ProFTPD's `PidFile` had world-writable permissions.

d-schramm:
@Castaglia The configured Umask is 0000. We have a special scenario where newly created files have to be writable by other because the user/owner is reserved for "superuser" access and we cannot use the group either. The shell says 0022 for normal users and root -- but that seems to have no effect on the PID file that proftpd creates.

Castaglia:
@d-schramm thanks for the context.  I've created a slightly different [PR](https://github.com/proftpd/proftpd/pull/1019) for this issue -- could you try it out in your scenario, make sure it solves the issue for you?

Castaglia:
Thanks for this PR!  I've now merged the other fix for this to the master branch, and backported it to the 1.3.6 branch.

PR #1028, opened 2020-05-31
contrib/mod_ldap.c: fix SIGSEGV in mod_ldap:ldap_mod_init () and add gcc-specific sentinel checks

trofi:


Castaglia:
Excellent.  Thank you!

PR #1029, opened 2020-06-01
Save errno across call to pr_trace_msg()

pghmcfc:
Since `pr_trace_msg()` can cause `errno` to be set, but we're actually interested in the `errno` values resulting from name lookups, save the value of `errno` across a call to `pr_trace_msg()`.
This PR is a result of tracking down a repeatable failure in the `netaddr` API test when run on Fedora 19 (and only on that OS release for some reason). The same test (name lookup for `134.289.999.0`) did not fail with 1.3.7rc3. With 1.3.7rc4 on Fedora 19, the `pr_trace_msg()` call sets `errno` to `EPERM` and, since the test expects `errno` to be `ENOENT`, it fails.
I haven't figured out why this only happens on Fedora 19 but I suspect that there may be a lot of places where tracing might affect `errno`.

### DIFF
--- a/src/netaddr.c
+++ b/src/netaddr.c
@@ -590,7 +590,7 @@ static pr_netaddr_t *get_addr_by_ip(pool *p, const char *name,
 static pr_netaddr_t *get_addr_by_name(pool *p, const char *name,
     array_header **addrs) {
   pr_netaddr_t *na = NULL;
-  int res;
+  int res, xerrno;
   struct addrinfo hints, *info = NULL;
 
   memset(&hints, 0, sizeof(hints));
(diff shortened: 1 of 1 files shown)


src/netaddr.c
```diff
@@ -590,7 +590,7 @@ static pr_netaddr_t *get_addr_by_ip(pool *p, const char *name,
 static pr_netaddr_t *get_addr_by_name(pool *p, const char *name,
     array_header **addrs) {
   pr_netaddr_t *na = NULL;
-  int res;
+  int res, save_errno;
```
Castaglia, line 593: Style nit:  I tend to use a variable name of `xerrno` for just this purpose; you'll see lots of `xerrno` variables in the ProFTPD codebase.

pghmcfc:
Yes, I saw that - in fact `xerrno` is declared twice in limited scopes later in the same function, so I wanted to avoid re-using the same name so as to keep the diff minimal and avoid confusion. I've now changed it to use the same `xerrno` throughout the function.

Castaglia:
Thanks!

PR #1030, opened 2020-06-02
More protection of errno from pr_trace_msg()

pghmcfc:
This one is similar to #1029 but manifested in Fedora 13 once I'd applied #1026 to support building with OpenSSL  1.1.0. I've basically restored the value of `errno` to what it was before a couple of calls to `pr_trace_msg()`.


Castaglia:
Thanks!  Preserving `errno` values appropriately is definitely something I'm contending with more of late -- I appreciate the help with this!

pghmcfc:
I wondered whether it might be appropriate to save/restore `errno` at the start and end of `pr_trace_msg()` itself but then I guess there could be cases where you want to know why tracing resulted in an error.

PR #1075, opened 2020-07-30
netaddr API test: check for EAGAIN too

pghmcfc:
If the test suite is run in an environment with networking disabled, `getaddrinfo()` can set `errno` to `EAGAIN` when doing lookups of things that don't look like valid IP addresses.

This cropped up in current Fedora Rawhide where the IPv6 lookups (but not the IPv4 lookups) of "134.289.999.0" and "foo.bar.castaglia.example.com" set `errno` to `EAGAIN` when the API tests were run in a network-disabled build environment.


Castaglia:
Thanks!

Issue #1084, opened 2020-08-12
Support configuring TLS 1.3 cipher suites


rlaager:
ProFTPD should have an option for configuring TLS 1.3 cipher suites. This is different from configuring TLS 1.2 ciphers. That is, OpenSSL does not using the traditional cipher strings to configure TLS 1.3 cipher suites but instead expects them to be configured separately.

For example, my Apache configuration looks like this:
```
SSLCipherSuite ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384
SSLCipherSuite TLSv1.3 TLS_AES_128_GCM_SHA256:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_256_GCM_SHA384
```

I'm asking for an equivalent to the latter option.

This was requested by someone else previously in #744, but it was mixed in with another request which became the main focus.

Castaglia:
@rlaager Thanks for the report.  Care to try out https://github.com/proftpd/proftpd/pull/1087, verify that it works as you expect/need?

Castaglia:
This should now be fixed in master.  Thanks!

PR #1087, opened 2020-08-14
Issue #1084: Modify the `TLSCipherSuite` directive to support configu…

Castaglia:
…ration of

TLSv1.3 ciphersuites.

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -562,6 +562,11 @@ static unsigned int tls_sscn_mode = TLS_SSCN_MODE_SERVER;
 #define TLS_X509V3_TLS_FEAT_STATUS_REQUEST_V2	{ 0x30, 0x03, 0x02, 0x01, 0x17 }
 
 static char *tls_cipher_suite = NULL;
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L && \
+    defined(TLS1_3_VERSION)
+static char *tlsv13_cipher_suite = NULL;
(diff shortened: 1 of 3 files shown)


contrib/mod_tls.c
```diff
@@ -13367,32 +13392,64 @@ MODRET set_tlscertchain(cmd_rec *cmd) {
...
+      CONF_ERROR(cmd, pstrcat(cmd->tmp_pool,
+        "unknown/unsupported protocol specifier: ", protocol_text, NULL));
+    }
+
+    ciphersuite = cmd->argv[2];
+  }
+
+  c = add_config_param(cmd->argv[0], 2, NULL, NULL);
+
+  if (protocol != TLS_PROTO_TLS_V1_3) {
```
rlaager, line 13425: This block pstrdup()s ciphersuite, but only in the !TLSv1.3 case. It needs to pstrdup() in both cases, like this:
```
  if (protocol == TLS_PROTO_TLS_V1_3) {
    ciphersuite = pstrdup(c->pool, ciphersuite);

  } else {
    if (strncmp(ciphersuite, "PROFILE=", 8) == 0) {
      ciphersuite = pstrdup(c->pool, ciphersuite);

    } else {
      ciphersuite = pstrcat(c->pool, "!EXPORT:", ciphersuite, NULL);
    }
  }
```

Castaglia: Ah, yes.  Good catch!  Now fixed.

Issue #1080, opened 2020-08-06
SSH Exchange Failure: payload len (X bytes) exceeds max payload len (Y), ignoring payload


jdurhamcgi:
### What I Did

Build proftpd 1.3.7a from source using OpenSSL 1.0.2u with FIPS 2.0.16 module. The ProFTPD configuration itself was migrated from an existing, working implementation using 1.3.6rc2.

### What I Expected/Wanted

I expect to be able to log in as a configured user, but am receiving an error during the SSH exchange prior to use authentication.

### ProFTPD Version and Configuration
**proftpd -V**
```bash
  Version: 1.3.7a (maint)
  Platform: LINUX [Linux 3.10.0-1127.18.2.el7.x86_64 x86_64]
  Built: Thu Aug 6 2020 10:10:58 MDT
  Built With:
    configure  '--prefix=/usr/local/proftpd/config' '--enable-dso' '--enable-ctrls' '--enable-openssl' '--with-includes=/usr/local/ssl/include' '--with-libraries=/usr/local/ssl/lib' '--with-modules=mod_sftp'

  CFLAGS: -g2 -O2 -Wall -fno-omit-frame-pointer
  LDFLAGS: -L$(top_srcdir)/lib -L$(top_builddir)/lib  -rdynamic -L/usr/local/ssl/lib
  LIBS:  -lssl -lcrypto -lpam  -lcrypto -lsupp -lcrypt -ldl 

  Files:
    Configuration File:
      /usr/local/proftpd/config/etc/proftpd.conf
    Pid File:
      /usr/local/proftpd/config/var/proftpd.pid
    Scoreboard File:
      /usr/local/proftpd/config/var/proftpd.scoreboard
    Header Directory:
      /usr/local/proftpd/config/include/proftpd
    Shared Module Directory:
      /usr/local/proftpd/config/libexec

  Info:
    + Max supported UID: 4294967295
    + Max supported GID: 4294967295

  Features:
    - Autoshadow support
    + Controls support
    - curses support
    - Developer support
    + DSO support
    + IPv6 support
    + Largefile support
    - Lastlog support
    - Memcache support
    - ncurses support
    - NLS support
    + OpenSSL support (OpenSSL 1.0.2u  20 Dec 2019)
    - PCRE support
    - POSIX ACL support
    - Redis support
    + Sendfile support
    + Shadow file support
    - Sodium support
    + Trace support
    + xattr support

  Tunable Options:
[...]

PR #1096, opened 2020-09-06
Issue #1080: Improve/fix ProFTPD's autoconf detection of a FIPS-enabl…

Castaglia:
…ed OpenSSL installation.

### DIFF
--- a/configure
+++ b/configure
@@ -24125,7 +24125,6 @@ else
 /* end confdefs.h.  */
 
       #include <openssl/crypto.h>
-      #include <openssl/fips.h>
       int main(int argc, char *argv[]) {
       # ifdef OPENSSL_FIPS
         return 0;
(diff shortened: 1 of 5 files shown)


configure
```diff
@@ -24125,7 +24125,6 @@ else
 /* end confdefs.h.  */
 
       #include <openssl/crypto.h>
-      #include <openssl/fips.h>
```
Castaglia, line 24128: This file is not present for _e.g._ OpenSSL 1.0.2 installations, and is not necessary for FIPS detection.

src/main.c
```diff
@@ -43,7 +43,7 @@
 #include "privs.h"
 
 #ifdef PR_USE_OPENSSL
-# include <openssl/opensslv.h>
+# include <openssl/crypto.h>
```
Castaglia, line 46: The _ order_ in which OpenSSL pulls in its header files is important; getting it wrong (as we were) would not properly set the build-time defines (such as `#define OPENSSL_FIPS`), and thus we might not properly detect FIPS support.

Issue #1076, opened 2020-07-31
TLS client-initiated renegotiations are supported unexpectedly


bratkartoffel:
### What I Did

~~The latest release does not handle various tls security related settings as defined by documentation.~~ (Update: https://github.com/proftpd/proftpd/issues/1076#issuecomment-689015778)
The latest release can be DOS'ed as client-initiated renegotiations are allowed.

### What I Expected/Wanted

Client initiated renegotiations should be disabled ~~and server cipher suite preffered order should be used.~~ (Update: https://github.com/proftpd/proftpd/issues/1076#issuecomment-689015778)

### ProFTPD Version and Configuration

proftpd 1.3.7a from alpine edge:
```
Compile-time Settings:
  Version: 1.3.7a (maint)
  Platform: LINUX [Linux 4.15.0-112-generic x86_64]
  Built: Sat Jul 25 2020 21:12:02 UTC
  Built With:
[...]

PR #1098, opened 2020-09-07
Issue #1076: If mod_tls-initiated renegotiations are disabled via "TL…

Castaglia:
…SRenegotiate none", then set the flag to reject client-initiated renegotiations.


contrib/mod_tls.c
```diff
@@ -17775,27 +17788,22 @@ static int tls_ctx_set_next_protocol(SSL_CTX *ctx) {
 
 static int tls_ctx_set_options(SSL_CTX *ctx) {
 #if OPENSSL_VERSION_NUMBER > 0x009080cfL
+  SSL_CTX_clear_options(ctx, SSL_OP_CIPHER_SERVER_PREFERENCE);
```
bratkartoffel, line 17791: The `SSL_OP_CIPHER_SERVER_PREFERENCE` option is used here, although there is an if-guard at line 17804.
I think the if-guard should be applied here too

Castaglia: Fair enough.  Done.

Issue #1112, opened 2020-09-18
bcrypt passwords dont work with mod_sql


pznamensky:
### What I Did
I've configured ProFTPD to use bcrypt passwords from mysql db:
```
LoadModule mod_sql.c
LoadModule mod_sql_passwd.c
LoadModule mod_sql_mysql.c
SQLBackend              mysql
SQLAuthenticate         users groups
SQLAuthTypes Bcrypt
SQLPasswordEncoding             base64
SQLConnectInfo  ftp@mysql-srv proftpd password
SQLUserInfo     ftpuser userid passwd uid gid homedir shell
SQLGroupInfo    ftpgroup groupname gid members
```
Create tables and insert users:
```
CREATE TABLE ftpgroup (
groupname varchar(16) NOT NULL default '',
gid smallint(6) NOT NULL default '5500',
members varchar(16) NOT NULL default '',
KEY groupname (groupname)
) COMMENT='ProFTP group table';


CREATE TABLE ftpuser (
id int(10) unsigned NOT NULL auto_increment,
userid varchar(32) NOT NULL default '',
passwd varchar(128) NOT NULL default '',
uid smallint(6) NOT NULL default '5500',
gid smallint(6) NOT NULL default '5500',
homedir varchar(255) NOT NULL default '',
shell varchar(16) NOT NULL default '/sbin/nologin',
count int(11) NOT NULL default '0',
accessed datetime NOT NULL default '0000-00-00 00:00:00',
modified datetime NOT NULL default '0000-00-00 00:00:00',
PRIMARY KEY (id),
UNIQUE KEY userid (userid)
) COMMENT='ProFTP user table';
```
`$2b$12$IoFxXvbRQUKssPqFacJFFuZl1KXl5ULppqf0aLFjwCFnLRh3NbYSG` stands for `password`
```
INSERT INTO `ftpgroup` (`groupname`, `gid`, `members`) VALUES ('ftp', 50, 'yo');
INSERT INTO `ftpuser` (`userid`, `passwd`, `uid`, `gid`, `homedir`, `shell`, `count`) VALUES ('yo', '$2b$12$IoFxXvbRQUKssPqFacJFFuZl1KXl5ULppqf0aLFjwCFnLRh3NbYSG', 14, 50, '/var/proftp/', '/sbin/nologin', 0);
```
Unfortunately ProFTPD says that password is wrong. 

### What I Expected/Wanted
I expected that I can store bcrypted passwords in a database.

[...]

PR #1116, opened 2020-09-24
Issue #1112: Prevent segfaults on session close, when TLS is used.

Castaglia:


contrib/mod_tls.c
```diff
@@ -11929,39 +11929,31 @@ static int tls_netio_close_cb(pr_netio_stream_t *nstrm) {
...
-        nstrm->strm_mode == PR_NETIO_IO_WR) {
-      if (tls_ctrl_rd_nstrm != NULL) {
-        pr_table_remove(tls_ctrl_rd_nstrm->notes, TLS_NETIO_NOTE, NULL);
+    if (nstrm->strm_type == PR_NETIO_STRM_CTRL) {
+      if (nstrm->strm_mode == PR_NETIO_IO_RD) {
         tls_ctrl_rd_nstrm = NULL;
-      }
 
-      if (tls_ctrl_wr_nstrm != NULL) {
-        pr_table_remove(tls_ctrl_wr_nstrm->notes, TLS_NETIO_NOTE, NULL);
```
Castaglia, line 11940: We don't actually need these `pr_table_remove()` calls here, since the `nstrm->notes` table is automatically cleared, when the `nstrm` pool is destroyed, in the calling `pr_netio_close()` function.

The actual segfault occurred because here, when dealing with the `PR_NETIO_IO_WR` nstrm object, we _also_ try to remove a key from the _read_ `nstrm->notes` table -- and that was already destroyed.  Hence the segfault, due to referencing a stale pointer.

PR #1137, opened 2020-10-23
Add logging in after dir_check() for <Limit> configuration

byshen:
Add the same logging at debug level for PWD, HOST, NOOP commands before sending the error msg.


Castaglia:
Thanks!

PR #1138, opened 2020-10-24
Add error message when deleting fails in opendir

byshen:
Create log messages to avoid deleting fails silently.

### DIFF
--- a/contrib/mod_site_misc.c
+++ b/contrib/mod_site_misc.c
@@ -181,10 +181,16 @@ static int site_misc_delete_dir(pool *p, const char *dir) {
   int res;
   cmd_rec *cmd;
   pool *sub_pool;
+  int xerrno;
 
   dirh = pr_fsio_opendir(dir);
-  if (dirh == NULL)
+  xerrno = errno;
+  
+  if (dirh == NULL) {
+    pr_log_debug(DEBUG2, MOD_SITE_MISC_VERSION ": error deleting '%s': %s",
+      dir, strerror(xerrno));
     return -1;
+  }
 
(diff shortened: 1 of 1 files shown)


contrib/mod_site_misc.c
```diff
@@ -183,8 +183,11 @@ static int site_misc_delete_dir(pool *p, const char *dir) {
   pool *sub_pool;
 
   dirh = pr_fsio_opendir(dir);
-  if (dirh == NULL)
+  if (dirh == NULL) {
+    pr_log_debug(DEBUG2, MOD_SITE_MISC_VERSION ": error deleting '%s': %s",
+      dir, strerror(xerrno));
```
Castaglia, line 188: There is no `xerrno` variable declared in this function.  Hence the CI build failures.

For this to work as you'd like, we'd need _e.g._:
```
  pool *sub_pool;
  int xerrno;

  dirh = pr_fsio_opendir(dir);
  xerrno = errno;

  if (dirh == NULL) {
    pr_log_debug(DEBUG2, MOD_SITE_MISC_VERSION ": error reading directory '%s': %s",
      dir, strerror(xerrno));
    errno = xerrno;
    return -1;
  }
```

byshen: Sorry for the mistake, updated it.

Castaglia: Looks like there were some other build issues; I've now added/fixed the above code to the master branch.  Thanks!

byshen: I see the new commit, thanks. I will close it now.

PR #1141, opened 2020-10-26
Add log message when modifying fact fails

byshen:
Add debug log message for modifying fact fails, before sending the error to the client.

### DIFF
--- a/contrib/mod_site_misc.c
+++ b/contrib/mod_site_misc.c
@@ -1029,6 +1029,8 @@ MODRET site_misc_utime_mtime(cmd_rec *cmd) {
   if (pr_fsio_utimes_with_root(path, tvs) < 0) {
     int xerrno = errno;
 
+    pr_log_debug(DEBUG2, MOD_SITE_MISC_VERSION
+        ": error modifying fact for '%s': %s", path, strerror(xerrno));
     pr_response_add_err(R_550, "%s: %s", cmd->arg, strerror(xerrno));
 
     errno = xerrno;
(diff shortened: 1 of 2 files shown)


contrib/mod_site_misc.c
```diff
@@ -1029,6 +1029,8 @@ MODRET site_misc_utime_mtime(cmd_rec *cmd) {
   if (pr_fsio_utimes_with_root(path, tvs) < 0) {
     int xerrno = errno;
 
+    pr_log_debug(DEBUG2, MOD_SITE_MISC_VERSION
+        ": error modifying fact for '%s': %s", path, strerror(xerrno));
```
Castaglia, line 1033: Using "fact" here is generic; perhaps we can better indicate what was being changed in the log message, _e.g._:
```
"error modifying timestamps for '%s': %s", path, strerror(xerrno))
```

contrib/mod_site_misc.c
```diff
@@ -1211,6 +1213,8 @@ MODRET site_misc_utime_atime_mtime_ctime(cmd_rec *cmd) {
...
```
Castaglia, line 1217: Similarly here, perhaps use "timestamps" rather than "fact", to be more specific?

modules/mod_facts.c
```diff
@@ -905,7 +905,7 @@ static int facts_modify_mtime(pool *p, const char *path, char *timestamp) {
     int xerrno = errno;
 
     pr_log_debug(DEBUG2, MOD_FACTS_VERSION
-      ": error modifying modify fact for '%s': %s", path, strerror(xerrno));
+      ": error modifying fact for '%s': %s", path, strerror(xerrno));
```
Castaglia, line 908: Thanks!

contrib/mod_site_misc.c
byshen: I change "fact" to be "timestamps". It is indeed more specific, thanks!

contrib/mod_site_misc.c
byshen: Also changed it here, thanks.

PR #1142, opened 2020-10-26
Add log message in site_misc_symlink()

byshen:
Create debug log messages before sending error to the clients.


Castaglia:
Thanks!

PR #1143, opened 2020-10-27
Add log message in handling MLST when facts_mlinfo_get() fails

byshen:
Create the debug message before sending the error.

### DIFF
--- a/modules/mod_facts.c
+++ b/modules/mod_facts.c
@@ -1766,6 +1766,9 @@ MODRET facts_mlst(cmd_rec *cmd) {
   pr_fs_clear_cache2(decoded_path);
   if (facts_mlinfo_get(&info, decoded_path, decoded_path, flags,
       fake_user, fake_uid, fake_group, fake_gid, fake_mode) < 0) {
+
+    pr_log_debug(DEBUG3, MOD_FACTS_VERSION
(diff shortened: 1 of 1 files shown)


modules/mod_facts.c
```diff
@@ -1766,6 +1766,9 @@ MODRET facts_mlst(cmd_rec *cmd) {
   pr_fs_clear_cache2(decoded_path);
   if (facts_mlinfo_get(&info, decoded_path, decoded_path, flags,
       fake_user, fake_uid, fake_group, fake_gid, fake_mode) < 0) {
+
+    pr_log_debug(DEBUG3, MOD_FACTS_VERSION
+      ": MLST: unable to get info for '%s': %s", decoded_path, strerror(errno));
```
Castaglia, line 1771: I'm not sure this is needed; all of the cases where `facts_mlinfo_get()` would return -1 already have associated logging for those error cases.

byshen: I see. I found `facts_mlinfo_get()` has been logged in handling MLSD(`facts_mlsd()`), but this `MLST` was not added, so I add it here. Maybe the log is more clear with one extra log.  If this is not needed, feel free to close it. Thank you.

Castaglia: Good point about the logging case for `facts_mlinfo_get()` and MLSD; I've now removed that redundant logging.

PR #1147, opened 2020-11-02
Add log message for pr_fsio_stat() failures

byshen:
Add a bunch of log messages and before sending the clients the error msg, when `pr_fsio_stat()` or `pr_fsio_fstat()` fails

### DIFF
--- a/contrib/mod_digest.c
+++ b/contrib/mod_digest.c
@@ -1872,7 +1872,7 @@ static modret_t *digest_xcmd(cmd_rec *cmd, unsigned long algo) {
     int xerrno = errno;
 
     pr_log_debug(DEBUG8, MOD_DIGEST_VERSION
-      ": unable to stat %s: %s", path, strerror(xerrno));
+      ": error checking %s: %s", path, strerror(xerrno));
     pr_response_add_err(R_550, "%s: %s", orig_path, strerror(xerrno));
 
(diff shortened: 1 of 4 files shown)


byshen:
@Castaglia Hi TJ, could you have a look at your convenience? Any comments are appreciated! Thanks!

contrib/mod_digest.c
```diff
@@ -1871,6 +1871,8 @@ static modret_t *digest_xcmd(cmd_rec *cmd, unsigned long algo) {
   if (pr_fsio_stat(path, &st) < 0) {
     int xerrno = errno;
 
+    pr_log_debug(DEBUG8, MOD_DIGEST_VERSION
+      ": unable to stat %s: %s", path, strerror(xerrno));
```
Castaglia, line 1875: I try to avoid using syscall-specific terminology in log messages; the function `stat(2)` means something to the developer, but not necessarily to the sysadmin/user _reading_ the log messages.  That's why in _e.g._ `mod_site_misc` I use something like:
```
"error checking '%s': %s", path, strerror(xerrno));
```
I'd recommend doing the same for all of these cases you want to add.

byshen: Make sense. I have changed the log messages in the new commit.  Thank you.

Castaglia:
Thanks!

Issue #1111, opened 2020-09-15
"Corrupted MAC on input" errors using SFTP umac-64@openssh.com digest


cedef:
### What I Did

Using:
- latest proftpd version (build from `master`) with sftp module, with default Digests (MAC) algorithms
- /usr/bin/sftp client from openssh-client (`7.6p1-4ubuntu0.3`)

I'm connecting using umac-64 (which seems to be the default for both server & client with both no explicit configuration) or umac128 MAC algorithm:
```
sftp -v -F /dev/null -o MACs=umac-64@openssh.com -P 22 foobar@W.X.Y.Z
...[cut some irrelevant info]
debug1: Reading configuration data /dev/null
debug1: Connecting to W.X.Y.Z [W.X.Y.Z] port 22.
debug1: Connection established.
debug1: Authentications that can continue: password
debug1: Next authentication method: password
foobar@W.X.Y.Z's password:
debug1: Authentication succeeded (password).
Authenticated to W.X.Y.Z ([W.X.Y.Z]:22).
debug1: channel 0: new [client-session]
debug1: Entering interactive session.
Connected to W.X.Y.Z
sftp> ls data
data/anydir01  data/anydir02  data/anydir03  data/anydir04  data/anydir05  data/anydir06  data/anydir07
```
(So far, everything went fine).

Now, if I run `mkdir /home/foobar/data/anydir08 && chown foobar /home/foobar/data/anydir08` on the server, when listing `data` directory content I get:
```
Connected to W.X.Y.Z.
sftp> ls
data  
sftp> ls data
Corrupted MAC on input.
ssh_dispatch_run_fatal: Connection to W.X.Y.Z port 22: message authentication code incorrect
Connection closed
```

And on the server side's proftpd sftp.log:
```
2020-09-15 15:03:34,077 mod_sftp/1.0.1[19761]: sent server version 'SSH-2.0-mod_sftp'
2020-09-15 15:03:34,077 mod_sftp/1.0.1[19761]: received client version 'SSH-2.0-OpenSSH_7.6p1 Ubuntu-4ubuntu0.3'
2020-09-15 15:03:34,077 mod_sftp/1.0.1[19761]: handling connection from SSH2 client 'OpenSSH_7.6p1'
2020-09-15 15:03:34,116 mod_sftp/1.0.1[19761]:  + Session key exchange: curve25519-sha256
[...]

PR #1181, opened 2021-03-04
Issue #1111: Update our implementation of UMAC algorithm with upstream.

Castaglia:
At the same time, ensure the use of the `-fno-strict-aliasing` compiler option,
where supported.

We found that, with this combination of changes, even with the "buggy"
gcc-10 versions, the `umac-64@openssh.com` SFTP algorithm works once more
as expected.

### DIFF
--- a/configure
+++ b/configure
@@ -15032,6 +15032,29 @@ $as_echo "no" >&6; }
 fi
 rm -f core conftest.err conftest.$ac_objext conftest.$ac_ext
 
+    { $as_echo "$as_me:${as_lineno-$LINENO}: checking whether the C compiler accepts -fno-strict-aliasing" >&5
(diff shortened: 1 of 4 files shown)


severach:
Just `-fno-strict-aliasing` fixes the problem with 1.3.7a, no patch required. The patch also works fine.

Castaglia:
Excellent, thanks for the confirmation @severach .  I think I'll just backport the `-fno-strict-aliasing` change to the 1.3.7 branch; for the master branch, I'll use that and the updated `umac.c` implementation.

gjaekel:
> Just -fno-strict-aliasing fixes the problem with 1.3.7a, no patch required. The patch also works fine.

I confirm this using `gcc version 10.3.0 (Gentoo 10.3.0-r2 p3)`

Issue #1134, opened 2020-10-21
AuthUserFile-based logins, directory listings are very slow due to unbuffered reads


VeryGames:
### What I Did

Using mod_auth_file, with about 2500 lines in the AuthUserFile file.

### What I Expected/Wanted

While running proftpd on a VPS, I saw high CPU usage and very slow commands (list, login..).
While running "strace" on proftpd process, I can see that:
```
31867 read(10, "2", 1)                  = 1
31867 read(10, "4", 1)                  = 1
31867 read(10, "6", 1)                  = 1
31867 read(10, "5", 1)                  = 1
31867 read(10, "-", 1)                  = 1
31867 read(10, "s", 1)                  = 1
31867 read(10, "a", 1)                  = 1
31867 read(10, "r", 1)                  = 1
31867 read(10, "a", 1)                  = 1
31867 read(10, "u", 1)                  = 1
31867 read(10, "d", 1)                  = 1
31867 read(10, ":", 1)                  = 1
31867 read(10, "$", 1)                  = 1
31867 read(10, "1", 1)                  = 1
```
ect...

This is the content of the AuthUserFile file.
It's like if proftpd is reading the AuthUserFile file, but only chars one by one.
Is that a normal behavior ?

### ProFTPD Version and Configuration

# proftpd -V
```
Compile-time Settings:
  Version: 1.3.6 (stable)
  Platform: LINUX [Linux 4.19.0-12-cloud-amd64 x86_64]
  Built: Tue Mar 10 2020 23:03:08 UTC
  Built With:
[...]

PR #1195, opened 2021-03-13
Issue #1134: Skip using `fgets(3)` entirely, and use our own FSIO API…

Castaglia:
… for

doing buffered reads of `AuthUserFile`, `AuthGroupFile`.

### DIFF
--- a/modules/mod_auth_file.c
+++ b/modules/mod_auth_file.c
@@ -40,10 +40,6 @@
 # error "ProFTPD 1.2.7rc2 or later required"
 #endif
 
-#ifndef BUFSIZ
-# define BUFSIZ          PR_TUNABLE_BUFFER_SIZE
-#endif /* !BUFSIZ */
-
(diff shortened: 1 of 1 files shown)


modules/mod_auth_file.c
```diff
@@ -632,15 +628,15 @@ static int af_setgrent(pool *p) {
       return -1;
     }
 
-    /* As the file may contain sensitive data, we do not want it lingering
-     * around in stdio buffers.
-     */
-    (void) setvbuf(af_group_file->af_file, NULL, _IONBF, 0);
```
Castaglia, line 638: Actually, I wonder if it is this `setvbuf(3)` call which is causing `fgets(3)` (and other library `FILE *` functions) to do the non-buffered reads.  Quite possibly.

Still, I think this is a good change, as it removes more "moving parts", and ensures that what we are doing is more within the ProFTPD source code.

PR #1203, opened 2021-03-21
build: Fix linking the static libsupp.a library.

orbea:
When building proftpd with slibtool (https://dev.midipix.org/cross/slibtool) it fails.
```
/usr/bin/ld: cannot find -lsupp
```
This is because only `lib/libsupp.a` is created without `$(LIBTOOL)` and there is no `.la` file to rely on which would tell where or what to link to. GNU libtool just consumes `-L./lib` and happens to work while slibtool changes it to `-L./lib/.libs`.

Given this I think the best choice is to link against `lib/libsupp.a` directly and get rid of all the `-lsupp` from `configure.in`.

Alternatives would be to use `-Wl,-L./libs` instead or to create libsupp with `$(LIBTOOL)`.

Also see this proftpd issue made for the slibtool tracker:  https://dev.midipix.org/cross/slibtool/issue/25

Full build log using the current git master: [proftpd.log](https://github.com/proftpd/proftpd/files/6176773/proftpd.log)

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -88,7 +88,7 @@ if test $ac_cv_prog_gcc = no -a "$OSTYPE" = "-DHPUX10"; then
   CFLAGS="$CFLAGS -Ae"
 fi
 
-LDFLAGS="-L\$(top_srcdir)/lib -L\$(top_builddir)/lib $LDFLAGS"
+LDFLAGS="-Wl,-L\$(top_srcdir)/lib,-L\$(top_builddir)/lib $LDFLAGS"
 
 # AIX has issues with the -rdynamic linker flag.  How many different AIX
 # versions should we support here?


orbea:
Also, I reproduced this in gentoo using both the ebuild and the git repo.
```
export MAKEFLAGS='LIBTOOL=rdlibtool' # For better debug output
autoconf
./configure
make
```

Castaglia:
I think the alternative of using `$(LIBTOOL)` to build libsupp would be more in line with the rest of ProFTPD's build system.

orbea:
I looked into using `$(LIBTOOL)` more, but in retrospect I think its a flawed idea. Since its a static library the libtool implementation will at best just be an abstraction layer around `ar` and `ranlib`. Additionally given the unique build system proftpd uses it was (At least for me) not trivial.

Probably the easiest and safest change is to use '-Wl`, I updated this PR with that.

Castaglia:
Another approach might be to handle those files as normal `.o` files, like the others.  The creation of this `libsupp` library is something that I inherited with the project; I surmise that the original author wanted to make those functions into a more generally-available library, used by other applications -- which never really materialized.  I may take a look into doing this as well.

PR #1211, opened 2021-04-02
Documentation - How to - Fix invalid use of a directive in mod_sql how to

ubawurinna:
Replace SQLBackend directive with SQLAuthTypes directive (copy/paste error?)


Castaglia:
Thanks!

PR #1222, opened 2021-04-14
Move fuzzer and oss-fuzz build script upstream

AdamKorcz:
With Proftpd just having been integrated to OSS-fuzz, I am moving the fuzzer and build script upstream to allow for an easier workflow when adding more fuzzers or modifying the existing.


tests/fuzzing/json_fuzzer.c
```diff
@@ -0,0 +1,49 @@
+/*
+ * ProFTPD - FTP server API testsuite
```
Castaglia, line 2: Nit: Maybe change this preamble to be:
```
ProFTPD - FTP server fuzzing testsuite
```
for clarity?

PR #1224, opened 2021-04-27
Add logging after dir_check() in TLS module

byshen:
Add logging for `<Limit>` configuration check in TLS module, besides the `tls_log()`.


Castaglia:
Thanks!

PR #1226, opened 2021-04-29
Add logging when removing directory failed

byshen:
there is no log message after removing directory failed in SITE command

### DIFF
--- a/contrib/mod_site_misc.c
+++ b/contrib/mod_site_misc.c
@@ -310,6 +310,9 @@ static int site_misc_delete_dir(pool *p, const char *dir) {
   xerrno = errno;
 
   if (res < 0) {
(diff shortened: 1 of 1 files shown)


contrib/mod_site_misc.c
```diff
@@ -310,6 +310,9 @@ static int site_misc_delete_dir(pool *p, const char *dir) {
   xerrno = errno;
 
   if (res < 0) {
+    pr_log_debug(DEBUG3, MOD_SITE_MISC_VERSION
+      ": error removing directory '%s', %s", dir,
```
Castaglia, line 314: I tend to use `:` as the separator between a log message and the error string, _e.g._:
```
error removing directory '%s': %s
```
Mind updating this to use that separator, rather than `,`?

byshen: I updated it. Thanks for the review :D

PR #1255, opened 2021-06-14
Attempting to address LGTM warnings about pointer arithmetic in `mod_…

Castaglia:
…tls_shmcache`.

### DIFF
--- a/contrib/mod_tls_shmcache.c
+++ b/contrib/mod_tls_shmcache.c
@@ -2,7 +2,7 @@
  * ProFTPD: mod_tls_shmcache -- a module which provides shared SSL session
  *                              and OCSP response caches using SysV shared
  *                              memory segments
- * Copyright (c) 2009-2020 TJ Saunders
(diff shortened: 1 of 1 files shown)


contrib/mod_tls_shmcache.c
```diff
@@ -564,7 +564,7 @@ static struct sesscache_data *sess_cache_get_shm(pr_fh_t *fh,
     "using shm ID %d for sesscache path '%s' (%u sessions)", sesscache_shmid,
     fh->fh_path, shm_sess_max);
 
-  data->sd_entries = (struct sesscache_entry *) (data + sizeof(struct sesscache_data));
+  data->sd_entries = (struct sesscache_entry *) (data + 1);
```
Castaglia, line 567: Since `data` is defined to be a `struct sesscache_data *`, we don't need to use `sizeof`; we only need to increment by one, and let C naturally Do The Right Thing(tm).

PR #1259, opened 2021-06-21
Specify file mode for open with O_CREAT or O_TMPFILE

pghmcfc:
Use of `open()` with `O_CREAT` or `O_TMPFILE` in second argument requires a third argument to specify the required file mode.  Without this value, a random value from the stack is used instead.

Furthermore, with `glibc` and compiling with `-D_FORTIFY_SOURCE=1` or higher, the absence of the third argument causes errors like this:
```
In file included from /usr/include/fcntl.h:329,
                 from /usr/include/sys/file.h:24,
                 from ../include/os.h:90,
                 from ../include/conf.h:32,
                 from api/tests.h:30,
                 from api/scoreboard.c:27:
In function 'open',
    inlined from 'scoreboard_set_test_fn' at api/scoreboard.c:146:8:
/usr/include/bits/fcntl2.h:50:11: error: call to '__open_missing_mode' declared with attribute error: open with O_CREAT or O_TMPFILE in second argument needs 3 arguments
   50 |           __open_missing_mode ();
      |           ^~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/tests/api/ctrls.c
+++ b/tests/api/ctrls.c
@@ -74,7 +74,7 @@ static int devnull_fd(void) {
 static int tmpfile_fd(void) {
   int fd;
 
-  fd = open(tmpfile_path, O_CREAT|O_RDWR);
+  fd = open(tmpfile_path, O_CREAT|O_RDWR, 0600);
   if (fd < 0) {
     fprintf(stderr, "Error opening %s: %s\n", tmpfile_path, strerror(errno));
     return -1;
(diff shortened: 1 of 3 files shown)


pghmcfc:
Whilst I'm here, would you be open to a PR that replaced the deprecated (since check 0.9.9 in 2012) `fail_if` and `fail_unless` macros used in the API tests with the preferred `ck_assert_msg` style? That would get rid of a large number of compiler warnings when compiling against check 0.15 or later.

Castaglia:
@pghmcfc Sure, a separate PR for the `fail_if`/`fail_unless` macro changes would be welcome.  (Sadly, I _like_ the `fail_if` macros, as I think they read better, more human-like.  Ah well.)

PR #1262, opened 2021-06-24
check API: update from deprecated fail*() macros to ck_*() macros

pghmcfc:
`fail()` replaced by `ck_abort_msg()`
`fail_unless()` replaced by `ck_assert_msg()`
`fail_if()` replaced by `ck_assert_msg()` and inverting the sense of the assertion

Having made those changes, I then cleared up some mismatches between the format strings used in the assertions and the supplied parameters. Mostly this was to handle objects of type `size_t`, by using the format `%lu` and casting the parameter to an `unsigned long`. There were a few other things though, such as missing parameters, `errno` and `strerror(errno)` having their parameters the wrong way around, etc.

### DIFF
--- a/contrib/mod_auth_otp/t/api/base32.c
+++ b/contrib/mod_auth_otp/t/api/base32.c
@@ -74,8 +74,8 @@ START_TEST (base32_encode_test) {
   int res;
 
   res = auth_otp_base32_encode(p, NULL, 0, NULL, NULL);
-  fail_unless(res < 0, "Failed to handle null arguments");
-  fail_unless(errno == EINVAL, "Expected errno %s (%d), got %s (%d)",
+  ck_assert_msg(res < 0, "Failed to handle null arguments");
+  ck_assert_msg(errno == EINVAL, "Expected errno %s (%d), got %s (%d)",
     strerror(EINVAL), EINVAL, strerror(errno), errno);
 
   for (i = 0; i < expected_code_count; i++) {
@@ -86,9 +86,9 @@ START_TEST (base32_encode_test) {
     raw_len = strlen((char *) raw);
 
     res = auth_otp_base32_encode(p, raw, raw_len, &encoded, &encoded_len);
-    fail_unless(res == 0, "Failed to base32 encode '%s': %s",
+    ck_assert_msg(res == 0, "Failed to base32 encode '%s': %s",
       expected_codes[i].raw, strerror(errno));
-    fail_unless(strcmp((char *) encoded, expected_codes[i].encoded) == 0,
+    ck_assert_msg(strcmp((char *) encoded, expected_codes[i].encoded) == 0,
       "Expected '%s' for value '%s', got '%s'", expected_codes[i].encoded,
       expected_codes[i].raw, encoded);
   }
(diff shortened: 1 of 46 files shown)


pghmcfc:
Any chance this might get merged before 1.3.8 final release? Are there any issues you'd like me to address?

Castaglia:
Do you have any instructions on how I might reproduce the deprecation warnings locally (ideally using a Docker container)?  I'd like to be able to do that -- and then to see how I can provide my own macros.  I honestly prefer the `fail_if/fail_unless` assertion language better, as it reads more like an assertion to my eyes -- so I'd be looking to address these deprecation warnings while also preserving that assertion terminology.

pghmcfc:
There are no deprecation warnings so it's not surprising that you don't see any.

The `fail*` APIs have been deprecated since `check` version 0.9.9 a decade ago. They no longer appear in the `check` API documentation: https://libcheck.github.io/check/doc/doxygen/html/check_8h.html
They're mentioned at https://libcheck.github.io/check/doc/check_html/check_4.html#Convenience-Test-Functions (scroll down to the bottom of the list) where they are explicitly described as deprecated.

Now, whilst there may not be any deprecation warnings being issued, there are a bazillion warnings about having too many arguments for the supplied format string, like this:
```
api/error.c:5195:24: warning: too many arguments for format [-Wformat-extra-args]
 5195 |   fail_unless(res < 0, "Unexpectedly explained error");
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I see that when building on RHEL9 or any recent Fedora. You might try a Rocky 9 docker image (https://hub.docker.com/_/rockylinux) in the same way that you might have used a CentOS image in the  past. I do my RHEL9 builds using a different (not docker) environment but based on Rocky 9 so it should be OK for you.

Now, why are these warnings showing up? The upstream discussion is at https://github.com/libcheck/check/issues/293
Basically, the `fail*` macros just call the underlying `ck_assert*` APIs and in `check` version 0.15.0 the `ck_assert*` APIs were enhanced to support `gcc`'s ability to check format string arguments. This resulted in some failures to compile code like `fail_if(some_boolean_value)` with no message, which upstream never intended to work but some people used anyway. To avoid having to remove the argument checks but still keep users' code working, an extra NULL argument was added to the end of the argument list in the macros. This is why any `fail*` macro used with an error message (as it is supposed to be of course) now generates "too many arguments for format" warnings and that's my primary reason for generating this PR. Whilst doing it, I actually fixed some real errors (see the "Match assert format strings and their parameters" commit) that were exposed by the format string checking, so if you don't want to take the whole PR, I could create another one that only includes those issues.

Incidentally, this PR came about after I asked last year if you'd be open to such a PR: https://github.com/proftpd/proftpd/issues/1259#issuecomment-865035885

Castaglia:
I've played with this a bit, trying to implement some wrappers -- and they get messy quickly.  Would be nicer if the `check` library provided a `ck_assert_msg` variant which accepted a `va_list` argument, but alas, the library does not.

So, for now, rather than try to fight that upstream currently, I'll merge this as-is.  And maybe sometime later, if it bothers me, I'll tackle the esthetics of assertion language separately then.

Thanks for doing this work, @pghmcfc!

Castaglia:
Hmm.  As a follow-up to this, I'll probably want to update the `check`-based unit tests for my other, non-standard ProFTPD modules' repos as well...

Issue #1260, opened 2021-06-22
Add use of Centos image to CI workflow


Castaglia:
The scope of this task is to add building/running of ProFTPD and its tests on a Centos image in the CI workflow, in addition to the existing Ubuntu image.

As @pghmcfc noted in [this PR](https://github.com/proftpd/proftpd/pull/1259), there are some compiler warnings/errors that might be picked up on that popular platform that may not necessarily be seen on Ubuntu.

Castaglia:
Note that this may be more involved than expected; Centos is _not_ one of the supported GitHub Action [runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).  Thus for this, we may need a separate (but similar) workflow that runs a Centos Docker image on an Ubuntu runner.

Or perhaps not; I think we might just need to specify the [container](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer) explicitly.  Which we should do anyway; I currently expect the build actions to be run in a Docker container, not on the runner host, anyway.

Castaglia:
Something like:
```
jobs:
  build:
    name: Build on ${{matrix.container }}
    runs-on: ubuntu-latest
    container: ${{ matrix.container }}
    strategy:
      matrix:
        container: ['ubuntu:18.04', 'centos:8']
```
And then the provisioning steps (installing necessary packages), compiler tests, _etc_ will need adjustment.

Castaglia:
Now, for extra credit, I'll see how to configure the `CFLAGS` environment variable -- for Centos -- to add/use _e.g._ `-D_FORTIFY_SOURCE=1`.  Or maybe it'll work for Ubuntu as well?  Let's experiment.

pghmcfc:
Install the `redhat-rpm-config` package and set CFLAGS to the output of this command:
```
rpm --eval '%optflags'
```
That should give you the default compiler flags used to build the rest of the distribution.

Even better, instead of doing that, try using the output of `rpm --eval %configure` instead of just `./configure`

Castaglia:
For my future reference, the above commands yield:
```
$ docker run --rm=true -it centos:8 bash
# yum install -y redhat-rpm-config

# rpm --eval '%optflags'
[...]

PR #1263, opened 2021-06-24
Issue #1260: Use the `-D_FORTIFY_SOURCE=1` CFLAGS value in our CI bui…

Castaglia:
…lds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,6 +183,7 @@ jobs:
       - name: Build with static modules
         env:
           CC: ${{ matrix.compiler }}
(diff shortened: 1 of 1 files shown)


Castaglia:
Looks like using `-D_FORTIFY_SOURCE=1` _also_ requires that optimization be enabled; our `--enable-devel` configure option, by default, _disables_ optimizations via `-O0`.  So now I'm thinking to support an _e.g._ `--enable-devel=fortify` configure option, such that the `fortify` category, if enabled, will ensure that optimization (at least `-O1`, of not `-O2`) is enabled.

Castaglia:
This has been superceded by the `--enable-devel=fortify` support now.

PR #1283, opened 2021-08-02
NLST command handler should treat paths beginning with './' as local (relative to cwd)

Sashan:
running ftp NLST command we obtain different results
for those commands `ls test/...` vs. `ls ./test/...`:

```
ftp> ls test/TEST????.dat
200 PORT command successful
test/TEST0001.dat
test/TEST0006.dat
test/TEST0008.dat
test/TEST0009.dat
test/TEST0007.dat
test/TEST0000.dat
test/TEST0003.dat
test/TEST0004.dat
test/TEST0005.dat
test/TEST0002.dat
226 Transfer complete
remote: test/TEST????.dat
190 bytes received in 0.0033 seconds (55.61 Kbytes/s)
ftp> ls ./test/TEST????.dat
200 PORT command successful
226 Transfer complete
```

output should be same in this case.

the proposed fix seems to solve the glitch (at least for me). I'm not familiar with proftpd code enough to tell the change is correct. there might be better way to fix it.

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -3211,7 +3211,7 @@ MODRET ls_nlst(cmd_rec *cmd) {
       p = *path;
       path++;
 
-      if (*p == '.' && (!opt_A || is_dotdir(p))) {
+      if (opt_A == 0 && *p == '.' && is_dotdir(p)) {
         continue;
       }
 


Sashan:
My colleague has pointed my earlier change is not quite right. His use case is as follows: `ls .test/test????.dat`. It looks like the glob results for his pattern are filtered here in master branch:
```
3204     /* Iterate through each matching entry */
3205     path = g.gl_pathv;
3206     while (path && *path && res >= 0) {
3207       struct stat st;
3208 
3209       pr_signals_handle();
3210 
3211       p = *path;
3212       path++;
3213 
3214       if (*p == '.' && (!opt_A || is_dotdir(p))) {
3215         continue;
3216       }
3217 
3218       pr_fs_clear_cache2(p);
```

I don't have a ftp client, which would allow me to pass options to `nlst` command, so I would be able to test `opt_A` case. For me `opt_A` is always zero, so `is_dotdir(p)` at line 3214 is never called. Therefore we always skip/remove result, which reads as `./test/test????.dat` is always removed.

I'm also not sure, what's the purpose of `is_dotdir()` function. Note there is also significant difference between 1.3.6 and master branch. 1.3.6 branch uses `strncmp()`,while master uses `strcmp()` to match patterns. This makes both version incompatible, see test case here:
```
#include <stdio.h>
#include <string.h>

#define TRUE    1
#define FALSE   0

/* Return true if dir is ".", "./", "../", or "..". */
int is_dotdir(const char *dir) {
  if (strcmp(dir, ".") == 0 ||
      strcmp(dir, "./") == 0 ||
      strcmp(dir, "..") == 0 ||
      strcmp(dir, "../") == 0) {
    return TRUE;
  }

  return FALSE;
}

/* Return true if dir is ".", "./", "../", or "..". */
int is_dotdir_n(const char *dir) {
  if (strncmp(dir, ".", 2) == 0 ||
      strncmp(dir, "./", 2) == 0 ||
      strncmp(dir, "..", 3) == 0 ||
      strncmp(dir, "../", 3) == 0) {
    return TRUE;
  }

  return FALSE;
}

int
main(int argc, const char *arrgv[])
{
        char *path = "./test/blame";
        char *dotpath = ".test/blame";

        printf("%s: %d\n", path, is_dotdir(path));
        printf("%s: %d\n", path, is_dotdir(dotpath));

        printf("%s: %d\n", path, is_dotdir_n(path));
        printf("%s: %d\n", path, is_dotdir_n(dotpath));

        return (0);
}
```
`is_dotdir_n()` comes from 1.3.6, the function `is_dotdir()` comes from master branch. To fix it for me in 1.3.6 I have to use change proposed in this pull request combined with `is_dotdir(), which comes from master branch.

Castaglia:
Can you provide the configuration you're using?  And the accompanying ProFTPD debug logging showing the above; there's a difference between `LIST` and `NLST`, and from the above, it's hard to see just which command your FTP client is using.  (FTP clients implement their client-side `ls` commands differently.)

Sashan:
This is the configuration:
```
ServerName                      "FTP server"
ServerIdent                     off
ServerType                      standalone
DefaultServer                   on
Port                            21

Umask                           022

# To prevent DoS attacks, set the maximum number of child processes
# to 30.  If you need to allow more than 30 concurrent connections
# at once, simply increase this value.  Note that this ONLY works
# in standalone mode, in inetd mode you should use an inetd server
# that allows you to limit maximum number of processes per service
# (such as xinetd).
MaxInstances                    30

# Set the user and group under which the server will run.
User                            ftp
Group                           ftp

# Normally, we want files to be overwritable.
AllowOverwrite          on

# Oracle Solaris ZFS/VFS provides enough caching already, therefore
# we disable proftpd file system cache.
FSCachePolicy           off

# Bar use of SITE CHMOD by default.
<Limit SITE_CHMOD>
  DenyAll
</Limit>

# Make PAM the final authority on what gets authenticated.
AuthOrder mod_auth_pam.c* mod_auth_unix.c

# Use system libraries for user lookups.
PersistentPasswd off

# Root login must be enabled explicitly. It is not permitted by default.
# Allow it only if you know what you are doing.
#RootLogin on

DisplayConnect          /etc/issue
```

I'm also attaching two packet traces (`no-fix.cap`, which shows packets from run _without_ patched version, `fix.cap` shows packet traces with fix in place). The sequence of commands one test_user logs in was as follows:
   - ls test/test000?.dat
   - ls ./test/test000?.dat
   - ls .test/test000?.dat
   - ls .test
   - ls ./.test/test000?.dat
 
[capturers.tar.gz](https://github.com/proftpd/proftpd/files/7075990/capturers.tar.gz)

Sashan:
the described issue is regression. proftpd version 1.3.3g works as expected. That means there is no difference in output between `ls test/TEST????.dat` and `ls ./test/TEST????.dat`. Both instances show the same files found in `test` subdirectory. proftpd version 1.3.4c contains the glitch described in the first comment. So it seems to me the issue got introduced with 1.3.4 version. I was not able to nail it down further as it looks like the oldest release tags found in repo fall into 1.3.5 release period. hope this helps.

Castaglia:
I opened #1325 to track this issue, and have put up a PR which addresses my local reproduction recipe, if you'd like to try it out: https://github.com/proftpd/proftpd/pull/1326

Castaglia:
With https://github.com/proftpd/proftpd/issues/1325 now fixed, I close this PR.  Thanks for reporting the issue!

PR #1304, opened 2021-08-15
Improve documentation regarding the usage of Device/Interface name in <VirtualHost>, MasqueradeAddress, and DefaultAddress.

mdell-seradex:
Added more information regarding the possibility of using the device/interface name for VirtualHost, MasqueradeAddress, and DefaultAddress.

Note that I was unable to determine how to adjust [http://www.proftpd.org/docs/directives/linked/config_ref_context_VirtualHost.html](http://www.proftpd.org/docs/directives/linked/config_ref_context_VirtualHost.html)
In addition, it appears as though there is no direct way to access `mod_core.html`, but I updated that as well.

I also found other issues which I was unable to determine how to address.  Specifically at [http://www.proftpd.org/docs/](http://www.proftpd.org/docs/)  the in `Directives list`, it offers access via a `single file` as a `ps` or a `pdf` file, but neither of those links work.

Let me know what you think of my suggested changes. 🙂

### DIFF
--- a/doc/Configuration.html
+++ b/doc/Configuration.html
@@ -5863,7 +5863,7 @@ CLASS="COMMAND"
 >DefaultAddress</B
 >  [  <CODE
 CLASS="OPTION"
->dns-names|ip-addresses seperated with spaces</CODE
+>ip-address|dns-name|device/interface-name [ip-address|dns-name|device/interface-name ...]</CODE
 >]</P
 ><P
 ></P
@@ -5908,7 +5908,7 @@ CLASS="SYNOPSIS"
 ></DT
 ><DD
 ><P
->1.2.7rc1 and later</P
+>1.2.7rc1 and later (Device/interface name only supported in 1.3.5rc1 and later)</P
 ></DD
 ></DL
 ></DIV
@@ -5926,7 +5926,8 @@ to, the default behaviour is to select whatever IP the system reports
 as being the primary IP.</P
 ><P
 >Starting with ProFTPD 1.3.0rc1 it's possible to use more than one FQDN or IP
-Address. With this change the old Bind directive has been deprecated.</P
+Address. With this change the old Bind directive has been deprecated. 
+ProFTPD 1.3.5rc1 added support for specifying teh device/interface name.</P
 ></DIV
 ><DIV
 CLASS="REFSECT1"
@@ -5957,7 +5958,10 @@ Port&nbsp;21<br>
 DefaultAddress&nbsp;192.168.10.30<br>
 <br>
 ##&nbsp;Since&nbsp;1.3.0rc1&nbsp;it's&nbsp;also&nbsp;possible&nbsp;to&nbsp;use&nbsp;the&nbsp;following:<br>
-#&nbsp;DefaultAddress&nbsp;192.168.10.30&nbsp;my.domain.tld</P
+#&nbsp;DefaultAddress&nbsp;192.168.10.30&nbsp;my.domain.tld<br>
+<br>
+##&nbsp;Since&nbsp;1.3.5rc1&nbsp;it's&nbsp;also&nbsp;possible&nbsp;to&nbsp;use&nbsp;the&nbsp;following:<br>
+#&nbsp;DefaultAddress&nbsp;eth0&nbsp;192.168.10.30&nbsp;my.domain.tld</P
 ></DIV
 ><H1
 ><A
@@ -15805,7 +15809,7 @@ CLASS="COMMAND"
 >MasqueradeAddress</B
 >  [  <CODE
 CLASS="OPTION"
->MasqueradeAddress ip-address|dns-hostname</CODE
+>MasqueradeAddress ip-address|dns-name|device/interface-name</CODE
 >]</P
 ><P
 ></P
@@ -15850,7 +15854,7 @@ CLASS="SYNOPSIS"
 ></DT
 ><DD
 ><P
->1.2.2 and later</P
+>1.2.2 and later (device/interface name only supported in 1.3.5rc1 and later)</P
 ></DD
 ></DL
 ></DIV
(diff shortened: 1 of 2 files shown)


mdell-seradex:
I forgot to mention that I was unable to determine how to adjust the `Directives list` `by context` and as such it is not updated.  Note that this is significantly lacking in information.
http://www.proftpd.org/docs/directives/linked/config_ref_context_VirtualHost.html

I also had a few thoughts for improving the documentation.
1. Change the `Compatibility` section to be something like this.
```
1.3.0     Introduced.
1.3.6rc2  Added support for X.
```
2. Change each directive to be stored in an include file (I think that works in html) instead of html files.  That way the information in each directive can be stored and maintained in one place and each `Directives list` html file(s) can refer to those include files, thereby ensuring that the information and format for each directive is consistent no matter how the directive is accessed.

mdell-seradex:
I just noticed that I only partly updated list in the `Directives list` `single file` `html` because it has two separate sections for several directives including `Virtualhost`.  The second section is lacking in a great deal of information.  [http://www.proftpd.org/docs/directives/configuration_full.html#CONTEXT-VIRTUALHOST](http://www.proftpd.org/docs/directives/configuration_full.html#CONTEXT-VIRTUALHOST)

doc/Configuration.html
```diff
@@ -5863,7 +5863,7 @@
 >DefaultAddress</B
 >  [  <CODE
 CLASS="OPTION"
->dns-names|ip-addresses seperated with spaces</CODE
```
Castaglia, line 5866: This single `Configuration.html` is a legacy artifact, from a previous doc maintainer who used doc generation tools to generate the HTML, PDF, etc files.  It is no longer really maintained, and has become stale.  I've been contemplating removing it altogether, and instead wanting to move most (if not all) of the docs to Markdown.

doc/modules/mod_core.html
```diff
@@ -614,11 +614,11 @@ <h3><a name="DebugLevel">DebugLevel</a></h3>
 <p>
 <hr>
 <h3><a name="DefaultAddress">DefaultAddress</a></h3>
-<strong>Syntax:</strong> DefaultAddress <em>ip-address|dns-name</em> <em>[ip-address|dns-name ...]</em><br>
+<strong>Syntax:</strong> DefaultAddress <em>ip-address|dns-name|device/interface-name [ip-address|dns-name|device/interface-name ...]</em><br>
 <strong>Default:</strong> None<br>
 <strong>Context:</strong> server config<br>
 <strong>Module:</strong> mod_core<br>
-<strong>Compatibility:</strong> 1.2.7rc1 and later
+<strong>Compatibility:</strong> 1.2.7rc1 and later (device/interface name only supported in 1.3.5rc1 and later)
```
Castaglia, line 621: I admit to a preference of having the description of the directive have a paragraph about which parameters are supported in which releases -- mostly because I feel that folks will actually _read_ those paragraphs; these directive headers are easy for the eye to just skip/skim over when sight-reading.  That's why the current docs use this approach.

doc/Configuration.html
mdell-seradex: That makes sense to me.  As you can tell, I didn't even realize that there was multiple sections for each directive at the beginning.
It is why I later suggested moving directives to include files to ensure consistent messaging.
I was thinking it was there as a single downloadable help file, but then using include files would probably require some tool to combine it as a single file if you wanted to keep that format.
I doubt most people would be upset if you deprecated some of those single file help files.

mdell-seradex (commented):
Moved comment.

mdell-seradex:
Feel free to make suggestions or changes.
We can use my fork as a test environment to review how it looks.
https://htmlpreview.github.io/?https://github.com/mdell-seradex/proftpd/blob/master/doc/Configuration.html
https://htmlpreview.github.io/?https://github.com/mdell-seradex/proftpd/blob/master/doc/modules/mod_core.html

mdell-seradex:
Draft pull requests... interesting & cool.
I converted it to that because this whole thing feels more like my suggested changes, although I suppose you could say that about any pull request.

doc/modules/mod_core.html
Castaglia: No worries.  Apologies for the delay in reviewing this further.

Right now, I'm wondering if we can choose either "device" or "interface" (rather than both), and align on that one word.  As a user, which term would be more natural/expected to you, for this documentation?

mdell-seradex: Being a Windows user, I am more familiar with the interface term, but that is only really from the use of the `netsh` console app.  Also I guess it really depends on what you mean.
Another term I am used to is `adapter`, as I usually see that in `Network & Internet Settings` where it has `Change my adapter options`, or in `Network and Sharing Centre` it has `Change adapter settings`.
Network interface, network device, and network adapter could be all interchangeable, but perhaps could also mean different things.  For example device could mean the actual network card, while interface could be the thing the networking messages are sent to, and adapter could effectively be the thing that connects those two.  I also could be completely wrong about this.  I just looked at wikipedia and it appears to say that an adapter is a N.I.C. [https://en.wikipedia.org/wiki/Network_adapter](https://en.wikipedia.org/wiki/Network_adapter)  It also calls it a "Physical network interface".

Perhaps the easiest way is if you had a glossary page, we could choose one of the words and have it be a link to the glossary, perhaps even have it be a pop-up like in wikipedia, although that may be asking a lot, and the glossary could explain this further and even state that often these words are used interchangeably.
Another option would be to instead link to an outside source, e.g. wikipedia, or dictionary.com, for the definition.
A third option would be to have a light glossary with links to an outside source.

If you are asking me what word I think would be better for the average user, I feel like it is adapter, however it does not really feel like it would lend itself to things like `eth0`.  I think interface would be better for that, but I am a more technical person, having been employed as a programmer analyst/developer for over 20 years.  I fear that some people may not know what interface is, which is why I suggest the glossary idea.
I also think that since I have spent limited time in the Linux world, perhaps I am not the best person to ask as I feel that it would be better to use terminology that Linux users are familiar with.

PR #1329, opened 2021-09-13
Mysql legacy auth

richard-whitty:
Adds support for the old mysql `PASSWORD` function output in mod_sql_mysql `Backend` auth.
Ignoring the quality of the code, is this something the maintainers would be opposed to?

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -11,7 +11,7 @@ ChangeLog files.
 
   + Fix memory disclosure to RADIUS servers by mod_radius (Issue #1284).
 
-  + PCRE expressions with capture groups are not being handled properly
+  + PCRE expressions with capture groups were not being handled properly
     (Issue #1300).
 
   + Add support for the libidn2 library, over libidn, for e.g. mod_rewrite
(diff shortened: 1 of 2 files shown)


Castaglia:
I understand the motivation behind this patch, but given that MySQL deprecated this functionality, I'm not sure of the long-term value of recreating in _this_ separate ProFTPD project.

Perhaps an alternative approach would to be work on a way to transparently upgrade/change these legacy MySQL `PASSWORD` entries into a newer MySQL algorithm that is supported...

richard-whitty:
Thanks for the response.

The situation I'm in is we migrated a client to a new linux install and they already had the database with hundreds of users on the old environment using the `PASSWORD` auth method - in this case the 41 byte. Unfortunately there were also php and django sites using the same database.

If it was a separate module that implemented all of the 'old' backend auth types would that be more palatable? Alternatively, we could remove the attempts to call unexported symbols and do it all in here. I did look at doing this in mod_sql_passwd.c but it didn't look like as good a fit.

Castaglia:
A separate `mod_sql_auth_legacy_mysql` module would be acceptable, yes -- that way, it need not be enabled by default, but sites that need it can enable/use it, especially if it can be a DSO/shared module.  I'd be happy to help answer questions/help with such a module.

PR #1368, opened 2021-12-29
Avoid error on some client using digest

manusfreedom:
Clients like Beyond Compare do not accept unwanted data in response.

### DIFF
--- a/contrib/mod_digest.c
+++ b/contrib/mod_digest.c
@@ -1955,11 +1955,11 @@ static modret_t *digest_xcmd(cmd_rec *cmd, unsigned long algo) {
     if (get_algo_md(algo) != NULL) {
       char *hex_digest;
 
-      pr_response_add(R_250, _("Computing %s digest"), get_algo_name(algo, 0));
+      /*pr_response_add(R_250, _("Computing %s digest"), get_algo_name(algo, 0));*/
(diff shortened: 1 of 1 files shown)


Castaglia:
Hmm.  If FTP clients cannot handle multi-line responses, that is a client bug; I'm not sure that working around it on the server side is the best solution.

manusfreedom:
There is no final RFC on CRC/hash (it's a big lack), but why add a comment in a lightweight protocol like FTP?
And when the client asks for the hash, the server should give only this data and not force the client to search in lines or other data.

And last remark, normally the answer code is 213, why is it 250 ? (250 is more used?)

The draft RFC does not mention any other line in the response: https://tools.ietf.org/id/draft-bryan-ftp-hash-05.html#rfc.section.3

Castaglia:
Per [RFC 2389, Section 2.2](https://datatracker.ietf.org/doc/html/rfc2389#section-2.2), "Server Replies":
> Implementors should assume that, unless stated to the contrary, any reply to any FTP command (including QUIT) may be of the multiline format described in [1].

I suspect the use of 250 as the response code was from the original implementation; the `mod_digest` was a contribution.  Changing it to use 213, per the Draft, seems fine.

Castaglia:
This should now be fixed in master; see Issue #1661 .

PR #1374, opened 2022-01-10
Fix typo about RFC number.

char-khan:
[In this comment](https://github.com/proftpd/proftpd/blob/master/src/main.c#L847), RFC 2460 is typo for RFC 2640.
RFC 2460 is about IPv6 and is not relevant.


Castaglia:
Thanks!

Issue #376, opened 2016-12-30
Run integration/regression testsuite as part of Travis build


Castaglia:
As part of the Travis CI builds, we should be running the Perl integration/regression testsuite.  [Bug#4064](http://bugs.proftpd.org/show_bug.cgi?id=4064) demonstrates that we *can* use `Test-Unit-0.25` (available as the `libtest-unit-perl` Ubuntu package).

Completion of this task will be done via this [existing PR](https://github.com/proftpd/proftpd/pull/373).

Castaglia:
@hpreusse I'm hoping you can help out here.  I am using the following Dockerfile, which provides all of the Perl modules expected/used by the ProFTPD regression suite -- except for `Net::FTPSSL`.  Do you know which Debian/Ubuntu package might provide that Perl module?
```
FROM ubuntu:18.04

# For the ProFTPD build
RUN apt-get update && \
  apt-get install -y \
    ftp \
    gcc \
    gdb \
    git \
    libgeoip-dev \
    libhiredis-dev \
    libmemcached-dev \
    libmysqlclient-dev \
    libpam-dev \
    libpcre3-dev \
    libpq-dev \
    libsqlite3-dev sqlite3 \
    libssl-dev \
    libwrap0-dev \
    make \
    ncurses-dev \
    net-tools \
    netcat \
    openssl \
    vim \
    zlib1g-dev

# For the ProFTPD regression tests
RUN apt-get install -y \
  libauthen-oath-perl \
  libcache-memcached-perl \
  libdata-dumper-simple-perl \
  libdatetime-perl \
  libdigest-crc-perl \
  libdigest-md5-file-perl \
  libdigest-sha-perl \
  libfile-copy-recursive-perl \
  libfile-path-tiny-perl \
  libfile-spec-native-perl \
  libio-socket-ssl-perl \
  libmime-base32-perl \
  libnet-address-ip-local-perl \
  libnet-ldap-perl \
  libnet-snmp-perl \
  libnet-ssh2-perl \
  libnet-ssleay-perl \
  libnet-telnet-perl \
  libposix-2008-perl \
  libredis-perl \
  libsnmp-perl \
  libsys-hostname-long-perl \
  libtest-unit-perl \
  libtime-hr-perl \
  liburi-escape-xs-perl \
  libwww-perl

CMD ["bash"]
```

Castaglia:
Ah, I found a solution for `Net::FTPSSL` in my Dockerfile -- adding this instruction:
```
[...]

PR #1381, opened 2022-01-23
mod_auth_file: Calculate the current buffer offset before reallocation

manuelm:
Hi. We recently upgraded to 1.3.7c and quickly noticed some crashes on login. GDB showed a `SIGSEGV` in `malloc` in `af_getgrent`. We have a big "all our ftp users have this supplementary group"-group. Easily more than 1024 chars. Since a crash in `malloc` is usually a hint for memory corruption I pocked around a bit more and stumbled upon `af_getgrentline`.

In https://github.com/proftpd/proftpd/blob/e428a379fc97db8400a497045a57a69468666901/modules/mod_auth_file.c#L397 it updates the copy pointer to the current offset inside the buffer. This calculation is obviously only meaningful if both pointers point to the same structure. However a few lines above the buffer reallocation happens and the pointer to the buffer gets updated. But the copy pointer does not. So the copy pointer still points to the old (and freed) buffer leading to garbage offset results.

This PR simply calculates the offset beforehand.

### DIFF
--- a/modules/mod_auth_file.c
+++ b/modules/mod_auth_file.c
@@ -385,6 +385,7 @@ static char *af_getgrentline(char **buf, int *buflen, pr_fh_t *fh,
 
     {
       char *new_buf;
+      size_t old_off = cp - *buf;
 
       new_buf = realloc(*buf, *buflen);
       if (new_buf == NULL) {
@@ -392,9 +393,9 @@ static char *af_getgrentline(char **buf, int *buflen, pr_fh_t *fh,
       }
 
       *buf = new_buf;
+      cp = *buf + old_off;
     }
 
-    cp = *buf + (cp - *buf);
     cp = strchr(cp, '\0');
   }
 


Castaglia:
Heh...good timing.  I encountered this exact same issue today, as I'm working on Issue #376.  I have a slightly different patch for this issue, which I'll post here in a minute.

Castaglia:
Here's the PR I mentioned; it includes some additional trace logging, for future debugging: https://github.com/proftpd/proftpd/pull/1382  Could you see if this also addresses the issues you're seeing?

Castaglia:
Hmm.  Further testing (with varying line lengths, buffer sizes) shows that my PR isn't quite correct yet; I'll try your change as well.  Thanks!

manuelm:
Shall I still take a look at your version? Haven't tested mine with a lot of line variations myself. Ran out of time. Sent you the patch at 1:30am local time :-)

On another topic: I had to rollback to 1.3.7a again, as downloading via SFTP isn't working anymore. Always returns `No such file or directory` although the file exists. Downloading via FTPS works. Uploading with both protocols works. We are using mod_vroot. Any hints where I should look?

Castaglia:
Yes, please do take a look/test my `mod_auth_file` patch; it has some additional trace logging that I think will be useful in the future, so I'm like to get that PR into shape, such that it fixes your case, too.

As for the `mod_vroot`+`mod_sftp` issue, please file a ticket in the https://github.com/Castaglia/proftpd-mod_vroot/issues project.  (I'm not surprised it's that commit (and thanks for tracking that down!); the interactions/differences between FTP and SFTP, and dealing with them in a module like `mod_vroot`, are making things quite tricksy.)

manuelm:
> Hmm. Further testing (with varying line lengths, buffer sizes) shows that my PR isn't quite correct yet

Can you tell me which lengths didn't work? I have tested the `AuthGroupFile` starting from 20 chars to (right now) 6450 chars and checked the trace log for the full string. No problems so far.

manuelm:
> My PR previously had an issue with the setting of the offset; I corrected that this past weekend. So the PR, in its current state, is what I'd like to commit (and backport to the 1.3.7 branch).

Oh ok. Yeah your current PR looks good to me.

Castaglia:
I'm going to close this PR, now that the other one has been merged.  Thanks for your help!

PR #1385, opened 2022-01-26
mod_tls: Load certificate chain also from certificate file

manuelm:
Hello again,

I'm currently adding dual certificate support for all our webservers and noticed that ProFTPD doesn't support multiple chains (per certificate type). So I simply added chain loading from the individual certificate files. The (new) certificate PEM file format is `[key] cert [chain]`, where as `key` and `chain` are optional. If both `key` and `chain` are combined into a single PEM, no other directives are required.

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -17523,6 +17523,15 @@ static int tls_ctx_set_dsa_cert(SSL_CTX *ctx, X509 **dsa_cert) {
         tls_dsa_key_file, tls_get_errors());
       return -1;
     }
+
+    res = SSL_CTX_use_certificate_chain_file(ctx, tls_dsa_key_file);
+    if (res != 1) {
+      PRIVS_RELINQUISH
+
+      tls_log("error loading optional chain from TLSDSACertificateKeyFile '%s': %s",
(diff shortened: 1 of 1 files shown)


Castaglia:
Would you mind opening a ticket, that contains an explanation of what you mean by "dual certificate support" -- and associating this PR with that ticket? That would help me understand (and to work on implementing accompanying regression tests for this change) -- and provide a historical reference, for others who're watching the project.  Thanks!

manuelm:
Sure. Will do this. I can also try to add a test myself.

"Dual certificate" in my case is having both RSA + ECDSA certificates configured at the same time. OpenSSL will select the certificate that best matches the clients supported ciphers. So in my case most clients will connect via an ECDSA cipher while older clients can still use an RSA based.

Also see:
https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/#different-certificates-per-algorithm or http://www.postfix.org/postconf.5.html#smtpd_tls_chain_files

Although this PR makes this configuration possible, it's also useful on its own. All softwares I know off support loading the chain from the certificate file. E.g. apache, nginx, dovecot, postfix.

PR #1391, opened 2022-02-05
lib/Makefile.in: add -static to libtool call

ffontaine:
Add `-static` to libtool call to avoid the following build failure raised since version 1.3.7a and https://github.com/proftpd/proftpd/commit/3b815f3306be378892d3916765659ba88e80952a on shared library systems where libtool install object files in `.libs` (https://www.gnu.org/software/libtool/manual/html_node/Creating-object-files.html):

```
libtool: compile:  /home/fabrice/buildroot/output/host/bin/arm-none-linux-gnueabi-gcc -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -DHAVE_CONFIG_H -DLINUX -I.. -I../include -I../include -g2 -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -Os -g0 -D_FORTIFY_SOURCE=1 -Wall -fno-omit-frame-pointer -fno-strict-aliasing -c openbsd-bcrypt.c  -fPIC -DPIC -o .libs/openbsd-bcrypt.o
/home/fabrice/buildroot/output/host/bin/arm-none-linux-gnueabi-gcc-ar rc libsupp.a pr_fnmatch.o sstrncpy.o strsep.o vsnprintf.o glibc-glob.o glibc-hstrerror.o glibc-mkstemp.o pr-syslog.o pwgrent.o hanson-tpl.o ccan-json.o openbsd-blowfish.o openbsd-bcrypt.o
/home/fabrice/buildroot/output/host/opt/ext-toolchain/bin/../lib/gcc/arm-none-linux-gnueabi/4.8.3/../../../../arm-none-linux-gnueabi/bin/ar: pr_fnmatch.o: No such file or directory
```

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/lib/Makefile.in
+++ b/lib/Makefile.in
@@ -22,7 +22,7 @@ Makefile: Makefile.in ../config.status
 	cd ../ && ./config.status
 
 .c.o:
-	$(LIBTOOL) --mode=compile --tag=CC $(CC) $(CPPFLAGS) $(CFLAGS) -c $<
+	$(LIBTOOL) --mode=compile --tag=CC $(CC) $(CPPFLAGS) $(CFLAGS) -static -c $<
 
 libltdl: dummy
 	cd libltdl/ && $(MAKE)


Castaglia:
Thanks!  What platform/distribution are you using, that encounters the above issue?

ffontaine:
I encountered the above issue when trying to bump proftpd on [buildroot](https://buildroot.org)

Castaglia:
Hmm.  I'm a little wary to merge this PR as-is; I'll need to look into what effects this would have for builds (such as non-buildroot builds) that currently don't need this flag.  It may be that this behavior should be conditional, _e.g._ via a command-line option to the `configure` script, if needed.

ffontaine:
An other option would be to avoid calling `libtool` to build objects which are only used to make `prbase.a` static library.
I don't really understand the rationale behind this change in 3b815f3306be378892d3916765659ba88e80952a.

Castaglia:
My main thought for the `lib/Makefile.in` change in https://github.com/proftpd/proftpd/commit/3b815f3306be378892d3916765659ba88e80952a was to make the compiling of `.o` files consistent across the build system; inconsistencies are harder to maintain.

That said, I've been thinking more about this `-static` addition, particularly in light of the fact that -- as you correctly point out -- the `.o` files built by `lib/Makefile` all end up in the `prbase.a` static archive anyway.  Thus I don't think it should cause any issues; if it does, we'll tackle them as they come up.

I'll thus merge this PR, and we'll go from here.  Thanks!

PR #1424, opened 2022-04-22
Add detailed error message for pr_display_file()

byshen:
pr_display_file() could fail for multiple reasons; adding detailed log messages would help admins debugging the error.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -447,12 +447,16 @@ int pr_display_file(const char *path, const char *fs, const char *resp_code,
 
   fh = pr_fsio_open_canon(path, O_RDONLY);
   if (fh == NULL) {
+    pr_log_debug(DEBUG6, "unable to open file '%s': %s",
+      path, strerror(errno));
     return -1;
   }
 
   res = pr_fsio_fstat(fh, &st);
   if (res < 0) {
     xerrno = errno;
+    pr_log_debug(DEBUG6, "unable to stat file '%s': %s",
+      path, strerror(errno));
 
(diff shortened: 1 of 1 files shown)


byshen:
@Castaglia I am not sure why it fails on freebsd-12-2  during the `pkginstall` step. Please let me know how I can fix it. Thanks!

Castaglia:
> @Castaglia I am not sure why it fails on freebsd-12-2 during the `pkginstall` step. Please let me know how I can fix it. Thanks!

I wouldn't worry too much about that; sometimes the FreeBSD VM support provided by CirrusCI is flaky.  As long as most of the other Linux checks succeed (especially with the FreeBSD 11.x build passing), it's safe to assume that the code will work across most platforms.

Castaglia:
The main consumers/users of the `pr_display_file()` function already do log, at debug level 6, the errors.  As such, I think this additional logging via `pr_log_debug()` might be redundant.

That _said_, for the different error cases it, it might be useful to instead use `pr_trace_msg()`, and log these errors using Trace logging, to a "display" trace channel.  Thoughts?

byshen:
> That _said_, for the different error cases it, it might be useful to instead use `pr_trace_msg()`, and log these errors using Trace logging, to a "display" trace channel. Thoughts?

I agree. I will change the logging level to trace level to provide additional information. Thank you for the review!

byshen:
Hi TJ, I have changed the logging level based on your suggestion. Please let me know if any other change is needed. Thanks!

PR #1425, opened 2022-04-22
Add log messages for sreaddir()

byshen:
sreaddir() could fail for multiple reasons; add detailed log messages to help admins understand why list dir fails.

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -1234,16 +1234,21 @@ static char **sreaddir(const char *dirname, const int sort) {
 
   pr_fs_clear_cache2(dirname);
   if (pr_fsio_stat(dirname, &st) < 0) {
+    pr_log_debug(DEBUG8, "error checking '%s': %s",
+      dirname, strerror(errno));
     return NULL;
   }
 
   if (!S_ISDIR(st.st_mode)) {
     errno = ENOTDIR;
+    pr_log_debug(DEBUG8, "'%s' is not a direcory", dirname);
     return NULL;
   }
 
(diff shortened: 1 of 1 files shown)


modules/mod_ls.c
```diff
@@ -1234,16 +1234,21 @@ static char **sreaddir(const char *dirname, const int sort) {
 
   pr_fs_clear_cache2(dirname);
   if (pr_fsio_stat(dirname, &st) < 0) {
+    pr_log_debug(DEBUG8, "error checking '%s': %s",
+      dirname, strerror(errno));
```
Castaglia, line 1238: FYI, you may notice in some places in the code that I use a pattern like this:
```
if (func(...) < 0) {
  int xerrno = errno;

  pr_log_debug(..., "some error occurred: %s", strerror(xerrno));
  errno = xerrno;
  return -1;
}
```

I often use a temporary `xerrno` variable to "cache" the actual `errno` value returned by the function, so that I can properly set/return that `errno` value for _this_ function's caller.  In the past, functions like `pr_log_debug()` (and others) have changed the global `errno` values themselves -- and that means that without careful handling, the caller might see the `errno` value as set by `pr_log_debug()`, and not by the actual function.

byshen: Thanks for the info, good to know that. I will change the pattern for all places.

byshen:
@Castaglia Hi TJ, I have updated this patch with suggested pattern for `errno`, as well as places in #1424, #1426 and #1427. Please let me know if there is anything else I need to change. Thanks :D

Issue #1384, opened 2022-01-26
unkillable connections to ftp port 21


gojun077:
### What I Did

I have been using proftpd 1.2.x in a production environment for several years and recently upgraded to the latest stable 1.3.7c and run proftpd inside a Docker container. Since the upgrade, security scanners like [Qualys](https://www.qualys.com/docs/web-application-scanning-datasheet.pdf) and random (malicious?) users create hanging connections stuck on status "Authenticating" that are unkillable with `ftpdctl` and that ignore `/etc/proftpd.conf` settings "TimeoutNoTransfer", "TimeoutStalled", "TimeoutIdle". I am able to kill the connections by using the connection PID from `ftptop` and executing `kill -9 <PID>`. These hanging connections are causing CPU 100%.

Here's what I see in the docker logs for proftpd for one such malicious connection that was made on Jan 13 and remained in COMMAND (authenticating) for 270h!:

```
{"log":"2022-01-13 23:35:22,627 ce99838480c9 proftpd[17641] 172.31.0.2 (193.187.91.226[193.187.91.226]): mod_delay/0.7: no DelayOnEvent rules configured with \"DelayTable none\" in effect, disabling module\n","stream":"stderr","time":"2022-01-13T23:35:22.627867719Z"}
{"log":"2022-01-13 23:35:22,628 ce99838480c9 proftpd[17641] 172.31.0.2 (193.187.91.226[193.187.91.226]): FTP session opened.\n","stream":"stderr","time":"2022-01-13T23:35:22.628442412Z"}
{"log":"2022-01-13 23:35:22,706 ce99838480c9 proftpd[17641] 172.31.0.2 (193.187.91.226[193.187.91.226]): USER admin: no such user found from 193.187.91.226 [193.187.91.226] to 172.31.0.2:21\n","stream":"stderr","time":"2022-01-13T23:35:22.707114467Z"}
{"log":"2022-01-13 23:35:22,786 ce99838480c9 proftpd[17641] 172.31.0.2 (193.187.91.226[193.187.91.226]): USER admin: no such user found from 193.187.91.226 [193.187.91.226] to 172.31.0.2:21\n","stream":"stderr","time":"2022-01-13T23:35:22.786513963Z"}
```

The attacker tries to login with an invalid user `admin` but this connection is never terminated until I discover it 1.5 weeks after Jan 13 and manually kill the PID.

### What I Expected/Wanted

[...]

PR #1428, opened 2022-04-24
Issue #1384: Refactor some common code in the Timer API for legibilit…

Castaglia:
…y/tracing, and add trace logging of alarm blocking.


src/timers.c
```diff
@@ -266,32 +275,39 @@ void handle_alarm(void) {
...
    */
-  while (nalarms) {
-    nalarms = 0;
 
-    if (!alarms_blocked) {
-      int alarm_elapsed, new_timeout;
-      time_t now;
+  if (alarms_blocked == TRUE) {
+    alarms_pending++;
+    return;
```
Castaglia, line 283: If alarms are blocked, we shouldn't enter the `while` loop at all.  Previously, we'd enter the `while` loop, set `nalarms = 0`, and _then_ check whether alarms were blocked.  If they were, then perhaps the `SIGALRM` signal which triggered this function call might not be properly handled, and in fact may have been "forgotten" due to the `nalarms = 0` setting.  Perhaps that might be the cause of some of the symptoms/behaviors reported in this issue?

src/pool.c
```diff
@@ -152,7 +152,7 @@ static void chk_on_blk_list(union block_hdr *blok, union block_hdr *free_blk,
 #endif /* PR_USE_DEVEL */
 }
 
-/* Free a chain of blocks -- _must_ call with alarms blocked. */
+/* Free a chain of blocks -- _must_ call with signals blocked. */
```
Castaglia, line 155: Turns out that attempting to use `pr_trace_msg()` in the `pr_alarms_block()` function leads to a stack overflow, since the trace logging code calls into the Pools API, which blocks alarms, which calls the trace logging code, etc.

To avoid this -- and to honor what the Pools API really needs here -- we block signals, rather than alarms.  And this is more desirable, as `pr_alarms_block()` does not block `SIGARLM` signals, it only blocks the _processing_ of the timer callbacks.  In these Pool routines, we don't want interruptions at all via signals, thus `pr_signals_block()` is better here anyway.

Castaglia: The follow-on, then, is that we cannot call `pr_trace_msg()` within the `pr_signals_block()` functions, either, lest we trigger the same stack overflow!

PR #1430, opened 2022-04-27
Update PR_SCORE_XFER_DONE in pr_throttle_pause

cus:
This has numerous advantages:

- Update the scoreboard in a single call making xfer_done and elasped_time come
  from the same moment and reducing scoreboard update overhead
- Eliminates excessive XFER_DONE updates in scoreboard, old code updated it
  after every block
- During FTP uploads the number of received bytes also get updated in XFER_DONE
  making ftpwho -o json display "transfer_bytes" for uploads as well

Signed-off-by: Marton Balint <cus@passwd.hu>

### DIFF
--- a/modules/mod_xfer.c
+++ b/modules/mod_xfer.c
@@ -2186,7 +2186,7 @@ MODRET xfer_stor(cmd_rec *cmd) {
     }
 
     /* If no throttling is configured, this does nothing. */
-    pr_throttle_pause(nbytes_stored, FALSE);
+    pr_throttle_pause(nbytes_stored, FALSE, nbytes_stored);
 
     if (session.range_len > 0) {
       if (nbytes_stored == upload_len) {
@@ -2256,7 +2256,7 @@ MODRET xfer_stor(cmd_rec *cmd) {
   }
 
   /* If no throttling is configured, this does nothing. */
-  pr_throttle_pause(nbytes_stored, TRUE);
+  pr_throttle_pause(nbytes_stored, TRUE, nbytes_stored);
 
   if (stor_complete(cmd->pool) < 0) {
     xerrno = errno;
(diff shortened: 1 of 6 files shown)


Castaglia:
Could you file a ticket explaining the background/use case for this PR, for historical reference?  I find it easier to track that kind of context in tickets, rather than in PR comments.  Thanks!

modules/mod_xfer.c
```diff
@@ -2853,11 +2853,6 @@ MODRET xfer_retr(cmd_rec *cmd) {
 
   pr_alarms_unblock();
 
-  cnt_steps = session.xfer.file_size / 100;
-  if (cnt_steps == 0) {
-    cnt_steps = 1;
-  }
```
Castaglia, line 2859: Why remove this code?  It is there to deliberately prevent `mod_xfer` from performing high amounts of Scoreboard IO.  Is it to improve the accuracy of the Scoreboard-reported progress?

cus: PR_SCORE_XFER_DONE updates in xfer_stor were moved to pr_throttle_pause and got unified with PR_SCORE_XFER_LEN and PR_SCORE_XFER_ELAPSED updates, so any code responsible for updates in xfer_stor, including update throttling is no longer needed.

In general, I believe a time based approach for throttling scoreboard updates (e.g. update in every 5 seconds) is better than a percent based approach. So IMHO that should be implemented in the future for the updates in pr_throttle_pause instead of the current PR_TUNABLE_XFER_SCOREBOARD_UPDATES based approach.

I was wrong however in the 2nd point of my commit message (I misread the code), because of the throttling you mentioned XFER_DONE updates are not excessive indeed.

cus:
The updated pull request contains a modified description and also keeps the percent based update logic, so scoreboard update rate should not change.

Castaglia:
I've committed this change (with a minor tweak) to master.  Thank you!

PR #1446, opened 2022-05-06
Issue #1439: Attempt to address regression in computing the hash requ…

Castaglia:
…ested via the `check-file` SFTP extension.

I am not entirely sure where the issue is; a small program that does (mostly)
what this code does, locally, computes the expected hash.  So I'm adding
some additional trace logging, and removing some unnecessary (due to file BIO
semantics) retry checking.  Let's see what happens.

### DIFF
--- a/contrib/mod_sftp/fxp.c
+++ b/contrib/mod_sftp/fxp.c
@@ -4007,21 +4007,62 @@ static void fxp_version_add_supported2_ext(pool *p, unsigned char **buf,
 
 /* SFTP Extension handlers */
 
+/* Returns -1 on error, 0 on EOF, and 1 on successful block read. */
+static int read_file_block(pr_fh_t *fh, EVP_MD_CTX *pctx, size_t blocksz,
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/fxp.c
```diff
@@ -4384,14 +4384,13 @@ static int fxp_handle_ext_check_file(struct fxp_packet *fxp, char *digest_list,
...
-        continue;
-      }
+    xerrno = errno;
 
-      /* error */
-      xerrno = errno;
+    pr_trace_msg(trace_channel, 19,
+      "reading %d bytes from '%s' via BIO yielded %d", data_len, path, res);
 
+    if (res <= 0) {
```
Castaglia, line 4392: Hmm.  What if the file size is legitimately zero bytes here?  I'll update this to add a check along these lines:
```
if (res < 0 ||
    (res == 0 && st.st_size != 0)) {
  /* error */
  ...
}
```

Issue #1445, opened 2022-05-05
BanOnEvent BadProtocol triggers segfault


ElDavoo:
### What I Did

I configured mod_ban in order to ban users and I tried banning myself.

### What I Expected/Wanted

I expected proftpd to ban me.

Note that I can get myself banned using ftpdctl .

### ProFTPD Version and Configuration

```
Compile-time Settings:
  Version: 1.3.6 (stable)
  Platform: LINUX [Linux 5.10.103-v7+ armv7l]
  Built: Sat Sep 18 2021 21:05:52 UTC
  Built With:
[...]

PR #1453, opened 2022-05-12
Issue #1445: The handling of the "bad protocol" event by `mod_ban` wa…

Castaglia:
…s incorrect, and leads to a segfault before the ban for the offending client is properly added.

### DIFF
--- a/contrib/mod_ban.c
+++ b/contrib/mod_ban.c
@@ -3398,8 +3398,8 @@ static void ban_anonrejectpasswords_ev(const void *event_data,
 
 static void ban_badprotocol_ev(const void *event_data, void *user_data) {
 
-  /* For this event, event_data is the client. */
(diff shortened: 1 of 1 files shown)


contrib/mod_ban.c
```diff
@@ -3398,8 +3398,8 @@ static void ban_anonrejectpasswords_ev(const void *event_data,
 
 static void ban_badprotocol_ev(const void *event_data, void *user_data) {
 
-  /* For this event, event_data is the client. */
-  conn_t *c = (conn_t *) event_data;
+  /* For this event, event_data is the bad command in question. */
+  conn_t *c = session.c;
```
Castaglia, line 3402: The `event_data` in question, for this event, is a `cmd_rec *`, not a `conn_t *`.  The typecast here leads to referencing illegal memory, hence the segfault.

PR #1461, opened 2022-05-16
Since we are now skipping setting a string for the MAC algorithm, for…

Castaglia:
… ciphers

where the MAC is implicit, we are uncovering null pointer dereferences causing
crashes, such as this one.


contrib/mod_sftp/mac.c
```diff
@@ -877,9 +877,9 @@ const char *sftp_mac_get_write_algo(void) {
 
   /* It is possible for there to be no MAC, as for some ciphers such as
    * AES-GCM.  Rather than returning NULL here, we indicate this by returning
-   * the empty string (see Issue #1411).
+   * a string (see Issue #1411).
    */
-  return "";
+  return "implicit";
```
Castaglia, line 882: These just make this `mac.c` implementation match what is logged during KEX, for consistency.

Issue #1463, opened 2022-05-17
mod_sftp client connection with hostkey ecdsa-sha2-nistp384 ecdsa-sha2-nistp256 hangs proftpd subprocess at 100% CPU


severach:
I'm using `SFTPClientMatch sftpHostKeys` to enable `ecdh` to clients otherwise limited to `ssh-rsa`. Sizes 256 and 384 hang on login. Size 521 works.

`ecdsa-sha2-nistp256` hangs
`ecdsa-sha2-nistp384` hangs
`ecdsa-sha2-nistp521` works

Clients tested JSCH-0.1.54, Bitvise sftp client 9.18, OpenSSH.

The standard size key is 256. Not sure how I can negotiate a 521 with only a 256 key so I replaced it with a 521 key. Switching back to a 256 key still hangs.

```
# ssh-keygen -l -f ssh_host_ecdsa256_key 
256 SHA256:THiDdF1gN1yw1soOjhaWxZI158ac/JIUCTYTcdX/sQA root@... (ECDSA)

# ssh-keygen -l -f ssh_host_ecdsa521_key 
521 SHA256:SZbdDq1vEsCUu4xYKUw7tvpso8dlpx2aE/p+4d0Phss root@... (ECDSA)

$ sftp -P 2224 -oHostKeyAlgorithms=ecdsa-sha2-nistp256 0.0.0.6
(long hang, systemctl restart proftpd)
Connection reset by 0.0.0.6 port 2224
Connection closed.  
Connection closed
```

Takes some time for systemd to shut proftpd down.

```
# pacman -Q proftpd-git
proftpd-git 1.3.8rc3.r42.g4392eff0d-1

# proftpd -V
Compile-time Settings:
  Version: 1.3.8rc4 (git)
  Platform: LINUX [Linux 5.15.39-1-lts x86_64]
  OS/Release:
    NAME="Arch Linux"
    PRETTY_NAME="Arch Linux"
    ID=arch
    BUILD_ID=rolling
  Built: Tue May 17 2022 00:07:42 UTC
  Built With:
[...]

PR #1464, opened 2022-05-17
Issue #1463: Correct typo (wrong loop iterator variable) when checkin…

Castaglia:
…g ECDSA host key algorithms for `SFTPHostKeys`.

### DIFF
--- a/contrib/mod_sftp/kex.c
+++ b/contrib/mod_sftp/kex.c
@@ -1629,6 +1629,8 @@ static const char *get_kexinit_hostkey_algo_list(pool *p) {
     const char *algo;
     int have_key = FALSE, supported_algo = FALSE;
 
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/kex.c
```diff
@@ -1649,7 +1649,7 @@ static const char *get_kexinit_hostkey_algo_list(pool *p) {
       if (res > 0) {
         register int j;
 
-        for (j = 0; j < res; i++) {
+        for (j = 0; j < res; j++) {
```
Castaglia, line 1652: When writing this code, I copy/pasted some other `i`-based loop -- and forgot to change all of the occurrences of `i` to `j` (so as to not interfere with the `i` usage by the outermost loop).

Issue #1314, opened 2021-08-22
extremely slow FTPS/FTPES download speed, using mod_tls


mvegetto:
### What I Did

Please describe what you did, including the clients you used, what errors
or messages you saw, _etc_.
```
wget https://github.com/proftpd/proftpd/archive/master.zip
unzip master.zip
wget https://github.com/Castaglia/proftpd-mod_vroot/archive/master.zip
unzip master.zip.1
sudo mv proftpd-mod_vroot-master/ mod_vroot
sudo mv mod_vroot/ proftpd-master/contrib/
cd proftpd-master/
sudo aptitude install libsodium-dev
./configure --enable-nls --prefix=/usr/local/proftpd --sysconfdir=/etc/proftpd --with-modules=mod_sftp:mod_tls:mod_vroot
make
sudo make install
sudo /usr/local/proftpd/sbin/proftpd -t
sudo /usr/local/proftpd/bin/ftpasswd --passwd --file=/etc/proftpd/ftpd.passwd --name=arutha --uid=1000 --gid=1000 --home=/data --shell=/bin/false
```
Encrypted FTPS or FTPES download  capped to 1-3 MB/s
Encrypted upload is fine (100 MB/s)
Non TLS download speed (Just FTP) is fine (100 MB/s)

[tls.log](https://github.com/proftpd/proftpd/files/7027430/tls.log)
[trace.log](https://github.com/proftpd/proftpd/files/7027431/trace.log)

No issues when I am using pure-ftpd (FTPES).

### What I Expected/Wanted

Normal encrypted download speed

### ProFTPD Version and Configuration
```
Compile-time Settings:
  Version: 1.3.8rc2 (git)
  Platform: LINUX [Linux 5.10.0-8-amd64 x86_64]
  Built: Sun Aug 22 2021 13:24:00 CEST
  Built With:
    configure  '--enable-nls' '--prefix=/usr/local/proftpd' '--sysconfdir=/etc/proftpd' '--with-modules=mod_sftp:mod_tls:mod_vroot'

  CFLAGS: -g2 -O2 -Wall -fno-omit-frame-pointer -fno-strict-aliasing
  LDFLAGS: -Wl,-L$(top_srcdir)/lib,-L$(top_builddir)/lib  -rdynamic
  LIBS:  -lssl -lcrypto -lsodium -lssl  -lcrypto -lcrypt -ldl  -pthread

  Files:
    Configuration File:
      /etc/proftpd/proftpd.conf
[...]

PR #1471, opened 2022-06-12
Issue #1314: I suspect that one potential cause of this unexpectedly …

Castaglia:
…slow FTPS download behavior is the so-called "adaptive buffering" behavior.

Well-intentioned, but I think this adapative buffering is unnecessarily
complex, and leading to undesirably small buffers/speeds.


Castaglia:
This effectively reverts [this commit](https://github.com/proftpd/proftpd/commit/87f849668e70458802d552d87ca1a9f1288896f2), which sadly is not linked to a ticket that provides more context for why I felt such adaptive buffering would be a good idea.  Without that context, simpler is better.

Issue #1476, opened 2022-06-28
Reload after omitting explicit ModulePath value causes fatal module load failures


voetsjoeba:
### What I Did

When you remove an explicit ModulePath line in your config that already matches the compiled-in default (and which therefore shouldn't cause a change in value), and then reload the proftpd master daemon, the daemon terminates with a fatal error that modules could not be found, like this:
```
Jun 28 14:32:35 bullseye proftpd[18029]: bullseye - received SIGHUP -- master server reparsing configuration file
Jun 28 14:32:35 bullseye proftpd[18029]: mod_dso/0.5: unable to load 'mod_site_misc.c'; check to see if 'site_misc/mod_site_misc.la' exists
Jun 28 14:32:35 bullseye proftpd[18029]: fatal: LoadModule: error loading module 'mod_site_misc.c': No such file or directory on line 2 of '/usr/local/etc/proftpd.conf'
```
But after this happens, the daemon (now without the explicit ModulePath line) will happily start up fresh again, finding and loading all of its modules without issue as before. Reloading works fine again after the restart, too.

It appears that the problem happens only for this particular sequence of events:
 - start the daemon while there is an explicit `ModulePath <same-path-as-compiled-in-value>` in the config file
 - remove the explicit ModulePath value from the config
 - reload the daemon

Here's the simplest possible way I've found to reproduce the error, with a fresh out-of-the-box build from master to `/usr/local` on a vanilla Debian bullseye machine:
```
apt-get -y install git build-essential

git clone https://github.com/proftpd/proftpd.git
cd proftpd/

./configure --enable-dso --with-shared=mod_site_misc && make && make install
```
Here's the relevant compiled in paths from the build (full output included below):
```
# /usr/local/sbin/proftpd -V
...
  Files:
    Configuration File:
      /usr/local/etc/proftpd.conf
    Pid File:
      /usr/local/var/proftpd.pid
    Scoreboard File:
      /usr/local/var/proftpd.scoreboard
    Header Directory:
      /usr/local/include/proftpd
    Shared Module Directory:
      /usr/local/libexec
...
```
Now creating a systemd unit (matching the one that ships with Debian's package) and adding the explicit ModulePath line matching the built-in value:
```
cat >/etc/systemd/system/proftpd.service <<'EOF'
# proftpd.service from Debian bullseye's proftpd-core 1.3.7a+dfsg-12+deb11u2
[Unit]
Description=proftpd
Wants=network-online.target
[...]

PR #1479, opened 2022-07-02
Issue #1476: Make sure to reinitialize the default `ModulePath` value…

Castaglia:
… in the `mod_dso` module during a SIGHUP/restart.

### DIFF
--- a/modules/mod_dso.c
+++ b/modules/mod_dso.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD: mod_dso -- support for loading/unloading modules at run-time
- * Copyright (c) 2004-2020 TJ Saunders <tj@castaglia.org>
+ * Copyright (c) 2004-2022 TJ Saunders <tj@castaglia.org>
  *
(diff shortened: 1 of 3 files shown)


modules/mod_dso.c
```diff
@@ -657,17 +657,21 @@ MODRET set_modulepath(cmd_rec *cmd) {
...
+  if (dso_pool != NULL) {
     destroy_pool(dso_pool);
+  }
 
   dso_pool = make_sub_pool(permanent_pool);
   pr_pool_tag(dso_pool, MOD_DSO_VERSION);
 
-#ifdef PR_USE_CTRLS
+  /* Reset the ModulePath (see Issue #1476). */
+  dso_module_path = PR_LIBEXEC_DIR;
```
Castaglia, line 672: This is the key change/fix here.  Without this, the `dso_module_path` value will be stale, as it is duplicated out of the now-destroyed previous `dso_pool`.

PR #1481, opened 2022-07-10
contrib: Update tls_mod for LibreSSL 3.5

orbea:
Tested with libressl-3.5.3.

Instead of making this a bigger mess I decided to not care about older libressl versions, but I can make effort to change that if someone really cares. :)

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -100,8 +100,7 @@ static DH *get_dh(BIGNUM *p, BIGNUM *g) {
     return NULL;
   }
 
-#if OPENSSL_VERSION_NUMBER >= 0x10100000L && \
-    !defined(HAVE_LIBRESSL)
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
   if (DH_set0_pqg(dh, p, NULL, g) != 1) {
     pr_trace_msg(trace_channel, 3, "error setting DH p/q parameters: %s",
       ERR_error_string(ERR_get_error(), NULL));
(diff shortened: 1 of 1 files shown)


Castaglia:
On what platforms are you using LibreSSL 3.5?  I'd like to see if it'd be possible to add that combination to the GitHub Actions CI workflow.  Thanks!

orbea:
I'm using it on Gentoo using the libressl overlay.

https://github.com/gentoo/libressl

The current situation is that the primary libressl user is OpenBSD and most linux distros dropped support, but I am trying to keep it working on Gentoo.

ffontaine:
It should be noted that `mod_sftp` also fails to build with libressl 3.5.

The most important issue is that `EVP_CIPHER_CTX` is opaque since https://github.com/libressl/openbsd/commit/3f84934ae67f005ce128c593cc9ec7fd7d3c80c5 however libressl doesn't provide any `EVP_CIPHER_meth_set_{iv_length,init,cleanup,do_cipher}`.

So, the update to `cipher.c` is not trivial.
Given that this PR on `tls_mod` hasn't been merged, do you plan on supporting libressl >= 3.5 or should I disable mod_sftp with libressl on buildroot?

Castaglia:
I'm currently looking through these changes, and the LibreSSL ChangeLog, with an eye toward the cost of code complexity, caused by maintaining compatibility with _both_ OpenSSL and LibreSSL across their versions.  It's getting to be painful.  So the questions I have to ask, @orbea are these:
* Why continue to use to LibreSSL instead of OpenSSL?
* For your LibreSSL needs, how many old LibreSSL versions need to be supported, and why?

Castaglia:
Would you care to see if https://github.com/proftpd/proftpd/pull/1587 meets your needs?

Castaglia:
I'm going to close this PR, in favor of https://github.com/proftpd/proftpd/pull/1587.  Thanks for the feedback/prompt on updating ProFTPD for newer LibreSSL versions.

PR #1498, opened 2022-08-09
test for HELP command needs to be updated

Sashan:
I've found t/commands/help.t test fails for 1.3.8-rc4
I think it is because the testcase needs to be updated.
it seems to miss some changes which accumulated in recent years.


Castaglia:
There are many, *many* changes to made to the integration tests -- what exactly are you wanting to accomplish with these PRs?

Castaglia:
This should now be fixed by https://github.com/proftpd/proftpd/commit/618d41a87f434ea6fe0c85d1fc9873192b76e3bd -- I also opportunistically updated the test to use the `test_setup()`, `test_cleanup()` helpers, to reduce some of the boilerplate.

Thanks!

PR #1504, opened 2022-08-18
different EPERM error string on Solaris

Sashan:
s/Operation not permitted/Insufficient privileges

### DIFF
--- a/tests/t/lib/ProFTPD/Tests/Commands/NOOP.pm
+++ b/tests/t/lib/ProFTPD/Tests/Commands/NOOP.pm
@@ -161,7 +161,11 @@ sub noop_limit_deny_issue684 {
       $self->assert($expected == $resp_code,
         "Expected response code $expected, got $resp_code");
 
-      $expected = 'Operation not permitted';
+      if ($^O eq 'solaris') {
+        $expected = 'Insufficient privileges';
+      } else {
+        $expected = 'Operation not permitted';
+      }
       $self->assert($expected eq $resp_msg,
         "Expected response message '$expected', got '$resp_msg'");
     };
--- a/tests/t/lib/ProFTPD/Tests/Commands/RANG.pm
+++ b/tests/t/lib/ProFTPD/Tests/Commands/RANG.pm
@@ -2170,7 +2170,11 @@ EOC
       $self->assert($expected == $resp_code,
         test_msg("Expected response code $expected, got $resp_code"));
 
-      $expected = "RANG: Operation not permitted";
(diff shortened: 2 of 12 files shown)


Castaglia:
And does the textual mapping of the `EPERM` errno value vary across different versions of Solaris?

Sashan:
> And does the textual mapping of the `EPERM` errno value vary across different versions of Solaris?

very good question... So I did poke to history which is very interesting. Solaris uses `Insufficient privileges` since 2020. So this string mapping is relevant for 11.4 only. Before that the `strerror()` gave `Not owner`. People were unhappy about `Not owner` message because it was creating confusion. The change request was filed ~2008. Let me quote some interesting parts here:
Roger Faulkner:
```
This is something of a history and standards issue.
For a long time (since SVR4 came from New Jersey in 1989),
we have had EPERM mapping into "Not owner".
It's in the file:

usr/src/lib/libc/port/gen/errlist

It would be easy enough to change this from this:
1       Not owner
to this:
1       Operation not permitted.

to follow the SUSv3 standard's wording.  But would this be wise?
I'm assigning Lee Damico as the RE so she can investigate
and answer the question.
```
Peter Memishian follows with:
```
This goes back *a lot* further than SVR4 -- from the UNIX v7 source
tree (src/libc/gen/errlst.c), circa 1979:

  char    *sys_errlist[] = {
        "Error 0",
-->     "Not owner",
        "No such file or directory",
        "No such process",
        "Interrupted system call",

One would hope that no scripts out there are actually parsing this
output message (especially given internationalization), but after
29 years, we may discover otherwise.

As an aside, we've worked around this in various places in the past,
such as the Perror2() routine in ifconfig(1M):

        switch (error) {
        case ENXIO:
                (void) fprintf(stderr, "%s: no such interface\n", str);
                break;
-->     case EPERM:
-->             (void) fprintf(stderr, "%s: permission denied\n", str);
                break;
        case EEXIST:
                (void) fprintf(stderr, "%s: already exists\n", str);
                break;
        default:
                errno = error;
                perror(str);
        }
```
In 2020 Solaris has settled on using `Insufficient privileges`. I don't know details behind why `Insufficient privileges` was chosen over `Operation not permitted`. Before that change we could see at least two different error strings for `EPERM`: we could see `Not owner` for cases where `strerror()` was used or `permission denied` which came from `ifconfig(8)` There could be other components which were deliberately avoiding `strerror()`.

The only thing which matters to proftpd is `strerror(). So there is `Insufficient privileges` for Solaris 11.4. For older versions there is `Not owner`.

PR #1506, opened 2022-08-18
doing 'ls -l' on synlink will always show lrwxrwxrwx on Solaris

Sashan:


tests/t/lib/ProFTPD/Tests/Commands/LIST.pm
```diff
@@ -3527,7 +3527,12 @@ sub list_symlink_issue940 {
       }
 
       my $file = 'src.link -> dst.d';
-      my $expected = 'lrwxr-xr-x';
+      my $expected;
+      if ($^O eq 'solaris') {
+        $expected = 'lrwxrwxrwx';
```
Castaglia, line 3532: _Sigh_.  I hate inconsistencies like this.  Why is Solaris so different from other platforms in this manner?  It only adds to brittleness of maintenance of things like this.

Sashan: there are more those `platforms` bends for other solaris cases. In case of `lrwxrwxrwx` symlink match there is similar platform hack used for linux in `t/lib/ProFTPD/Tests/Config/ShowSymlinks.pm` here:
```
 463 
 464       # Different platforms' Perl create symlinks with different perms
 465       my $expected_perms = 'lrwxr-xr-x';
 466       if ($^O eq 'linux') {
 467         $expected_perms = 'lrwxrwxrwx';
 468       }
 469 
```
snippet comes from current proftpd source code.

PR #1507, opened 2022-08-18
we need to add a delay at least 5 second (we add 6 secs to

Sashan:
get margin). I believe it may take up to 5 second for server
to detect command connection got closed while data were
in fly. In this period number of running instances is not dropped,
therefore new client can not connect (when MaxInstances is set to 1).

### DIFF
--- a/tests/t/lib/ProFTPD/Tests/Commands/RETR.pm
+++ b/tests/t/lib/ProFTPD/Tests/Commands/RETR.pm
@@ -2575,6 +2575,7 @@ sub retr_bug3496 {
 
       # Close the _control_ connection immediately
       $client->{ftp}->close();
+      sleep(6);
 
       my $client2 = ProFTPD::TestSuite::FTP->new('127.0.0.1', $port, 1, 1);
       $client2->login($setup->{user}, $setup->{passwd});


tests/t/lib/ProFTPD/Tests/Commands/RETR.pm
```diff
@@ -2575,6 +2575,7 @@ sub retr_bug3496 {
 
       # Close the _control_ connection immediately
       $client->{ftp}->close();
+      sleep(6);
```
Castaglia, line 2578: These sorts of timing changes can be _very_ dependent on the infrastructure running these tests; why is this needed here, _outside_ of any test equipment-dependent timings?

Sashan: > These sorts of timing changes can be _very_ dependent on the infrastructure running these tests; why is this needed here, _outside_ of any test equipment-dependent timings?

the way I understand the issue does not depend on equipment(*) much. The current code in question reads as follows:
```
2527   my $config = {
2528     PidFile => $setup->{pid_file},
2529     ScoreboardFile => $setup->{scoreboard_file},
2530     SystemLog => $setup->{log_file},
2531     TraceLog => $setup->{log_file},
2532     Trace => 'DEFAULT:10',
2533 
2534     AuthUserFile => $setup->{auth_user_file},
2535     AuthGroupFile => $setup->{auth_group_file},
2536     MaxInstances => 1,
2537 
2538     IfModules => {
2539       'mod_delay.c' => {
2540         DelayEngine => 'off',
2541       },
2542     },
2543   };
.....
2563       # When run too quickly with the other tests, this test can fail.  So
2564       # pause a little here.
2565       sleep(1);
2566 
2567       my $client = ProFTPD::TestSuite::FTP->new('127.0.0.1', $port, 1, 1);
2568       $client->login($setup->{user}, $setup->{passwd});
2569 
2570       my $conn = $client->retr_raw($test_file);
2571       unless ($conn) {
2572         die("Failed to RETR: " . $client->response_code() . " " .
2573           $client->response_msg());
2574       }
2575 
2576       # Close the _control_ connection immediately
2577       $client->{ftp}->close();
2578 
2579       my $client2 = ProFTPD::TestSuite::FTP->new('127.0.0.1', $port, 1, 1);
2580       $client2->login($setup->{user}, $setup->{passwd});
```
the test case launches server with maximum child (session) instances set to 1. at line 2577 we close the first session. The way I understand it may take the first session up to 5 seconds to disappear so master process decrements number of existing sessions. If second client 2579 comes fast enough it will hit the `MaxInstances` limit.

---
(*) My test system is VirtualBox instance thus some timing might be involved I can try to tune the `sleep()` down to 1 sec and will let you know.

PR #1511, opened 2022-08-18
must fix regular expression to cover groupname, ownername

Sashan:
file attributes on Solaris come with `groupname=...;` and `ownername=...` this changeset updates regular expressions to take that into an account.

### DIFF
--- a/tests/t/lib/ProFTPD/Tests/Commands/MLST.pm
+++ b/tests/t/lib/ProFTPD/Tests/Commands/MLST.pm
@@ -157,7 +157,11 @@ sub mlst_file_ok {
       $self->assert($expected == $resp_code,
         test_msg("Expected response code $expected, got $resp_code"));
 
-      $expected = 'modify=\d+;perm=adfr(w)?;size=\d+;type=file;unique=\S+;UNIX.group=\d+;UNIX.mode=\d+;UNIX.owner=\d+; \/.*\/cmds\.conf$';
+      if ($^O eq 'solaris') {
+        $expected = 'modify=\d+;perm=adfr(w)?;size=\d+;type=file;unique=\S+;UNIX.group=\d+;UNIX.groupname=\S+;UNIX.mode=\d+;UNIX.owner=\d+;UNIX.ownername=\S+; \/.*\/cmds\.conf$';
+      } else {
+        $expected = 'modify=\d+;perm=adfr(w)?;size=\d+;type=file;unique=\S+;UNIX.group=\d+;UNIX.mode=\d+;UNIX.owner=\d+; \/.*\/cmds\.conf$';
+      }
       $self->assert(qr/$expected/, $resp_msg,
         test_msg("Expected response message '$expected', got '$resp_msg'"));
     };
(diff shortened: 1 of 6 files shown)


Castaglia:
Hrm.  `username`, `groupname` facts are non-standard for the `MSLD`, `MLST` responses.  Local facts must be prefaced with a `X.`, per [RFC 3659, SEction 7.6](https://www.rfc-editor.org/rfc/rfc3659.html#section-7.6), and I don't see the `username`, `groupname` facts in the relevant [IANA registry](https://www.iana.org/assignments/os-specific-parameters/os-specific-parameters.xhtml#fact-names).

Since these additional facts are non-standard, I'm not inclined to update the testsuite to support them.

Sashan:
> Hrm. `username`, `groupname` facts are non-standard for the `MSLD`, `MLST` responses. Local facts must be prefaced with a `X.`, per [RFC 3659, SEction 7.6](https://www.rfc-editor.org/rfc/rfc3659.html#section-7.6), and I don't see the `username`, `groupname` facts in the relevant [IANA registry](https://www.iana.org/assignments/os-specific-parameters/os-specific-parameters.xhtml#fact-names).
> 
> Since these additional facts are non-standard, I'm not inclined to update the testsuite to support them.

How about changing the regex to make those facts optional? I would like to find solution so the test will pass on Solaris (and perhaps other OSes too). It's not clear to me how does it come response sent by`mod_facts.c` comes with `UNIX.groupname` and `UNIX.username` on Solaris.

If I understand code in `mod_facts.c` right client must explicitly ask for those attributes:
```
1880 
1881 MODRET facts_opts_mlst(cmd_rec *cmd) {
1882   register unsigned int i;
1883   char *method, *facts, *ptr, *resp_str = "";
1884 
1885   method = pstrdup(cmd->tmp_pool, cmd->argv[0]);
...
1915 
1916   /* Do not show any facts by default at this point; the processing of the
1917    * facts requested by the client will enable just the ones the client
1918    * wishes to receive.
1919    */
1920   facts_opts = 0;
1921   facts_mlst_feat_remove();
....
1925 
1926   while (ptr != NULL) {
1927     pr_signals_handle();
1928 
1929     *ptr = '\0';
....
1954 
1955     } else if (strcasecmp(facts, "UNIX.groupname") == 0) {
1956       facts_opts |= facts_get_show_opt(facts);
1957       resp_str = pstrcat(cmd->tmp_pool, resp_str, "UNIX.groupname;", NULL);
1958 
....
1966 
1967     } else if (strcasecmp(facts, "UNIX.ownername") == 0) {
1968       facts_opts |= facts_get_show_opt(facts);
1969       resp_str = pstrcat(cmd->tmp_pool, resp_str, "UNIX.ownername;", NULL);
1970 
```
The client here is perl Net::FTP client module/library. I have no idea why it asks for those attributes.

Castaglia:
There's a list of facts that will be provided by `mod_facts` by default; the client can explicitly request other facts, but if the client does not, the module still needs to know which facts to provide.

Castaglia:
These particular tests should now all pass, as of https://github.com/proftpd/proftpd/commit/c6502647e442f8affcfed97d3d2522353b43dbfb.

Castaglia:
Thanks for bringing this to my attention; these tests should now be fixed.

Sashan:
> Thanks for bringing this to my attention; these tests should now be fixed.

thank you for a quick turn around on this. the fix works for Solaris as expected.

PR #1514, opened 2022-08-19
variable $target_dir is part of global scope

Sashan:
I think we want to refer to target_dir from global scope, because we use the target_dir variable later at line 245.


Castaglia:
The `$target_dir` variable is actually not in "global" scope; it's in the enclosing scope.  But good enough for me.

PR #1515, opened 2022-08-19
looks like something has changed in NET::FTP client library

Sashan:
I understand error code 599 used to be generated by FTP client library itself.
We are seeing 421 error code at places where 599 used to be seen.


Castaglia:
And now these tests should be working as expected, as of https://github.com/proftpd/proftpd/commit/9a21ba31b75f872955f5fb2f2710c270bc33a090.

Sashan:
> And now these tests should be working as expected, as of [9a21ba3](https://github.com/proftpd/proftpd/commit/9a21ba31b75f872955f5fb2f2710c270bc33a090).

yes, this works too.
thanks.

PR #1518, opened 2022-08-19
Fixes failure reported by t/config/socketoptions.t testsuite

Sashan:
there are both (`SO_KEEPALIVE` and `TCP_KEEPIDLE`) TCP socket options defined on Solaris. In order to make test suite happy I've decided to let `TCP_KEEPIDLE` precedence over `SO_KEEPALIVE` when building proftpd on OS which supports both those options.

this pull request just highlights the issue. perhaps there is better way to fix it.

### DIFF
--- a/src/inet.c
+++ b/src/inet.c
@@ -989,7 +989,7 @@ int pr_inet_set_proto_keepalive(pool *p, conn_t *c,
   if (val != -1) {
     int option_name;
 
-# if defined(TCP_KEEPALIVE)
+#ifndef TCP_KEEPIDLE
     option_name = TCP_KEEPALIVE;
 # else
     option_name = TCP_KEEPIDLE;


Castaglia:
Does Solaris have both `TCP_KEEPALIVE` and `SO_KEEPALIVE` defined?  Are they defined as the exact same thing for some reason?

Sashan:
> Does Solaris have both `TCP_KEEPALIVE` and `SO_KEEPALIVE` defined? Are they defined as the exact same thing for some reason?

Sorry there was typo in my earlier command. The typo really confuses things. The issue is not about `SO_KEEPALIVE` vs `TCP_KEEPALIVE` it is `TCP_KEEPIDLE` vs. `TCP_KEEPALIVE`. Both those `TCP_KEEP*` options are handled here in `src/inet.c`
```
 978   /* We only try to set the TCP keepalive specifics if SO_KEEPALIVE was
 979    * enabled successfully.
 980    */
 981   pr_trace_msg(trace_channel, 15, "enabled SO_KEEPALIVE on socket fd %d",
 982     c->listen_fd);
 983 
 984   /* On Mac OS, the socket option is TCP_KEEPALIVE rather than
 985    * TCP_KEEPIDLE.
 986    */
 987 #if defined(TCP_KEEPIDLE) || defined(TCP_KEEPALIVE)
 988   val = tcp_keepalive->keepalive_idle;
 989   if (val != -1) {
 990     int option_name;
 991 
 992 # if defined(TCP_KEEPALIVE)
 993     option_name = TCP_KEEPALIVE;
 994 # else
 995     option_name = TCP_KEEPIDLE;
 996 # endif /* TCP_KEEPALIVE or TCP_KEEPIDLE */
 997 
 998 # ifdef __DragonFly__
 999     /* DragonFly BSD uses millsecs as the KEEPIDLE unit. */
1000     val *= 1000;
1001 # endif /* DragonFly BSD */
1002     if (setsockopt(c->listen_fd, IPPROTO_TCP, option_name, (void *) &val,
1003         sizeof(int)) < 0) {
1004       pr_log_pri(PR_LOG_NOTICE,
1005         "error setting TCP_KEEPIDLE %d on fd %d: %s", val, c->listen_fd,
1006        strerror(errno));
1007 
1008     } else {
1009       pr_trace_msg(trace_channel, 15,
1010         "enabled TCP_KEEPIDLE %d on socket fd %d", val, c->listen_fd);
1011     }
1012   }
1013 #endif /* TCP_KEEPIDLE */

```
The thing is the Solaris uses both options `TCP_KEEPALIVE` and `TCP_KEEPIDLE` The TCP_KEEPALIVE is used on Solaris to change keep alive TCP timer. While TCP_KEEPIDLE option tunes keep-alive timer per-socket (https://docs.oracle.com/cd/E88353_01/html/E37851/tcp-4p.html)

PR #1532, opened 2022-10-01
Issue #1384: Handle potentially infinite loop due to traversing a lin…

Castaglia:
…ked list which references itself.


Castaglia:
This did not work as hoped, as the loop in the linked-list is more complex, involves many more links in the chain.  Closing.

PR #1542, opened 2022-10-22
We're seeing `mod_sql_mysql` build errors on the FreeBSD Cirrus CI, w…

Castaglia:
…hich look to be related to inability to find the `<mysql.h>` header file.

Try specifying the expected path to the directory where it might be.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -22,7 +22,7 @@ task:
     - pkg install -y libmemcached
     - pkg install -y mysql57-client
     - pkg install -y ncurses
-    - pkg install -y openldap24-client
+    - pkg install -y openldap26-client
     - pkg install -y openssl
     - pkg install -y pcre
     - pkg install -y postgresql10-client


.cirrus.yml
```diff
@@ -22,7 +22,7 @@ task:
     - pkg install -y libmemcached
     - pkg install -y mysql57-client
     - pkg install -y ncurses
-    - pkg install -y openldap24-client
+    - pkg install -y openldap26-client
```
Castaglia, line 25: Examining the Cirrus CI logs for the "install" step showed a conflict between the `mysql57-client` package (which has a transitive dependency on OpenLDAP for some reason, and installed `openldap26-client`), and the `openldap24-client` package -- the latter being "older".  Installing this older OpenLDAP package resulted in `mysql57-client` being automatically (and unexpectedly!) uninstalled, thus the `mod_sql_mysql` build error.

Issue #1539, opened 2022-10-17
Confusing asymmetrical "session closed" log message without preceding "session opened" log message


3dw1n:
### What I Did

I have a proftpd server setup with ftp, sftp, explicit ftps and implicit ftps running on AWS behind a Network Loadbalancer where I try to get rid of any logging which is related to Loadbalancer monitoring.
With the configs shown below I managed to get rid of all NLB monitoring traffic except the implicit ftps 'FTP session closed." messages in system.log.

With ftp/ftps/sftp enabled and all logging enabled, we see connect on port 21 and disconnect, we see connect on port 22 and SSH2 session closed.
We don't see port 990 connect but we do see an extra FTP session closed.:
```
2022-10-17 09:26:08,014 ftp-server.com proftpd[1373047] ftp-server.com: ProFTPD 1.3.8rc4 (devel) (built Mon Oct 17 2022 08:34:45 UTC) standalone mode STARTUP
2022-10-17 09:26:13,002 ftp-server.com proftpd[1373149] ftp-server.com (network.loadbalancer.aws): connected - local  : ftp-server.com:21
2022-10-17 09:26:13,002 ftp-server.com proftpd[1373149] ftp-server.com (network.loadbalancer.aws): connected - remote : network.loadbalancer.aws:15859
2022-10-17 09:26:13,002 ftp-server.com proftpd[1373149] ftp-server.com (network.loadbalancer.aws): FTP session opened.
2022-10-17 09:26:13,003 ftp-server.com proftpd[1373149] ftp-server.com (network.loadbalancer.aws): FTP session closed.
2022-10-17 09:26:13,786 ftp-server.com proftpd[1373172] ftp-server.com (network.loadbalancer.aws): FTP session closed.
2022-10-17 09:26:17,181 ftp-server.com proftpd[1373233] ftp-server.com (network.loadbalancer.aws): connected - local  : ftp-server.com:22
2022-10-17 09:26:17,181 ftp-server.com proftpd[1373233] ftp-server.com (network.loadbalancer.aws): connected - remote : network.loadbalancer.aws:41353
2022-10-17 09:26:17,181 ftp-server.com proftpd[1373233] ftp-server.com (network.loadbalancer.aws): SSH2 session opened.
2022-10-17 09:26:17,181 ftp-server.com proftpd[1373233] ftp-server.com (network.loadbalancer.aws): mod_sftp/1.1.1: scrubbing 1 passphrase from memory
[...]

PR #1545, opened 2022-10-29
Issue #1539: Avoid logging the "session closed" message, if we have n…

Castaglia:
…ot logged the "session opened" message, for a given TCP connection.

Such "session closed" without "session opened" messages can occur if a module closes/rejects the TCP connection during session initialization, before all modules have successfully initialized themselves for the connection.  This leads to user confusion about such asymmetrical logging.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1495,6 +1495,8 @@ static void fork_server(int fd, conn_t *l, unsigned char no_fork) {
     pr_session_disconnect(NULL, PR_SESS_DISCONNECT_SESSION_INIT_FAILED, NULL);
   }
 
+  pr_event_generate("core.connected", conn);
+
   pr_log_debug(DEBUG4, "connected - local  : %s:%d",
     pr_netaddr_get_ipstr(session.c->local_addr), session.c->local_port);
   pr_log_debug(DEBUG4, "connected - remote : %s:%d",
--- a/src/session.c
+++ b/src/session.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD - FTP server daemon
- * Copyright (c) 2009-2020 The ProFTPD Project team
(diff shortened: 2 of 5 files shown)


src/main.c
```diff
@@ -1495,6 +1495,8 @@ static void fork_server(int fd, conn_t *l, unsigned char no_fork) {
     pr_session_disconnect(NULL, PR_SESS_DISCONNECT_SESSION_INIT_FAILED, NULL);
   }
 
+  pr_event_generate("core.connected", conn);
```
Castaglia, line 1498: We generate a new `core.connected` event only if all modules have successfully initialized themselves for this connection.

In the future, the following debug logging can be moved into an event listener for this event.

src/session.c
```diff
@@ -68,29 +72,44 @@ static void sess_cleanup(int flags) {
...
+    log_sess_closed = TRUE;
+  }
+
+  if (ServerType == SERVER_INETD &&
+      !(flags & PR_SESS_END_FL_SYNTAX_CHECK)) {
+    log_sess_closed = TRUE;
+  }
+
+  if (sess_connected == FALSE) {
+    log_sess_closed = FALSE;
```
Castaglia, line 98: Here, we skip logging the "session closed" message, if the session is not considered successfully connected.

src/session.c
```diff
@@ -379,7 +398,7 @@ int pr_session_set_idle(void) {
   pr_scoreboard_entry_update(session.pid,
     PR_SCORE_CMD_ARG, "%s", "", NULL, NULL);
 
-  if (session.user) {
+  if (session.user != NULL) {
```
Castaglia, line 401: Stylistic preference nit; no functional change.

PR #1546, opened 2022-10-30
Issue #1384: One possible cause of the infinite loops reported in thi…

Castaglia:
…s issue, when freeing pools, is that of corrupted pointers/memory, as from use-after-free occurrences.

With this hypothesis, I compiled ProFTPD using the `-DPR_DEVEL_NO_POOL_FREELIST` flag, and ran it under Valgrind, to see if Valgrind detected any use-after-free occurrences.  And it did report some, for my initial simple use case.

Let's address those, and see if the reported behavior improves.

### DIFF
--- a/src/fsio.c
+++ b/src/fsio.c
@@ -2208,7 +2208,6 @@ pr_fs_t *pr_register_fs(pool *p, const char *name, const char *path) {
         name, path);
 
       destroy_pool(fs->fs_pool);
-      fs->fs_pool = NULL;
 
       errno = xerrno;
       return NULL;
@@ -2451,7 +2450,6 @@ int pr_unregister_fs(const char *path) {
   fs = pr_remove_fs(path);
   if (fs != NULL) {
     destroy_pool(fs->fs_pool);
-    fs->fs_pool = NULL;
     return 0;
   }
 
@@ -4989,7 +4987,6 @@ pr_fh_t *pr_fsio_open_canon(const char *name, int flags) {
     int xerrno = errno;
 
     destroy_pool(fh->fh_pool);
-    fh->fh_pool = NULL;
 
     errno = xerrno;
     return NULL;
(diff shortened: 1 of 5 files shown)


src/fsio.c
```diff
@@ -2208,7 +2208,6 @@ pr_fs_t *pr_register_fs(pool *p, const char *name, const char *path) {
         name, path);
 
       destroy_pool(fs->fs_pool);
-      fs->fs_pool = NULL;
```
Castaglia, line 2211: The `fs` object here is itself allocated out of the `fs_pool`, thus attempting to dereference the `fs` pointer here, after the call to `destroy_pool()`, is itself a use-after-free case.

src/fsio.c
```diff
@@ -5130,8 +5126,6 @@ int pr_fsio_close(pr_fh_t *fh) {
 
   if (fh->fh_pool != NULL) {
     destroy_pool(fh->fh_pool);
-    fh->fh_pool = NULL;
-    fh->fh_path = NULL;
```
Castaglia, line 5134: The `fh` object here is itself allocated out of the `fh_pool`, thus attempting to dereference the `fh` pointer here, after the call to `destroy_pool()`, is itself a use-after-free case.

src/inet.c
```diff
@@ -685,7 +685,6 @@ void pr_inet_close(pool *p, conn_t *c) {
 
   if (c->pool != NULL) {
     destroy_pool(c->pool);
-    c->pool = NULL;
```
Castaglia, line 688: The `c` object here is itself allocated out of the `c->pool`, thus attempting to dereference the `c` pointer here, after the call to `destroy_pool()`, is itself a use-after-free case.

src/inet.c
```diff
@@ -697,7 +696,7 @@ void pr_inet_lingering_close(pool *p, conn_t *c, long linger) {
 
   (void) pr_inet_set_block(p, c);
 
-  if (c->outstrm) {
+  if (c->outstrm != NULL) {
```
Castaglia, line 699: These are stylistic preference nits; no functional change.

PR #1548, opened 2022-11-05
Issue #1384: While examining the reported stack trace closely, I noti…

Castaglia:
…ced that the Auth API creates `cmd_recs` whose `pool` and `tmp_pool` pointers point to the same pool, which is somewhat surprising.

So let's use separate pools for those.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -68,16 +68,19 @@ static struct config_src *parser_sources = NULL;
  */
 
 static struct config_src *add_config_source(pr_fh_t *fh) {
-  pool *p = pr_pool_create_sz(parser_pool, PARSER_CONFIG_SRC_POOL_SZ);
-  struct config_src *cs = pcalloc(p, sizeof(struct config_src));
+  pool *p;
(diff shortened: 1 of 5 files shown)


src/parser.c
```diff
@@ -68,16 +68,19 @@ static struct config_src *parser_sources = NULL;
...
-  pool *p = pr_pool_create_sz(parser_pool, PARSER_CONFIG_SRC_POOL_SZ);
-  struct config_src *cs = pcalloc(p, sizeof(struct config_src));
+  pool *p;
+  struct config_src *cs;
 
+  p = pr_pool_create_sz(parser_pool, PARSER_CONFIG_SRC_POOL_SZ);
   pr_pool_tag(p, "configuration source pool");
-  cs->cs_next = NULL;
+
+  cs = pcalloc(p, sizeof(struct config_src));
```
Castaglia, line 77: This is a stylistic cleanup; I prefer separating variable declaration from variable assignments due to function calls.

PR #1554, opened 2022-11-18
configure.ac: fix bashisms in configure.ac

thesamesam:
configure scripts need to be runnable with a POSIX-compliant /bin/sh.

On many (but not all!) systems, /bin/sh is provided by Bash, so errors like this aren't spotted. Notably Debian defaults to /bin/sh provided by dash which doesn't tolerate such bashisms as '=='.

This retains compatibility with bash.

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -852,7 +852,7 @@ AC_ARG_ENABLE(pcre,
   )],
   [
     if test x"$enableval" = xyes ; then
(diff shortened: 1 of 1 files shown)


Castaglia:
Thanks!

thesamesam:
Thanks!

PR #1555, opened 2022-11-20
Bug #4491: The handling of RFC 8332 support for SHA2 signatures for R…

Castaglia:
…SA keys adds some interesting edge cases.

### DIFF
--- a/contrib/mod_sftp/keys.c
+++ b/contrib/mod_sftp/keys.c
@@ -5695,7 +5695,8 @@ int sftp_keys_verify_signed_data(pool *p, const char *pubkey_algo,
 
 #if defined(HAVE_SHA256_OPENSSL)
   } else if (strcmp(sig_type, "rsa-sha2-256") == 0) {
-    if (strcmp(pubkey_algo, sig_type) != 0) {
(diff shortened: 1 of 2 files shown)


contrib/mod_sftp/keys.c
```diff
@@ -5695,7 +5695,8 @@ int sftp_keys_verify_signed_data(pool *p, const char *pubkey_algo,
 
 #if defined(HAVE_SHA256_OPENSSL)
   } else if (strcmp(sig_type, "rsa-sha2-256") == 0) {
-    if (strcmp(pubkey_algo, sig_type) != 0) {
+    if (strcmp(pubkey_algo, sig_type) != 0 &&
+        strcmp(pubkey_algo, "ssh-rsa") != 0) {
```
Castaglia, line 5699: Ideally, the `pubkey_algo` value matches the `sig_type` here, i.e. "rsa-sha2-256".  However, the client could also be sending a `pubkey_algo` value of "ssh-rsa", as the public key is an RSA key in both cases, and it is only the signature algorithm that differs (see RFC 8332).

So here, if `pubkey_algo` and `sig_type` do not match, we also should allow a `pubkey_algo` value of "ssh-rsa".

PR #1556, opened 2022-11-22
Add CIFuzz GitHub action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where ProFTPD already runs. CIFuzz can help detect regressions and catch fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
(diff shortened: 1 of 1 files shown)


Castaglia:
Thanks for this!  Before I merge this, I'd like to spend a little more time working on fuzz tests for ProFTPD, so that I personally know how to write them better.  Hopefully I can spend some time on that in the coming weeks.  That way, the coverage we get from fuzz tests _and_ with this CI workflow help catch regressions.

Issue #1114, opened 2020-09-20
include others returns from execution in ftpctl


willyamcts:
Hi, 

Return code 0 is used in all ftpctl runs. It is not bug, is enhancement.

Using the "ftpdctl" utility does not return the default error code so that it can be handled via script. The default is to return 0 for the returns that should be 1, 127 and 128 (it was the ones I tested). The closing (cod 130) is ok:

// **Return ok**
```
deb1:~# ftpdctl status 0.0.0.0:22
^C
deb1:~# echo $?
130
```

// **non possible desconnect client because it is not connected (cod. 1)**
```
deb1:~# ftpdctl kick user dave
deb1:~# ftpdctl: user 'user' not connected
deb1:~# echo $?
0
```

// **non possible desconnect client because it is not connected (cod. 1)**
```
deb1:~# ftpdctl permit user user1
ftpdctl: permit request unnecessary
ftpdctl: no users are banned
deb1:~# echo $?
0
```

// **Invalid command (cod. 127) or invalid argument (cod. 128)**
```
deb1:~# ftpdctl **statusA** 0.0.0.0:22
ftpdctl: unsupported action requested
deb1:~# echo $?
0
```

// **Invalid argument (cod. 128) maybe**
```
deb1:~# ftpdctl status 
ftpdctl: status: missing required parameters
deb1:~# echo $?
0
```

Castaglia:
Why these particular exit codes/return values?  I agree this sounds like a useful enhancement.  I am just curious why you specifically mentioned 127, 128, 130 as values, _vs_ 1, 2, 3, _etc.

willyamcts:
> Why these particular exit codes/return values? I agree this sounds like a useful enhancement. I am just curious why you specifically mentioned 127, 128, 130 as values, _vs_ 1, 2, 3, _etc_.

Hi, thanks for accept.

Based from [documentation ](https://tldp.org/LDP/abs/html/exitcodes.html) the project Linux, these are the standard error codes. The goal is to maintain the standard, for those who need a quick search to find the meaning of the code or even for those who already know the meaning of the standard codes.

Castaglia:
[...]

PR #1583, opened 2023-01-08
Issue #1114: Provide richer exit status values for ftpdctl for better…

Castaglia:
… scripting.

### DIFF
--- a/src/ftpdctl.c
+++ b/src/ftpdctl.c
@@ -343,6 +343,11 @@ int main(int argc, char *argv[]) {
     exit(1);
   }
 
+  if (verbose == TRUE) {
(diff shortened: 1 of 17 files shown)


src/ftpdctl.c
```diff
@@ -355,7 +355,10 @@ int main(int argc, char *argv[]) {
   (void) close(sockfd);
   destroy_pool(ctl_pool);
 
-  return 0;
+  /* Controls action status values are zero or negative, thus we need to
+   * convert them to appropriate Unix exit status values.
+   */
+  return (status * -1);
```
Castaglia, line 361: Here is where we actually _use_ the richer Controls API status values for the exit status.

PR #1602, opened 2023-02-20
- I believe those file should be generated at build time

Sashan:
I've just noticed Solaris has stopped shipping `/usr/share/locale/en_US/LC_MESSAGES/proftpd.mo` file. After looking at what's going on there I've found out `msgfmt ` on Solaris build system is not able to process file `en_US.po` file which is found in .tar.gz file which comes from `proftpd-1.3.8.tar.gz`.

When I regenerate `en_US.po` file from existing proftpd.pot file things seem to work as expected and I obtain desired en_US.po file which `msgfmt` is able to process.

`msgfmt` in question comes from `text/locale` package
```
-bash-5.2$ pkg info text/locale
             Name: text/locale
          Summary: System Localization
      Description: localization utilities and C locale (POSIX default)
                   definitions
         Category: System/Localizations
            State: Installed
        Publisher: solaris
          Version: 11.4
           Branch: 11.4.57.0.0.140.0
   Packaging Date: Thu Feb  9 12:45:56 2023
Last Install Time: Mon Apr  1 07:53:56 2019
 Last Update Time: Wed Feb 15 16:22:17 2023
             Size: 526.62 kB
             FMRI: pkg://solaris/text/locale@11.4-11.4.57.0.0.140.0:20230209T124556Z
```

I'm not too much experienced with `msgfmt`/locale handling so I'm not sure if removing .po file here is the right move for everyone. I just want to point out it works for me and solves problem on Solaris.


Castaglia:
@Sashan Are there any errors/messages emitted by the Solaris `msgfmt`, when it attempts to process the provided `.po` files?

Sashan:
> @Sashan Are there any errors/messages emitted by the Solaris `msgfmt`, when it attempts to process the provided `.po` files?

unfortunately not. the tool just dies without producing anything (output file or error message) furthermore it does exit with 0 indicating success, hence make process does not stop. I'll try to figure out where the tool is coming from and let you know.

Castaglia:
In the mean time, as a workaround, for your Solaris builds you might try something like:
```
./configure ...
cd locale/
make clean
make locale
cd ../
make
make install
```
By doing a `make clean` in the `locale/` directory, the `*.mo` files should be deleted.  The subsequent `make locale` command should re-generate them, using the Solaris `msgfmt`.

Sashan:
> In the mean time, as a workaround, for your Solaris builds you might try something like:
> 
> ```
> ./configure ...
> cd locale/
> make clean
> make locale
> cd ../
> make
> make install
> ```
> 
> By doing a `make clean` in the `locale/` directory, the `*.mo` files should be deleted. The subsequent `make locale` command should re-generate them, using the Solaris `msgfmt`.

yes this is the exact workaround I did.

the msgfmt which is in default path on build machine comes from Solaris. It's the legacy msgfmt which should be used just for core-os. For stuff we build in userland consolidation (think of *BSD ports) we should be using gnu version of msgfmt which is found in `/usr/gnu/bin`. The build environment must be set properly in Makefile. I did not know that.

Sorry for extra noise. this pull request should be closed.

PR #1605, opened 2023-02-22
Add C_USER POST_CMD to mod_ifsession to allow per-user SFTPAuthTypes to be set

offsides:
SFTPAuthTypes should be able to be set on a per-user basis so that different users with different security requirements can be accommodated rather than forcing all users to have the same auth requirements.  This adds functionality to mod_ifsession.c to check after the SFTP user has been accepted but before authenticating, and if there's a user or group match (<IfUser> or <IfGroup>).  I'm not sure how to do a class match for this, or how useful that would be either, but at least the group matching is working (I haven't tested a user config, as group is more logical for my setup).  This was basically created by copying and munging the code in the ifsess_pre_pass and ifsess_post_pass functions, so it's quite possible that there's some tightening up that can be done as well.

### DIFF
--- a/contrib/mod_ifsession.c
+++ b/contrib/mod_ifsession.c
@@ -1088,6 +1088,234 @@ MODRET ifsess_post_pass(cmd_rec *cmd) {
   return PR_DECLINED(cmd);
 }
 
+MODRET ifsess_post_user(cmd_rec *cmd) {
+  register unsigned int i = 0;
+  config_rec *c = NULL;
+  char *smtpauthmethods = NULL;
+  pool *tmp_pool = make_sub_pool(session.pool);
+  array_header *group_remove_list = make_array(tmp_pool, 1,
+    sizeof(config_rec *));
+  array_header *user_remove_list = make_array(tmp_pool, 1,
+    sizeof(config_rec *));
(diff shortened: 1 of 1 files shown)


Castaglia:
Here's how I might do/implement this kind of thing: https://github.com/proftpd/proftpd/pull/1606

With that patch, in your SFTP `<VirtualHost>` configurations, you would use "IfSessionOptions PerUnauthenticatedUser", _e.g._:
```
<VirtualHost 1.2.3.4>
  SFTPEngine on
  ...

  <IfModule mod_ifsession.c>
    IfSessionOptions PerUnauthenticatedUser
    ...
  </IfModule>
</VirtualHost>
```

I refactored some of the `mod_ifsession` internals to cut down on duplication where possible.  And it might even be more legible this way.

offsides:
I knew there had to be a better way to do it, but my limited hacking of proftpd hasn't yet given me the level of understanding that you (obviously) do.  Taking a quick look a at the diff, I see what you did and it makes a lot of sense.  I'll try applying it locally and see how it goes.  Thanks a bunch!

Castaglia:
Closing this, in favor to the other PR.  Thanks!

PR #1606, opened 2023-02-23
Implement a way to have `mod_ifsession` work its magic with condition…

Castaglia:
…al configurations using the _unauthenticated_ username sent by a client.

This should not be a common configuration, but there are valid cases which need this kind of functionality.


offsides:
Initial testing with SFTPAuthMethods works the same way my patch did, but your way looks a lot better and more generic.  Thanks much!

Issue #1604, opened 2023-02-22
Odd (Erroneous?) error from mod_sftp_sql with 2-factor SFTP auth


offsides:
### What I Did

FOr a user configured to use publickey+password auth, after matching the public key (obtained using mod_sftp_sql), before it asks for the password, I'm getting the following error in the SFTP logs:

```
https://raw.githubusercontent.com/offsides/proftpd/master/contrib/mod_ifsession.c
```

It doesn't cause authentication to fail, and I don't see if using publickey auth alone, so I suspect that something erroneous is happening.  But I don't understand what you're doing with the keys to know for certain.

### What I Expected/Wanted

No error messages in the logs

### ProFTPD Version and Configuration

Please help us reproduce the problem/issue you are encountering.  To do this,
we need to know which version of ProFTPD you are using, how it was built,
_etc_.  The following command is an easy way to get all of this information:
```bash
proftpd -V
Compile-time Settings:
  Version: 1.3.8 (stable)
  Platform: LINUX [Linux 5.14.0-162.12.1.el9_1.x86_64 x86_64]
  OS/Release:
    NAME="Red Hat Enterprise Linux"
    VERSION="9.1 (Plow)"
    ID="rhel"
    ID_LIKE="fedora"
    VERSION_ID="9.1"
    PLATFORM_ID="platform:el9"
    PRETTY_NAME="Red Hat Enterprise Linux 9.1 (Plow)"
    CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
    
    REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
    REDHAT_BUGZILLA_PRODUCT_VERSION=9.1
    REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
    REDHAT_SUPPORT_PRODUCT_VERSION="9.1"
  Built: Fri Feb 3 2023 00:00:00 UTC
  Built With:
[...]

PR #1608, opened 2023-02-24
Issue #1604: Tweak handling of keys that failed verification in `mod_…

Castaglia:
…sftp_sql`, such that we fall back to attempting verification of the keys as "raw" encoding only if the verification failure was indeed do to format/parsing issues, _vs_ key mismatches.

### DIFF
--- a/contrib/mod_sftp_sql.c
+++ b/contrib/mod_sftp_sql.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD: mod_sftp_sql -- SQL backend module for retrieving authorized keys
- * Copyright (c) 2008-2022 TJ Saunders
+ * Copyright (c) 2008-2023 TJ Saunders
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -50,7 +50,7 @@ struct sqlstore_data {
 
 static const char *sqlstore_user = NULL;
 
-static const char *trace_channel = "ssh2";
+static const char *trace_channel = "sftp.sql";
 
 static cmd_rec *sqlstore_cmd_create(pool *parent_pool, unsigned int argc, ...) {
   register unsigned int i = 0;
@@ -386,7 +386,7 @@ static struct sqlstore_key *sqlstore_get_key_rfc4716(pool *p, char **blob,
       break;
 
     } else {
-      if (key) {
+      if (key != NULL) {
         if (strstr(line, ": ") != NULL) {
           if (strncasecmp(line, "Subject: ", 9) == 0) {
             key->subject = pstrdup(p, line + 9);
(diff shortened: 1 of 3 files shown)


contrib/mod_sftp_sql.c
```diff
@@ -50,7 +50,7 @@ struct sqlstore_data {
 
 static const char *sqlstore_user = NULL;
 
-static const char *trace_channel = "ssh2";
+static const char *trace_channel = "sftp.sql";
```
Castaglia, line 53: Note that I created a separate trace channel for this module.  I've found that, as I develop more modules, having focused trace channels whose names align more closely with their module makes the logging easier to follow.

contrib/mod_sftp_sql.c
```diff
@@ -465,23 +465,29 @@ static int sqlstore_verify_key_raw(pool *p, struct sqlstore_data *store_data,
       store_data->select_query, strerror(errno));
 
   } else if (res == FALSE) {
-    (void) pr_log_writefile(sftp_logfd, MOD_SFTP_SQL_VERSION,
+    pr_trace_msg(trace_channel, 3,
```
Castaglia, line 468: These "key does not match" log messages might be a bit noisy for `SFTPLog`, so I've moved them to trace logging.

offsides:
This looks good.  The one thing I will point out is that mod_sftp is still suggesting that the key is not RFC4716 formatted when it doesn't match, in mod_sftp/keystore.c on line 246.  I'm seeing the following in my logs:
```
2023-02-24 11:28:03,613 mod_sftp/1.1.1[708547]: public key SHA256 fingerprint: d6:0f:e8:3f:c7:03:55:e1:8e:83:1d:07:e6:f7:7e:51:e8:ab:9f:e6:52:16:18:7e:08:b5:43:4b:91:93:41:58
2023-02-24 11:28:03,615 mod_sftp_sql/0.4[708547]: SQLNamedQuery 'get-user-authorized-keys' returned 1 row
2023-02-24 11:28:03,615 mod_sftp/1.1.1[708547]: no matching public key found for user 'sftptest2' in '/get-user-authorized-keys'; perhaps keys are not RFC4716-formatted
```
So mod_sftp_sql is properly decoding the key and reporting normally, but mod_sftp either can't tell that the key was formatted properly or isn't checking for that difference.  Not a big deal, but perhaps something to add to the rainy-day todo list. :)

Otherwise, I think this resolves the erroneous error about decoding raw keys nicely.  Thanks!

Castaglia:
> This looks good. The one thing I will point out is that mod_sftp is still suggesting that the key is not RFC4716 formatted when it doesn't match, in mod_sftp/keystore.c on line 246. I'm seeing the following in my logs:
> ```
> 2023-02-24 11:28:03,613 mod_sftp/1.1.1[708547]: public key SHA256 fingerprint: d6:0f:e8:3f:c7:03:55:e1:8e:83:1d:07:e6:f7:7e:51:e8:ab:9f:e6:52:16:18:7e:08:b5:43:4b:91:93:41:58
> 2023-02-24 11:28:03,615 mod_sftp_sql/0.4[708547]: SQLNamedQuery 'get-user-authorized-keys' returned 1 row
> 2023-02-24 11:28:03,615 mod_sftp/1.1.1[708547]: no matching public key found for user 'sftptest2' in '/get-user-authorized-keys'; perhaps keys are not RFC4716-formatted
> ```
> 
> So mod_sftp_sql is properly decoding the key and reporting normally, but mod_sftp either can't tell that the key was formatted properly or isn't checking for that difference. Not a big deal, but perhaps something to add to the rainy-day todo list. :)

This is due to the separation of concerns.  Specifically, `mod_sftp` only knows that no matching key was found; it's the provider module (`mod_sftp_sql` in this case, but it could be some other keystore provider) that knows more about _why_ the matching/verification failed.

PR #1614, opened 2023-03-01
Deduplicate logfile timestamp generation

offsides:
Timestamp generation used duplicated code for stderr and syslog/logfile that wasn't quite identical but should have been.  This moves the code to a separate function and then calls said function in both places.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -489,6 +489,35 @@ void log_discard(void) {
   syslog_discard = TRUE;
 }
 
+char *log_time(char *buf, int max) {
+      pool *tmp_pool;
+      struct timeval now;
+      struct tm *tm = NULL;
+      unsigned long millis;
+      int buflen = 0, len = 0;
+
+      tmp_pool = make_sub_pool(permanent_pool);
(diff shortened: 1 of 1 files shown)


src/log.c
```diff
@@ -489,6 +489,35 @@ void log_discard(void) {
   syslog_discard = TRUE;
 }
 
+char *log_time(char *buf, int max) {
+      pool *tmp_pool;
```
Castaglia, line 493: I generally prefer 2-space indentation (and no tabs); PRs are more likely to be accepted if they are consistent with the coding style of the rest of the codebase.

Castaglia:
Any objections if I go with https://github.com/proftpd/proftpd/pull/1616 instead?  It's slightly different, in that it _a)_ uses the existing buffer instead of allocating a new one on the stack, and _b)_ it relies on the callers to allocate/destroy the pool to use.  Otherwise, it should be the same.

Thanks for the nudge to do some DRYing up of that code; I'm sure there's much to be done.

PR #1616, opened 2023-03-04
Refactor some duplicated timestamp logging code into a reusable funct…

Castaglia:
…ion, to make things more DRY.


offsides:
Looks reasonable to me, and you know the ins and outs of your code better than I do, so I'm assuming your way of doing it is better than mine.  It certainly resolves the duplicated code issue nicely.

PR #1620, opened 2023-03-08
Update mod_sftp.html to document ExtendedLog

offsides:
Add a table of classes and commands for ExtendedLog with mod_sftp, along with a few notes about how certain things interact.

### DIFF
--- a/doc/contrib/mod_sftp.html
+++ b/doc/contrib/mod_sftp.html
@@ -1692,6 +1692,116 @@ What about the <code>READ</code> and <code>WRITE</code> command groups?
   </li>
 </ul>
 
+<p><a name="ExtendedLog">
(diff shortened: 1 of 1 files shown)


Castaglia:
Thanks for the doc PR!  I'll take a look to see if it documents what we want the expected behavior (_vs_ some of the built-up crufty behavior) to be.

Castaglia:
Hmm.  I'm wondering whether this information should be in the `mod_sftp` docs, or part of the `ExtendedLog` documenation?  In which location would you have first looked for this info?

Castaglia:
Here's a slightly different version of this (thanks!), based on the recent changes: https://github.com/proftpd/proftpd/pull/1632

Issue #1617, opened 2023-03-06
ExtendedLog SSH, SFTP classes not working as expected


offsides:
Broken off from #1611, as this appears to be a genuine bug rather than a question/difference of philosophies.

### What I Did

I tried logging ALL,!SSH to get everything but CL_SSH messages, but it also suppressed CL_SFTP messages.

### What I Expected/Wanted

I was hoping to get the CL_SFTP log messages in my log file.  Based on playing with adding SFTP, SSH, !SFTP and !SSH to the ExtendedLog statements, it appears that the SFTP class has no impact on ExtendedLogs, and that CL_SFTP and CL_SSH are both included or excluded based on the SSH class.  Digging the code doesn't look like it should work that way, so I really am at a loss as to why it's doing so.

### ProFTPD Version and Configuration

Proftpd 1.3.8
```
### BEGIN DEVELOPMENT ONLE
LoadModule mod_ctrls_admin.c
ControlsEngine on
ControlsACLs all allow user root
AdminControlsEngine on
AdminControlsACLs all allow user root
### END DEVELOPMENT ONLE

# Load modules to extend ProFTPD.  Note that the order below MUST be kept!
LoadModule              mod_sql.c
LoadModule              mod_sql_mysql.c
LoadModule              mod_exec.c
LoadModule              mod_exec_mqueue.c
LoadModule              mod_sftp.c
LoadModule              mod_sftp_sql.c
LoadModule              mod_wrap2.c
LoadModule              mod_wrap2_sql.c
LoadModule              mod_unique_id.c
LoadModule              mod_ifsession.c

# FTP Server (base server settings)

# Basic server setup for a standalone server.
ServerType              standalone
# Port 21 is the standard FTP port.
Port                    21
# Specify the server IP.
DefaultAddress          <IP>
# Specify the name of the server.
ServerName              <FTP SERVERNAME>
# Disable IPv6
UseIPv6                 off
# Disable Reverse DNS lookups - we only care about the client IP address.
UseReverseDNS           off
# To prevent DoS attacks, set the maximum number of child processes
# to 30.  If you need to allow more than 30 concurrent connections
# at once, simply increase this value.  Note that this ONLY works
# in standalone mode, in inetd mode you should use an inetd server
# that allows you to limit maximum number of processes per service
# (such as xinetd).
MaxInstances            101

SystemLog		/var/log/proftpd/ftpserver.log
SyslogLevel		debug

# Global settings (FTP and SFTP)
<Global>
[...]

PR #1621, opened 2023-03-11
Issue #1617: The SSH, SFTP `ExtendedLog` classes have been broken for…

Castaglia:
… a while, due to filtering in the Jot API of commands lacking known command IDs.

The core ProFTPD engine has command IDs for FTP commands, but not for the SSH, SFTP requests supported by the `mod_sftp` module.  The workaround is for `mod_sftp` to manually assign a fake command ID, such that the Jot API will think these commands are known, and thus handle them as expected.

### DIFF
--- a/contrib/mod_sftp/fxp.c
+++ b/contrib/mod_sftp/fxp.c
@@ -4799,6 +4799,7 @@ static int fxp_handle_ext_fsync(struct fxp_packet *fxp,
 
   cmd = fxp_cmd_alloc(fxp->pool, "FSYNC", args);
   cmd->cmd_class = CL_MISC|CL_SFTP;
+  cmd->cmd_id = SFTP_CMD_ID;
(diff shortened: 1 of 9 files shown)


contrib/mod_sftp/fxp.c
```diff
@@ -5162,7 +5165,7 @@ static int fxp_handle_ext_posix_rename(struct fxp_packet *fxp, char *src,
     src, dst);
 
   cmd = fxp_cmd_alloc(fxp->pool, "RENAME", args);
-  cmd->cmd_class = CL_MISC|CL_SFTP;
+  cmd->cmd_class = CL_MISC|CL_WRITE|CL_SFTP;
```
Castaglia, line 5168: Here we try to align the command classes for SFTP requests more closely with the command classes assigned to their FTP equivalents, where possible.

PR #1629, opened 2023-03-13
Add %S to mod_exec for sftp-non-fatal tagging

offsides:
This adds %S as a mod_exec substitution, which is replaced by the "mod_sftp.nonfatal-attempt" note or an empty string if it doesn't exist.  Thus an empty string is a fatal error (i.e., failed login) whereas a non-empty string indicates a non-fatal error as part of the login process.

I also added the %V and %{time:} variables to the docs, as they are missing and I use at least one of those (time).  I did not add %l (ident info) since mod_ident is deprecated anyway, or %{env:} because it doesn't appear to work right now.

### DIFF
--- a/contrib/mod_exec.c
+++ b/contrib/mod_exec.c
@@ -1062,6 +1062,15 @@ static const char *exec_subst_var(pool *tmp_pool, const char *varstr,
     }
   }
 
+  ptr = strstr(varstr, "%S");
+  if (ptr != NULL) {
+    const void *sftp_nonfatal_note;
+
+    sftp_nonfatal_note = pr_table_get(cmd->notes, "mod_sftp.nonfatal-attempt", NULL);
+    varstr = sreplace(tmp_pool, varstr, "%S",
+      sftp_nonfatal_note ? sftp_nonfatal_note : "", NULL);
+  }
+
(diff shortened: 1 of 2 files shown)


Castaglia:
Hmm.  Better would be for `mod_exec` to support `${note:...}` style substitutions, as is supported elsewhere as well.

Castaglia:
> Hmm. Better would be for `mod_exec` to support `${note:...}` style substitutions, as is supported elsewhere as well.

I've filed https://github.com/proftpd/proftpd/issues/1630 to address this.

Castaglia:
With the fix for Issue #1630 now being merged, I think you'll find that `mod_exec` works as you expect.

For this specific case of the `mod_sftp.nonfatal-attempt` note, you'll want to use a variable of `%{note:mod_sftp.nonfatal-attempt}` in  your `mod_exec` commands.  (Yes, the `%{note:...}` and `%{var:...}` syntax aren't documented well for `LogFormat`; in my mind these fall into the category of "advanced usage", which hasn't needed that much public documentation as yet.)

Castaglia:
In light of Issue #1630, should I close this PR now?

Issue #1611, opened 2023-02-28
Is there a table of command class/command mapping for SFTP extended logging


offsides:
The ExtendedLog docs for mod_log have a table that maps Command Classes to specific FTP commands for logging purposes.  I've noticed that there are additional commands logged for mod_sftp, but they aren't documented for for logging purposes. Is there a similar table for mod_sftp so that I can craft my logging statements to be (roughly) equivalent for FTP and SFTP without logging lots of additional stuff that I don't necessarily want for just SFTP sessions?

As an example, here's logs from an FTP session and an SFTP session perfrming the exact same commands after login (the only difference being key login for SFTP).
FTP:
```
127.0.0.1 [28/Feb/2023:10:33:15 -0500] Y/4euawZ115/AAABABNpjD3k - "USER jsmftptest" 331 -
127.0.0.1 [28/Feb/2023:10:33:16 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "PASS (hidden)" 230 -
127.0.0.1 [28/Feb/2023:10:33:16 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "SYST" 215 -
127.0.0.1 [28/Feb/2023:10:33:19 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "PASV" 227 -
127.0.0.1 [28/Feb/2023:10:33:19 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "LIST -al" 226 260
127.0.0.1 [28/Feb/2023:10:33:32 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "TYPE I" 200 -
127.0.0.1 [28/Feb/2023:10:33:32 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "PASV" 227 -
127.0.0.1 [28/Feb/2023:10:33:32 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "STOR trytest.pl" 226 444
127.0.0.1 [28/Feb/2023:10:33:35 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "MKD test" 257 -
127.0.0.1 [28/Feb/2023:10:33:36 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "CWD test" 250 -
127.0.0.1 [28/Feb/2023:10:33:37 -0500] Y/4euawZ115/AAABABNpjD3k jsmftptest "TYPE A" 200 -
[...]

PR #1632, opened 2023-03-15
Issue #1611: Provide a table showing which SSH, SFTP requests map to …

Castaglia:
…which `ExtendedLog` command classes.


offsides:
This looks good - I'm always happy to defer to your way of doing things since you know the whole ecosystem better than I do.

Issue #1630, opened 2023-03-13
Update mod_exec to use the Jot API for resolving variables


Castaglia:
The scope of this ticket is to update the `mod_exec` module's variable substitution code [here](https://github.com/proftpd/proftpd/blob/master/contrib/mod_exec.c#L871) to instead use the Jot API, so that it automatically benefits from new variable additions there.

offsides:
I'm not really familiar with the jot API, but in principle this sounds like an excellent idea to centralize things and avoid code duplication.

Castaglia:
> Add %S to mod_exec for sftp-non-fatal tagging #1629

The Jot API is ProFTPD's internal API that is used to resolve all of the `LogFormat` variables when writing _e.g._ `ExtendedLog` entries.  It is also used, to support those `LogFormat` variables in things like `SQLNamedQuery`, and in logging modules like `mod_redis` and its ilk.  So updating `mod_exec` to use this API as well (which it should be) is a Good Thing.  I consider the fact that `mod_exec` does _not_ use this API to be tech debt to be retired; it's one less place that implements its own handling of these variables.

Castaglia:
Addressed in master now.

PR #1633, opened 2023-03-16
Issue #1630: Update the `mod_exec` module to use the Jot API for supp…

Castaglia:
…orting the normal range of `LogFormat` variables using the shared API for doing so.


include/jot.h
```diff
@@ -76,6 +76,7 @@
 #define PR_JOT_LOGFMT_TIME_KEY		"local_time"
 #define PR_JOT_LOGFMT_UID_KEY		"uid"
 #define PR_JOT_LOGFMT_USER_KEY		"user"
+#define PR_JOT_LOGFMT_VAR_KEY		"VAR:"
```
Castaglia, line 79: The Jot API didn't actually support values that used the Var API, surprisingly.  This work revealed a gap, so part of this PR is adding that support to the Jot API.

PR #1641, opened 2023-03-25
Reduce the number of times any configured `AuthUserFile`, `AuthGroupF…

Castaglia:
…ile` files are scanned during authentication by refactoring the `mod_ifsession` fix for [Bug#3882](http://bugs.proftpd.org/show_bug.cgi?id=3882) to happen just before any chroot, rather than during the `PRE_CMD` handling of a `PASS` command.

### DIFF
--- a/contrib/mod_ifsession.c
+++ b/contrib/mod_ifsession.c
@@ -690,6 +690,172 @@ static int ifsess_sess_merge_user(pool *p) {
   return 0;
 }
 
+static void ifsess_sess_process_displayfiles(void) {
+  config_rec *c;
+  char *displaylogin = NULL;
+  xaset_t *config_set = NULL;
+
+  /* Look for a DisplayLogin file which has an absolute path.  If we find one,
+   * open a filehandle, such that that file can be displayed even if the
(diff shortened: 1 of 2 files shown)


contrib/mod_ifsession.c
```diff
@@ -1222,6 +1179,8 @@ MODRET ifsess_post_user(cmd_rec *cmd) {
 
 static void ifsess_chroot_ev(const void *event_data, void *user_data) {
   ifsess_home_dir = (const char *) event_data;
+
+  ifsess_sess_process_displayfiles();
```
Castaglia, line 1183: This works because at the time the `core.chroot` event is published, authentication (and thus reading of the `AuthUserFile`, `AuthGroupFile` files) will have already occurred, and the necessary applicable user/group lookups will have occurred; we don't need to do those lookups again ourselves.

Previously, the `mod_ifsession` _would_ do those lookups as part of the `PRE_CMD` phase of the `PASS` command, which happens _before_ the core `mod_auth` module does _its_ lookups -- resulting in redundant user/group membership lookups, redundant `AuthUserFile`, `AuthGroupFile` scans.

Issue #1639, opened 2023-03-23
Gracefully handle directory creation requests when the directory already exists


zrlw:
### What I Did

we have multiple clients to upload files and  the directories on proftpd server are created as needed.  Only one client will get successful return if mutiple clients send making same directory request simutaneously, others get failure return immediately, 
 but the directory is not created yet, so we must wait some time after catching the exception which wrapped mkdir failure like this:
```
try {
    ftp.mkdir(needed_subdir);
    ftp.cd(needed_subdir);
} catch (SomeExceptionWhichWrappedMakeDirectoryFailure) {
    Thread.sleep(waitTime);
    ftp.cd(needed_subdir); 
    ...
}
```
but we could not determine the wait time.

### What I Expected/Wanted
cache directory creation request, clear up cache when the creation process is completed.
if the same directory creation request is existed in the cache,  don't return failure msg until the cache is cleared.
    
### ProFTPD Version
1.3.7d

Castaglia:
What is the error that the clients see, when their directory creation commands fail?  Can you provide the ProFTPD [debug logging](http://www.proftpd.org/docs/howto/Debugging.html) for these sessions, so that we can see what the server-side errors might be?

zrlw:
debug logging may not be enabled in product environment, i'll check it on working day.
some client will get 'No such file' from proftpd for request 'cd xxx' even after waiting 3s, all clients login proftpd by the same account. i doublt it might be caused by FSCachePolicy mechanism:
![proftpd](https://user-images.githubusercontent.com/40652892/227750471-74922b8a-b825-4dd2-b965-1b7b05b2b96a.png)

zrlw:
our java clients use ChannelSftp of jsch.

Castaglia:
Hmm.  I'm not sure this is a server-side issue, honestly.

ProFTPD is not a multi-threaded server; it uses separate _processes_, not threads, for each client connection.  Each session process is completely independent, and has no knowledge of what any other session process is doing.

It sounds like your application has multiple separate independent connections to ProFTPD (which means each one would have its own separate server-side process).  And your application ends up having these connections "race" each other to create the same directory.  The server-side processes have no knowledge of what any other session is doing, so they will each independently try to reference (then create) the same directory.

Without seeing the specific server-side error for the `MKD` command failure, I cannot offer more useful information at this time.

zrlw:
there are no TraceLog setting in our /etc/proftpd.conf.

[...]

PR #1645, opened 2023-03-28
Issue #1639: When directory creation requests (FTP and SFTP) are rece…

Castaglia:
…ived for directories that already exist (leading to `EEXIST` errors), gracefully handle such errors as "success".

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -48,6 +48,8 @@
   ExtendedLog class.
 - Issue 1630 - Update mod_exec to use the Jot API for resolving variables.
 - Issue 1636 - Use system TCP backlog value by default.
+- Issue 1639 - Gracefully handle directory creation requests when the directory
+  already exists.
 
 1.3.8 - Released 04-Dec-2022
 --------------------------------
--- a/contrib/mod_sftp/fxp.c
+++ b/contrib/mod_sftp/fxp.c
@@ -8664,7 +8664,7 @@ static int fxp_handle_lstat(struct fxp_packet *fxp) {
 static int fxp_handle_mkdir(struct fxp_packet *fxp) {
   unsigned char *buf, *ptr;
   char *attrs_str, *cmd_name, *path;
-  struct stat *attrs, st;
+  struct stat *attrs;
   int have_error = FALSE, res = 0;
   mode_t dir_mode;
   uint32_t attr_flags, buflen, bufsz, status_code;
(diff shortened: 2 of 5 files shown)


zrlw:
```
res = pr_fsio_smkdir(cmd->tmp_pool, dir, 0777, session.fsuid, session.fsgid);
if (res < 0) {
    if (errno == EEXIST) {
        // should clear the cache of previous command result like 'No such file' of cd.
        pr_fs_clear_cache2(path);
    } else {
         ...
    }
}
```

zrlw:
if treating EEXIST as success, the cache should be cleared to avoid:
1. cd xxx : cache 'No such file' result.
2. mkdir xxx: treat EEXIST as success but the failure cache is still existed.
3. cd xxx: return cached result 'No such file'

Castaglia:
The session-specific cache of `stat(2)` data _is_ cleared, in the code paths that handle those requests.  As such, it does *not* need to be cleared right after an `EEXIST` error for the `pr_fsio_smkdir()` function.

Castaglia:
Also, when reporting issues, it helps to avoid client-specific commands like "cd", which are not actual protocol-level requests; they are a facade, an illusion that your specific client presents to you, but which mean nothing to the server implementation.

zrlw:
it seemed that the session-specific cache of stat(2) is cleared in pr_fsio_smkdir() only if the directory is created successfully.
next stat(2) will read the session-specific cache first.

zrlw:
stat(2): pr_fsio_stat -> fs_cache_stat -> cache_stat:
```
// get result from stat cache
sc = fs_statcache_get(cache_tab, cache_set, cleaned_path, path_len, now);
if (sc != NULL) {
    // if existed then return session-specific cache 
    return sc->sc_retval;
}
retval = mystat(fs, cleaned_path, st);
// Update the cache
res = fs_statcache_add(cache_tab, cache_set, cleaned_path, path_len, st,
    xerrno, retval, now);
```
> The session-specific cache of stat(2) data is cleared, in the code paths that handle those requests. 

could you remind me the code lines that clear the cache?

zrlw:
create_dir() in mkhome.c might have the same issue.

zrlw:
'EEXIST' error of pr_fsio_smkdir() in mod_sftp/scp.c should be also treated as success.
by the way, scp.c clear the cache before pr_fsio_smkdir().

zrlw:
it might be better to modify pr_fsio_smkdir() and pr_fsio_mkdir() directly if treating EEXIST as success: 
```
if (errno == EEXIST) {
    pr_fs_clear_cache2(path); // in case there are no subsequent cleanup operation, e.g., create_dir() in 'contrib/mod_copy.c'
    return 0;
}
```

PR #1648, opened 2023-03-29
Optimize the `pr_fs_interpolate()` function for the case where the gi…

Castaglia:
…ven path is _just_ `~`, _and_ we know the user and their home directory.


tests/t/lib/ProFTPD/Tests/Modules/mod_ifsession.pm
```diff
@@ -1379,7 +1379,7 @@ sub ifgroup_displaylogin_bug3882 {
     AuthGroupFile => $setup->{auth_group_file},
     AuthOrder => 'mod_auth_file.c',
 
-#    DefaultRoot => '~',
+    DefaultRoot => '~',
```
Castaglia, line 1382: Reverts some additional changes for https://github.com/proftpd/proftpd/pull/1641 which should not have been committed.

PR #1665, opened 2023-05-03
configure: Remove several implicit function declarations

submachine:
During configure, some checks omit the corresponding include. A compiler defaulting to C99 mode could cause those checks to fail since C99 does not allow implicit function declarations.  This commit fixes the same. The configure script is re-generated.


submachine:
Thanks! I'll work on getting this right.

Issue #1663, opened 2023-05-02
Deadlock on MySQL-database when 4-5 simultaneous logins to ProFTPd with updatecount query causing authentication timeout


maltris:
### What I Did

Set up ProFTPd with mod_sql and an update-query to the database after FTP PASS. Run 4-5 parallel client logins on the same user against ProFTPd that PASS within the same second (0-1ms from each other). Get this in the sql log:

```
2023-05-02 07:04:46,817 mod_sql/4.5[501930]: unrecoverable backend error
2023-05-02 07:04:46,817 mod_sql/4.5[501930]: error: '1213'
2023-05-02 07:04:46,817 mod_sql/4.5[501930]: message: 'Deadlock found when trying to get lock; try restarting transaction'
```

And a timed out authentication. 

### What I Expected/Wanted

Get the count of logins and the date of the last login into the database table and run 4-5 simultaneous logins with the same user (in the same second, a few milliseconds from each other) against the ProFTPd.

Example as referenced [here](http://www.proftpd.org/docs/contrib/mod_sql.html):

```
SQLLog PASS updatecount
SQLNamedQuery updatecount UPDATE "count=count+1, accessed=now() WHERE userid='%u'" ftpuser
```

### ProFTPD Version and Configuration

```bash
proftpd -V
```

```
Compile-time Settings:
  Version: 1.3.8 (stable)
  Platform: LINUX [Linux 5.4.0-96-generic x86_64]
  OS/Release:
    NAME="Alpine Linux"
    ID=alpine
    VERSION_ID=3.16.5
    PRETTY_NAME="Alpine Linux v3.16"
  Built: Tue Apr 25 2023 13:56:18 UTC
  Built With:
    configure  '--enable-ctrls' '--enable-openssl' '--enable-quotatab' '--with-modules=mod_tls:mod_quotatab:mod_sql:mod_sql_passwd:mod_quotatab_sql:mod_quotatab_file:mod_ctrls_admin:mod_auth_otp:mod_sql_sqlite:mod_sql_mysql:mod_ban:mod_readme:mod_sftp:mod_sftp_sql:mod_proxy_protocol:mod_ident:mod_ifsession' '--enable-nls'

  CFLAGS: -g2 -O2 -Wall -fno-omit-frame-pointer -fno-strict-aliasing
  LDFLAGS: -Wl,-L$(top_srcdir)/lib,-L$(top_builddir)/lib  -rdynamic  -L/usr/lib/
[...]

PR #1668, opened 2023-05-07
Issue #1663: The `mod_sql_mysql` module should attempt to retry state…

Castaglia:
…ments when it detects a server-side deadlock, at least a few times.


maltris (approved):
The patch is excellently working for me.

PR #1677, opened 2023-05-31
Update fsio.c - if mkdir fails with EEXIST also clear the cache

pzeidler-dlr:
Rationale: if you have two (or more) ftp connections that want to create a mostly identical directory tree, and MKD the same directory single-digit msec apart, the slower client gets a failure and can't work with the new directory. With the patch, the client still gets a failure but can use the directory afterwards.


Castaglia:
Thanks!

Issue #1683, opened 2023-06-06
Out-of-bounds buffer read when handling FTP commands


Marti2203:
### What I Did

Cloned latest version of ProFTPD (5e57d41) in a container running Ubuntu 20.04, configured with flags for ASAN and UBSan. 
Started the tool as such `./proftpd -n -c <SOMEDIR>/basic.conf -X`.
Sent over a sequence of packets, gathered from a fuzz campaign using the `aflnet-replay` utility. 

### What I Expected/Wanted

Expected to have ProFTPD not crash.

### ProFTPD Version and Configuration
```
Compile-time Settings:
  Version: 1.3.9rc1 (git)
  Platform: LINUX [Linux 5.4.0-120-generic x86_64]
  OS/Release:
    NAME="Ubuntu"
    VERSION="20.04.6 LTS (Focal Fossa)"
    ID=ubuntu
    ID_LIKE=debian
    PRETTY_NAME="Ubuntu 20.04.6 LTS"
    VERSION_ID="20.04"
    VERSION_CODENAME=focal
    UBUNTU_CODENAME=focal
  Built: Tue Jun 6 2023 05:18:02 UTC
  Built With:
    configure  '--enable-devel=nodaemon:nofork' 'CC=cc' 'CFLAGS=-fsanitize=address,undefined -g' 'LDFLAGS=-fsanitize=address,undefined'

  CFLAGS: -g2 -fsanitize=address,undefined -g -Wall -fno-omit-frame-pointer -fno-strict-aliasing -Wdangling-else -Wextra -Werror=implicit-function-declaration -Winit-self -Wno-missing-field-initializers -Wno-unused-parameter -Wnull-derefere
nce -fdelete-null-pointer-checks -DPR_DEVEL_NO_FORK -DPR_DEVEL_NO_DAEMON -g3 -O0 -Wcast-align -Wchar-subscripts -Winline -Wstrict-prototypes -Wmissing-declarations -Wnested-externs -Wpointer-arith -Wshadow -Wundef -Wcomment -Wdangling-else
-Wdeclaration-after-statement -Wduplicated-branches -Wfloat-equal -Wformat -Wformat-security -Wimplicit-function-declaration -Wmaybe-uninitialized -Wmissing-braces -Wparentheses -Wpointer-to-int-cast -Wreturn-local-addr -Wstack-protector -W
strict-overflow -Wswitch -Wunreachable-code -Wunused-variable -fstack-protector -fstack-protector-all -fno-common
[...]

PR #1696, opened 2023-08-06
Issue #1683: Avoid an edge case when handling unexpectedly formatted …

Castaglia:
…input text from client, caused by quote/backslash semantics, by skipping those semantics.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -821,14 +821,35 @@ static cmd_rec *make_ftp_cmd(pool *p, char *buf, size_t buflen, int flags) {
     return NULL;
   }
 
+  /* By default, pr_str_get_word will handle quotes and backslashes for
+   * escaping characters.  This can produce words which are shorter, use
+   * fewer bytes than the corresponding input buffer.
+   *
+   * In this particular situation, we use the length of this initial word
+   * for determining the length of the remaining buffer bytes, assumed to
+   * contain the FTP command arguments.  If this initial word is thus
+   * unexpectedly "shorter", due to nonconformant FTP text, it can lead
+   * the subsequent buffer scan, looking for CRNUL sequencees, to access
+   * unexpected memory addresses (Issue #1683).
+   *
+   * Thus for this particular situation, we tell the function to ignore/skip
+   * such quote/backslash semantics, and treat them as any other character
+   * using the IGNORE_QUOTES flag.
+   */
+
   ptr = buf;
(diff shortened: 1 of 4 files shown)


src/main.c
```diff
@@ -821,15 +821,36 @@ static cmd_rec *make_ftp_cmd(pool *p, char *buf, size_t buflen, int flags) {
...
     /* Nothing there...bail out. */
     pr_trace_msg("ctrl", 5, "command '%s' is empty, ignoring", buf);
     errno = ENOENT;
     return NULL;
   }
 
+  /* Note that this first word is the FTP command.  This is why we make
+   * use of the ptr buffer, which advances through the input buffer as
+   * we read words from the buffer.
+   */
```
Castaglia, line 852: These textual breadcrumbs are for my future self, for when I have to again wonder about the reuse of `ptr` in this function, _vs_ using `buf` directly.

src/str.c
```diff
@@ -1270,24 +1270,28 @@ char *pr_str_get_word(char **cp, int flags) {
     }
   }
 
-  if (**cp == '\"') {
-    quote_mode++;
-    (*cp)++;
+  if (!(flags & PR_STR_FL_IGNORE_QUOTES)) {
```
Castaglia, line 1273: This new `IGNORE_QUOTES` flag, and its use in `make_ftp_cmd`, comprise the heart of the fix here.

src/str.c
```diff
@@ -1270,24 +1270,28 @@ char *pr_str_get_word(char **cp, int flags) {
...
-    if (**cp == '\\' && quote_mode) {
-
+    if (**cp == '\\' &&
+        quote_mode == TRUE) {
       /* Escaped char */
       if (*((*cp)+1)) {
-        *dst = *(++(*cp));
+        *dst++ = *(++(*cp));
+        (*cp)++;
+        continue;
```
Castaglia, line 1289: Note that there was an additional latent bug here as well, when handling backslashes, such that the `dst` pointer was *not* being advanced properly after copying the escaped character.

src/str.c
```diff
@@ -1270,24 +1270,28 @@ char *pr_str_get_word(char **cp, int flags) {
...
       }
     }
 
     *dst++ = **cp;
-    ++(*cp);
+    (*cp)++;
```
Castaglia, line 1294: No functional change here, just making the style consistent within this function.

Issue #1678, opened 2023-05-31
High session memory usage caused by SFTP outgoing data buffering


pepi55:
### What I Did

[Using Python's paramiko library](https://pastebin.com/CDNKt1h5), download a file >9GB.
When running this script on the host (downloading from localhost) it causes RAM usage to steadily increase to ridiculously high amounts and 100% cpu spikes on several cores.
In addition, the download takes about 2hrs while the regular sftp client takes about 5-10mins to complete. Sometimes the downloads complete after 2hrs and the file is corrupted/incomplete when using paramiko.

The same script was also tested using sftpgo (default config) and it worked as expected (fast download and complete files).

paramiko version:
paramiko          3.2.0

### What I Expected/Wanted

Download speeds matching unix sftp client, correct download sizes.

### ProFTPD Version and Configuration

```
Compile-time Settings:
  Version: 1.3.7e (maint)
  Platform: LINUX [Linux 5.10.179-166.674.amzn2.x86_64 x86_64]
  Built: Tue May 30 2023 10:43:54 UTC
  Built With:
    configure  '--enable-openssl' '--with-modules=mod_sftp' '--enable-dso' '--with-shared=mod_tls:mod_auth_file' '--disable-auth-pam' '--enable-ctrls' '--enable-ltdl-convenience' '--cache-file=/dev/null' '--srcdir=./' '--enable-devel=nodaemon:nofork:stacktrace'

[...]

PR #1699, opened 2023-08-12
Issue #1678: If our SSH remote window closes, but the client is still…

Castaglia:
… sending channel data messages that will result in more output, start buffering up those incoming messages.

### DIFF
--- a/contrib/mod_sftp/packet.c
+++ b/contrib/mod_sftp/packet.c
@@ -127,6 +127,9 @@ static int packet_poll(int sockfd, int io) {
     "using select(2)", (unsigned long) tv.tv_sec, sockfd,
     io == SFTP_PACKET_IO_RD ? "reading" : "writing");
 
+  /* Clear any possibly stale pointers. */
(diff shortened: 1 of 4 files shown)


contrib/mod_sftp/packet.c
```diff
@@ -127,6 +127,9 @@ static int packet_poll(int sockfd, int io) {
     "using select(2)", (unsigned long) tv.tv_sec, sockfd,
     io == SFTP_PACKET_IO_RD ? "reading" : "writing");
 
+  /* Clear any possibly stale pointers. */
+  session.curr_cmd_rec = NULL;
```
Castaglia, line 131: This change is not necessarily related to the SSH data buffering fixes.  It does, however, quell a stacktrace encountered by folks testing this change.  The stacktrace occurs when logging-related code uses this variable, and it points to a stale `cmd_rec`, one allocated from a pool that has already been destroyed.  Thus we want to clear it here, before reading any network packet, as a best-effort mitigation/cleanup of such stale pointers.

Issue #1694, opened 2023-08-05
SSH key exchanges fail unexpectedly with "unable to write X bytes of raw data" errors due to small ProFTPD buffer


jlecour:
### What I Did

After upgrading from Debian 11 (ProFTPd 1.3.7) to Debian 12 (ProFTPd 1.3.8) I can't connect using SFTP.

In server logs I have this :

```
2023-08-05 18:57:02,433 mod_sftp/1.1.1[581227]: client (<REDACTED_IP>:51472) connected to server (<REDACTED_IP>:2022) [session ID ZM5/Xh+qC30fqgmBAAjea6TB]
2023-08-05 18:57:02,434 mod_sftp/1.1.1[581227]: sent server version 'SSH-2.0-FTP Server Ready'
2023-08-05 18:57:02,435 mod_sftp/1.1.1[581227]: received client version 'SSH-2.0-Cyberduck/8.4.1.38006 (Mac OS X/12.6.7) (x86_64)'
2023-08-05 18:57:02,435 mod_sftp/1.1.1[581227]: handling connection from SSH2 client 'Cyberduck/8.4.1.38006'
2023-08-05 18:57:02,478 mod_sftp/1.1.1[581227]:  + Session key exchange: curve25519-sha256
2023-08-05 18:57:02,478 mod_sftp/1.1.1[581227]:  + Session host key algorithm: ecdsa-sha2-nistp256
2023-08-05 18:57:02,478 mod_sftp/1.1.1[581227]:  + Session server hostkey: ecdsa-sha2-nistp256
2023-08-05 18:57:02,478 mod_sftp/1.1.1[581227]:  + Session client-to-server encryption: aes128-cbc
2023-08-05 18:57:02,478 mod_sftp/1.1.1[581227]:  + Session server-to-client encryption: aes128-cbc
2023-08-05 18:57:02,478 mod_sftp/1.1.1[581227]:  + Session client-to-server MAC: hmac-sha1
2023-08-05 18:57:02,478 mod_sftp/1.1.1[581227]:  + Session server-to-client MAC: hmac-sha1
2023-08-05 18:57:02,478 mod_sftp/1.1.1[581227]:  + Session client-to-server compression: none
2023-08-05 18:57:02,478 mod_sftp/1.1.1[581227]:  + Session server-to-client compression: none
2023-08-05 18:57:02,515 mod_sftp/1.1.1[581227]: message format error: unable to write 104 bytes of raw data (buflen = 96)
[...]

PR #1704, opened 2023-08-17
Issue #1694: Increase the size of the static buffers used for SSH KEX…

Castaglia:
… replies.

While there, refactor these static sizes into a single variable for later tuning.


Castaglia:
I was finally able to locally reproduce the issue reported in the ticket; this PR does *not* address/fix the issue.

PR #1707, opened 2023-09-02
Issue #1694: Use larger (and consistently sized) buffers when calcula…

Castaglia:
…ting the `H` value, to accommodate larger KEXINIT messages.

### DIFF
--- a/contrib/mod_sftp/kex.c
+++ b/contrib/mod_sftp/kex.c
@@ -52,6 +52,12 @@
  */
 #define SFTP_DH_PREF_MIN_LEN	2048
 
+/* Define the size of the buffer we use to compute H.  Note that this buffer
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/kex.c
```diff
@@ -4148,7 +4154,7 @@ static const unsigned char *calculate_curve25519_h(struct sftp_kex *kex,
   unsigned char *buf, *ptr;
   uint32_t buflen, bufsz;
 
-  bufsz = buflen = 4096;
+  bufsz = buflen = SFTP_KEX_H_BUFLEN;
```
Castaglia, line 4157: In the case of this particular ticket/client/configuration, this 4K buffer size was not large enough for the combined KEXINIT messages from Cyberduck and ProFTPD, both supporting many different algorithms.

PR #1708, opened 2023-09-04
Avoid resolving reverse DNS when UseReverseDNS is off

endreferencz:
The code sets pr_netaddr_set_reverse_dns(TRUE) explicitly even if UseReverseDNS is off. As such we need to avoid calling pr_netaddr_get_dnsstr(...) to avoid unwanted reverse DNS lookup. The current code always does the reverse DNS lookup.

### DIFF
--- a/contrib/mod_wrap2/mod_wrap2.c
+++ b/contrib/mod_wrap2/mod_wrap2.c
@@ -365,7 +365,7 @@ static char *wrap2_get_hostname(wrap2_host_t *host) {
       wrap2_log("'UseReverseDNS off' in effect, NOT resolving %s to DNS name "
         "for comparison", pr_netaddr_get_ipstr(session.c->remote_addr));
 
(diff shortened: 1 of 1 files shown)


Castaglia:
This shouldn't be needed, as the `pr_netaddr_get_dnsstr` function internally also checks the same "is UseReverseDNS enabled?" value here:
* https://github.com/proftpd/proftpd/blob/master/src/netaddr.c#L1955

Are you seeing anything else which indicates that DNS resolution is occurring?

Castaglia:
Oh, I see what you're seeing; it's about this bit of code, earlier:
```
    reverse_dns = pr_netaddr_set_reverse_dns(TRUE);
    if (reverse_dns) {
```
I think I'd rather make a slightly different change, to achieve the same effect.

Castaglia:
https://github.com/proftpd/proftpd/pull/1711 should achieve the same thing, if you'd like to verify it?  Thanks!

endreferencz:
Your change works as well, thanks for looking into it!

Issue #1721, opened 2023-09-23
Null pointer reference encountered for FTPS connection due to config parser ignoring Include file problem


dmak:
### Description

After upgrading Debian from bullseye to bookworm (ProFTPD v1.3.7a to v1.3.8) I started to see plenty of messages like below (the configuration was not changed).

```
(192.241.219.39[192.241.219.39]): -----BEGIN STACK TRACE-----
(192.241.219.39[192.241.219.39]): [0] /usr/lib/proftpd/mod_tls.so(+0x19ef5) [0x7efec5f2def5]
(192.241.219.39[192.241.219.39]): [1] /usr/lib/proftpd/mod_tls.so(+0x19ef5) [0x7efec5f2def5]
(192.241.219.39[192.241.219.39]): [2] /usr/lib/proftpd/mod_tls.so(+0x26a7e) [0x7efec5f3aa7e]
(192.241.219.39[192.241.219.39]): [3] proftpd: (accepting connections)(modules_session_init+0x57) [0x55629c61abc7]
(192.241.219.39[192.241.219.39]): [4] proftpd: (accepting connections)(+0x204ad) [0x55629c5ed4ad]
(192.241.219.39[192.241.219.39]): [5] proftpd: (accepting connections)(+0x20ea1) [0x55629c5edea1]
(192.241.219.39[192.241.219.39]): [6] proftpd: (accepting connections)(main+0x5f8) [0x55629c5ebb98]
(192.241.219.39[192.241.219.39]): [7] /lib/x86_64-linux-gnu/libc.so.6(+0x271ca) [0x7efec66501ca]
(192.241.219.39[192.241.219.39]): [8] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x85) [0x7efec6650285]
(192.241.219.39[192.241.219.39]): [9] proftpd: (accepting connections)(_start+0x21) [0x55629c5ec1c1]
(192.241.219.39[192.241.219.39]): -----END STACK TRACE-----
(192.241.219.39[192.241.219.39]): ProFTPD terminating (signal 11)
```

Effectively ProFTPD is not working. Probably the bug is specific to Debian and my local configuration.

Update to ProFTPD from sid (1.3.8+dfsg-8) didn't help.

### ProFTPD Version and Configuration

```
Compile-time Settings:
  Version: 1.3.8 (stable)
  Platform: LINUX [Linux 6.1.0-10-amd64 x86_64]
  OS/Release:
[...]

PR #1722, opened 2023-09-30
Issue #1721: When parsing an Include path (such as a directory), make…

Castaglia:
… sure we halt parsing on errors, and not ignore/continue parsing.

This will help ensure that such configuration errors are caught early on, and addressed, rather than being silently ignored and causing problems later.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -82,6 +82,8 @@
   mod_ldap.
 - Issue 1712 - SFTP algorithm settings in <Global> section not being used.
 - Issue 1433 - Update upload progress in scoreboard when throttling.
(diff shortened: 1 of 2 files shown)


Castaglia:
FYI, @pghmcfc @hpreusse this patch will change the ProFTPD behavior when parsing config files, which may reveal previously-silently-broken configs.  Per my comments [here](https://github.com/proftpd/proftpd/issues/1721#issuecomment-1741782663), while this may cause backward compatibility issues, I'm not sure it'd be a good idea to provide a way to restore this "ignore broken Include files" behavior.

Issue #1746, opened 2023-11-13
Error resolving DNS name for implicit "server config" vhost leads to DelayTable not being found


ia2115:
### What I Did

freebsd 13.2 
```
pkg install proftpd
1.3.8_2
```
when i try to start it i get:
```
mod_delay/0.8: unable to open DelayTable '/var/run/proftpd.delay': No such file or directory
```
if i try create manually, i get:
```
 mod_delay/0.8: unable to load DelayTable '/var/run/proftpd.delay' (fd 3) into memory: Invalid argument
```
i try disable the module with the instructions from this:
http://www.proftpd.org/docs/directives/configuration_full.html#DELAYENGINE instructions

but nothing i do, changes that i cant start proftpd.

i have it running fine on older freebsd 12.2, same version and that file is nowhere on fs i searched.. 
i tried use default conf with a VirtualHost as only add, i get this err.
i reused working conf from older pc, and same err ;) 

im a bit puzzled how i can make proftpd work on my new host.

### What I Expected/Wanted

proftpd to just work, og instructions to tell me i could create file myself and how or i could tell it to ignore this

### ProFTPD Version and Configuration

# proftpd -V
```
Compile-time Settings:
  Version: 1.3.8 (stable)
  Platform: FREEBSD13 (FREEBSD13_2) [FreeBSD 13.2-RELEASE-p2 amd64]
  OS/Release:
    NAME=FreeBSD
    VERSION="13.2-RELEASE"
    VERSION_ID="13.2"
    ID=freebsd
    PRETTY_NAME="FreeBSD 13.2-RELEASE"
    CPE_NAME="cpe:/o:freebsd:freebsd:13.2"
  Built: Sat Oct 7 2023 19:21:53 UTC
  Built With:
[...]

PR #1752, opened 2023-11-25
Issue #1746: If we fail to resolve the DNS name for the implicit "ser…

Castaglia:
…ver config" vhost to an IP address, we need to fail completely on startup.

Failure to do so will lead to unexpected null pointer dereferences elsewhere in modules.


Castaglia:
Ideally I would backport this to the 1.3.8 branch.  However, it changes the parsing behavior, making what is (in the 1.3.8 branch) a warning into a fatal error.

PR #1754, opened 2023-12-08
configure: Use char ** for the iconv input argument

fweimer-rh:
The standard `iconv` function uses `char **` even for the input buffer. Using the incompatible `const char **` type causes the check to fail if the compiler treats such type errors as errors, instead of merely warning about it.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/configure
+++ b/configure
@@ -21475,7 +21475,7 @@ main ()
 {
 
     size_t res, in_len = 0, out_len = 0;
-    const char *in = NULL;
+    char *in = NULL;
     char *out = NULL;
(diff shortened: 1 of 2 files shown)


Castaglia:
Could you provide details about the compiler, version, and flags being used, that encounter this issue?  I ask in order to check other compilers/versions on other platforms, for compatibility.  Thanks!

Issue #1756, opened 2023-12-17
Build system fails for specific module names


Castaglia:
While working on a new 3rd-party module to address Issue #1748 I stumbled across a subtle issue in ProFTPD's module build system.

_If_ you request a list of modules to be built, and the list of module names includes a short name that is actually a prefix of another, longer name, _and_ that longer name follows the prefix with `_o` (followed by other characters), then the `./configure` command will fail.

Here's a concrete example.  In my case, I am working on a new [`mod_sftp_openssh`](https://github.com/Castaglia/proftpd-mod_sftp_openssh) module.  So I went to build ProFTPD with this module, and with `mod_sftp`, using:
```
$ ./configure --with-modules=mod_sftp:mod_sftp_openssh
...
configure: error: source file './modules/penssh.c' cannot be found -- aborting
```
This is a very strange build error!

Castaglia:
This error happens only under rare, and very specific circumstances.  In the above case, it happens because my new module name, `mod_sftp_openssh`, has `mod_sftp` as a prefix, _and_ it follows that prefix with `_o`.

Seems odd, right?  The reason lies in the Autoconf script, where there are examples of using `sed`, with a variable, to remove items from lists, such as [here](https://github.com/proftpd/proftpd/blob/master/configure.in#L4213):
```
for module in $ac_static_modules ; do
  moduledir=`echo "$module" | sed -e 's/\.o$//'`;
  ...

    ac_static_modules=`echo "$ac_static_modules" | sed -e "s/$module//"`
```
In this case, the `$module` variable will be a value such as "mod_sftp.o", an object file.  The `$ac_static_modules` variable is a list of such object files, such as "mod_auth_pam.o mod_sftp.o mod_sftp_openssh.o".

So what happens when a string "mod_sftp.o" is interpolated into that `sed` command?
```
echo "mod_auth_pam.o mod_sftp.o mod_sftp_openssh.o" | sed -e "s/mod_sftp.o//"
```
The cause is the `.` in the left-hand value of the `sed` substitution: that `.` character is treated as a regular expression wildcard, _which matches any character_.  This means that the `sed` command will remove "mod_sftp.o" _and_ "mod_sftp_o" (from "mod_sftp_openssh.o"), leading to `$ac_static_modules` having a new value of "mod_auth_pam.o penssh.o".

This build system bug would arise for any module whose name uses another module name as a prefix, followed by `_o`.

Like I said, subtle.  And annoying.

PR #1757, opened 2023-12-17
Issue #1756: Be more careful/explicit when specifying the text to be …

Castaglia:
…removed, when handling lists of module names.

### DIFF
--- a/configure
+++ b/configure
@@ -24220,12 +24220,13 @@ ac_static_module_dirs=
 
 for amodule in $ac_shared_modules; do
   module=`echo "$amodule" | sed -e 's/\.la//g'`.o;
+  modulename=`echo "$amodule" | sed -e 's/\.la$//'`;
 
-  ac_core_modules=`echo "$ac_core_modules" | sed -e "s/$module//g"`;
-  ac_build_core_modules=`echo "$ac_build_core_modules" | sed -e "s/modules\/$module//g"`;
+  ac_core_modules=`echo "$ac_core_modules" | sed -e "s/$modulename\.o//g"`;
+  ac_build_core_modules=`echo "$ac_build_core_modules" | sed -e "s/modules\/$modulename\.o//g"`;
 
-  ac_static_modules=`echo "$ac_static_modules" | sed -e "s/$module//g"`;
-  ac_build_static_modules=`echo "$ac_build_static_modules" | sed -e "s/modules\/$module//g"`;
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -25685,33 +25685,39 @@ fi
 
 for module in $ac_shared_modules ; do
   moduledir=`echo "$module" | sed -e 's/\.la$//'`;
+  modulename=`echo "$module" | sed -e 's/\.o$//'`;
```
Castaglia, line 25688: Here we define a new variable, `$modulename`, which is the module without the `.o` suffix, as it would have from the source list.

configure
```diff
@@ -25685,33 +25685,39 @@ fi
...
+  src=`echo "$module" | sed -e 's/\.o$//'`.c;
+  srcinc=`echo "$module" | sed -e 's/\.o$//'`.h;
 
   if test -f $srcdir/modules/$src -o -f $srcdir/contrib/$src; then
     continue
 
   elif test -d $srcdir/modules/$moduledir; then
     ac_shared_module_dirs="$ac_shared_module_dirs modules/$moduledir";
-    ac_shared_modules=`echo "$ac_shared_modules" | sed -e "s/$module//"`
+    ac_shared_modules=`echo "$ac_shared_modules" | sed -e "s/$modulename\.o//"`
```
Castaglia, line 25697: We then use the new `$modulename` variable, _and explicitly add the `.o` suffix_ to be removed, with `.` character properly escaped.

configure
```diff
@@ -25745,8 +25751,8 @@ for module in $ac_static_modules ; do
         as_fn_error $? "specified archive '$thearch' does not match expected module archive name '$moduledir.a' -- aborting" "$LINENO" 5
       fi
 
-      ac_static_modules=`echo "$ac_static_modules" | sed -e "s/$module//g"`
-      ac_build_static_modules=`echo "$ac_build_static_modules" | sed -e "s/modules\/$module//g"`
+      ac_static_modules=`echo "$ac_static_modules" | sed -e "s/$modulename\.o//g"`
+      ac_build_static_modules=`echo "$ac_build_static_modules" | sed -e "s/modules\/$modulename\.o//g"`
```
Castaglia, line 25755: This build bug is very specifically located here.   Why?  It's the combination of the `.` wildcard character, _and_ the use of the `/g` modifier, which says to remove _all_ matching occurrences, not just the first matching occurrence.

While I could fix this issue simply by removing that `g` modifier, I feel that that would still leave the subtle `.` matching in place, which would arise to cause surprising issues later.

PR #1769, opened 2024-01-08
Fetch pubkey from certificate before extracting its parameters

manuelm:
This line has accidentally been removed in 028ce362dbfc1110cf0970483516f47f6732a9e5


Castaglia:
Thanks!

PR #1774, opened 2024-01-31
Add a second optional parameter to the SQLPasswordRounds option

Koloboc:
SQLPasswordRounds N [ PrependSalt | AppendSalt ]

The option allows you to add Salt in each iteration of "TRANSFORM" This type of password encryption is used on many websites (for example, CMS OpenCart). Fkr example
sha1(salt + sha1(salt + sha1(password)))

### DIFF
--- a/contrib/mod_sql_passwd.c
+++ b/contrib/mod_sql_passwd.c
@@ -93,6 +93,12 @@ static unsigned long sql_passwd_opts = 0UL;
 
 static unsigned long sql_passwd_nrounds = 1;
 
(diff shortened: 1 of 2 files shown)


Castaglia:
Would you be willing to update the documentation as well, for this new optional parameter, including providing examples of where it would be used?

maltris:
@Koloboc I find this to be a useful addition. If you briefly explain a use-case and example to be, I would like to take over the documentation part and add this.

Issue #1764, opened 2023-12-20
Using mod_quotatab hard limits in combination with mod_vroot causes inaccessible files


cfiehe:
### What I Did

I am using ProFTPD 1.3.8 as SFTP server in combination with mod_vroot. All users get change rooted to their home directory. When the user creates a directory and wants to access or delete the directory, access fails with `realpath /<folder-name>/: No such file`. This was working in ProFTPD 1.3.7f with mod-vroot  0.9.11. Because of Issue #1597, I had to migrate to 1.3.8.

Test case:
```
sftp> mkdir test-folder
sftp> cd test-folder/
realpath /test-folder/: No such file
sftp> ls
prod         test-folder
sftp> rm test-folder
Removing /test-folder
remote delete /test-folder: Failure
```

### What I Expected/Wanted

The user can access the formely created directory.

### ProFTPD Version and Configuration

Please help us reproduce the problem/issue you are encountering.  To do this,
we need to know which version of ProFTPD you are using, how it was built,
_etc_.  The following command is an easy way to get all of this information:
```bash
Compile-time Settings:
  Version: 1.3.8 (stable)
  Platform: LINUX [Linux 5.15.0-91-generic x86_64]
  OS/Release:
    PRETTY_NAME="Ubuntu 22.04.3 LTS"
    NAME="Ubuntu"
    VERSION_ID="22.04"
    VERSION="22.04.3 LTS (Jammy Jellyfish)"
    VERSION_CODENAME=jammy
    ID=ubuntu
    ID_LIKE=debian
    UBUNTU_CODENAME=jammy
  Built: Sun Jul 2 2023 10:56:40 UTC
  Built With:
[...]

PR #1778, opened 2024-02-11
Issue #1764: When mod_quotatab registers its own FS handlers for enfo…

Castaglia:
…rcing hard limits, it needs to be aware of other modules' FS handlers, such as mod_vroot.


contrib/mod_quotatab_file.c
```diff
@@ -41,9 +41,14 @@
 
 module quotatab_file_module;
 
+static const char *trace_channel = "quotatab.file";
```
Castaglia, line 44: Most of the changes here are for better traceability; no functional changes.

Issue #1803, opened 2024-05-06
SSH public key fingerprints in mod_sftp logs


offsides:
### What I Did

Logged in using SSH key authentication with mod_sftp

### What I Expected/Wanted

The public key SHA256 fingerprint is listed as '94:37:47:e7:77:82:c6:4a:d3:80:a8:66:09:2e:58:4b:39:40:22:63:b8:e7:e0:9f:20:94:1e:d8:e1:fa:c0:9e'.  When I use ssh-keygen to show the SHA256 fingerprint, it looks like this: 'SHA256:lDdH53eCxkrTgKhmCS5YSzlAImO45+CfIJQe2OH6wJ4'.  If I decode that base64 formatted fingerprint it looks like this: '00:b4:2d:87:eb:51:ee:8a:7a:3b:ce:57:e7:a9:3f:11:d5:ec:24:4c:c4:fe:2e:88:fb:b2:e3:67:39:19:d8:76'.  I would have expected that the 2 fingerprints that were formatted the same would be the same, but they aren't.  Ideally I'd like to see the SHA256 fingerprint in the same format that openssh uses, but that's not as critical.  But I do need to be able to match fingerprints and something is definitely not matching up.  If I need to do something different to calculate the hashes the way you are, I'm also open to that I just need to know what to do differently.

### ProFTPD Version and Configuration

Please help us reproduce the problem/issue you are encountering.  To do this,
we need to know which version of ProFTPD you are using, how it was built,
_etc_.  The following command is an easy way to get all of this information:
```bash
# proftpd -V
Compile-time Settings:
  Version: 1.3.9rc2 (devel)
  Platform: LINUX [Linux 5.14.0-362.24.1.el9_3.x86_64 x86_64]
  OS/Release:
    NAME="Red Hat Enterprise Linux"
    VERSION="9.3 (Plow)"
    ID="rhel"
    ID_LIKE="fedora"
    VERSION_ID="9.3"
    PLATFORM_ID="platform:el9"
    PRETTY_NAME="Red Hat Enterprise Linux 9.3 (Plow)"
    CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"

    REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
    REDHAT_BUGZILLA_PRODUCT_VERSION=9.3
    REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
[...]

PR #1804, opened 2024-05-08
Issue #1803: Only log the SSH key fingerprint when we also have the s…

Castaglia:
…igned data for verifying the key.


Castaglia:
Hmm.  In thinking about this, I do think there's value in logging the fingerprint of the offered key, even before the signed data is provided -- mostly for correlation/debugging.  So I'm going to close this PR.

PR #1811, opened 2024-06-11
Registering new event handler as head of the doubly linked list issue.

andreaszervas:
When registering multiple event handlers for the same event (for example in the case that an event should be handled differently by the various loaded modules), the registering of a new handler in the double linked list is not performed correctly.

With this modification the latest added event handler is positioned in the head of the doubly linked list.

### DIFF
--- a/src/event.c
+++ b/src/event.c
@@ -160,8 +160,8 @@ int pr_event_register(module *m, const char *event,
         }
 
         if (evh->module != NULL) {
-          if (evl->handlers->next != NULL) {
-            evl->handlers->next->prev = evh;
+          if (evl->handlers != NULL) {
+            evl->handlers->prev = evh;
           }
 
           evh->next = evl->handlers;


Castaglia:
Can you describe the specific context in which you encountered this, the unexpected behavior you observed, so that we can better understand how you ran into this?  Thanks!

andreaszervas:
The problem was encountered in the session context (in the child process created for the clients session).

2 custom modules are loaded, module_A and module_B with specified event handlers for "core.exit" event. 

In each of the handlers specific actions are supposed to be performed on exit. 

It was noticed that:
- when unregistering the event handler for module_A the remaining handlers of module_B and core were fine.
- when unregistering the event handler for module_B also the event handler of module_A was unregistered and only core event handler remained.
(This was the symptom)

Then the addresses of the prev and next member variables of the event handlers were printed out for debug purposes,
 and we encountered what you can see in the screenshot:
problem_event_reg
- handler of module_A: both prev and next are correct
- handler of module_B: prev is (nil) while it should be 0x20cffc0, next is correct
- handler of CORE: prev points to handler of module_A instead of module_B, next is correct

So the registering part was the problem.

Castaglia:
Thanks!  I've also cherry-picked this commit onto the 1.3.8 branch.

andreaszervas:
Thanks a lot!! 👍

PR #1812, opened 2024-06-24
fix: infinite loop when erroring on setting journal_mode

shieldwed:
In this instance of an SQLite busy detection loop, the loop is infinite in case the returned result (`res`) by SQLite is an error (!= `SQLITE_OK` and != `SQLITE_BUSY`). In this state, `res` always matches the `while` loop condition, but never matches the `if` condition. Additionally, the result (`res`) never changes, either.

This PR changes the loop condition to only enter, in case the SQLite returned result (`res`) is actually busy (`SQLITE_BUSY`).

However, this change also allows that the failed `PRAGMA journal_mode = MEMORY;` SQL statement is basically ignored, where it got stuck in the infinite loop before.

### DIFF
--- a/contrib/mod_sql_sqlite.c
+++ b/contrib/mod_sql_sqlite.c
@@ -377,26 +377,24 @@ MODRET sql_sqlite_open(cmd_rec *cmd) {
   /* Make sure we handle contention here, just like any other statement
    * (Issue#385).
    */
-  while (res != SQLITE_OK) {
-    if (res == SQLITE_BUSY) {
-      struct timeval tv;
+  while (res == SQLITE_BUSY) {
+    struct timeval tv;
 
-      nretries++;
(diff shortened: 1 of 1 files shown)


shieldwed:
Is there more error handling required for the case that `PRAGMA journal_mode = MEMORY;` failed? After all it was specifically introduced (85daefd23868b1b4797f87ef1509765db09272bb) to allow for chrooted instances of ProFTPD.

Castaglia:
> Is there more error handling required for the case that `PRAGMA journal_mode = MEMORY;` failed? After all, it was specifically introduced ([85daefd](https://github.com/proftpd/proftpd/commit/85daefd23868b1b4797f87ef1509765db09272bb)) to allow for chrooted instances of ProFTPD.

Good catch!  I'm not sure if you ran into this infinite loop; if so, to answer your question, I think we'd have to see just what the reported error was.  So let's get this merged, and then go from there.  Sound good?

PR #1816, opened 2024-07-11
Don't include <openssl/engine.h> if ENGINE support not detected

pghmcfc:
OpenSSL ENGINE support is deprecated in Fedora 41 onwards: https://fedoraproject.org/wiki/Changes/OpensslDeprecateEngine

As part of this change, OpenSSL is still built with ENGINE support in order to maintain compatibility with existing applications, but the <openssl/engine.h> header is moved to a different package so that users have to make an explicit choice to support it.

Since proftpd already supports building without ENGINE support, we can qualify inclusion of <openssl/engine.h> using the results of the configure check. If there's no ENGINE support, there's no point including it.

It would be nice to include a similar change in mod_proxy. I can do an additional PR for that if it's helpful, though it seems a fairly trivial change.

### DIFF
--- a/contrib/mod_sftp/mod_sftp.h.in
+++ b/contrib/mod_sftp/mod_sftp.h.in
@@ -112,7 +112,9 @@
 #include <openssl/rsa.h>
 #if OPENSSL_VERSION_NUMBER > 0x000907000L
 # include <openssl/aes.h>
-# include <openssl/engine.h>
+# ifdef PR_USE_OPENSSL_ENGINE
+#  include <openssl/engine.h>
+# endif /* PR_USE_OPENSSL_ENGINE */
 # include <openssl/ocsp.h>
 #endif
 #if defined(PR_USE_OPENSSL_ECC)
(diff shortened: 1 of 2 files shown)


Castaglia:
Bleh.  Looks like I'll need to sort out the fallout to our GitHub workflows, caused by:
* https://github.com/aws-actions/configure-aws-credentials/issues/1106
* https://github.com/actions/checkout/issues/1809

Castaglia:
Thanks!  I'll merge this now, and deal with the CI fallout manually as needed.

Castaglia:
Here's the accompanying mod_proxy change: https://github.com/Castaglia/proftpd-mod_proxy/commit/1c6d8543d2013788a9d4122978ccec139a6161e4

Castaglia:
And I've backported this to the 1.3.8 branch, too: https://github.com/proftpd/proftpd/commit/9812830dea511f1b26beaa73536ce4826e167c7b

PR #1817, opened 2024-07-11
Address some compiler warnings relating to null pointer dereferences

pghmcfc:
This addresses warnings like this:
```
mod_tls.c: In function ‘tls_tlsext_cb’:
mod_tls.c:4975:61: warning: ‘%.*s’ directive argument is null [-Wformat-overflow=]
 4975 |         "[tls.tlsext] TLS %s extension \"%s\" (ID %d, %d %s)%.*s",
      |                                                             ^~~~
```
The warning is because ext_info (returned from BIO_get_mem_data) can be NULL.

This may not be your preferred method of fixing this warning but it does silence the compiler regarding this issue.

There are similar issues in mod_proxy too.


Castaglia:
Nice.  This will do quite nicely -- much appreciated!

Here're the equivalent changes done in mod_proxy: https://github.com/Castaglia/proftpd-mod_proxy/commit/884a57c68218cc8b495766f9f0e34e51f4c07dd0

Issue #1813, opened 2024-06-27
How to exclude mod_sftp READ and WRITE commands from the log


offsides:
### What I Did

My mod_sftp server section has looging configured as this:
```
ExtendedLog /var/log/proftpd/proftpd.log ALL,!SSH customs
```

### What I Expected/Wanted

I want to exclude the READ and WRITE commands from the log.  but READ and WRITE are also classes, which means excluding those keywords excludes the entire class.  How do I just exclude the specific commands rather than the class?

### ProFTPD Version and Configuration

Proftpd 1.3.9rc2 + patches.

Castaglia:
Is this something that can be done via post-processing of the generated logs?

Alternatively, you might try configuring the full list of classes to include (and omit the classes/commands you don't want), in your `ExtendedLog` filtering.

offsides:
(I thought I'd submitted this a couple weeks ago... oops)

Except that I really do want to log all commands except for the SFTP READ/WRITE, which for large files generates huge amounts of data.  I suppose I can just do all classes except READ & WRITE, and then include the individual commands in those classes, but it seems to me that having both a **class** and a **command** with the same name is probably not the best idea.  Maybe the solution is to change the names of READ/WRITE commands in mod_sftp to something like READFILE/WRITEFILE to differentiate them better.

Castaglia:
The `READ`, `WRITE` class names predate the advent of mod_sftp; changing them creates a risk of backward incompatibility/breaking config changes, which I'd prefer not to do if possible.

offsides:
That's why I suggested changing the names of the commands in mod_sftp rather than changing the class names.  I agree that changing the class names is a bad idea, but the naming collision should be resolved somehow.

Castaglia:
Hmm.  In some ways, this request reminds me of Issue #1822 as well -- the desire to do filtering of log messages.

Ideally such site-specific log processing is done by some separate tool/process, outside of ProFTPD.

If this ideal is untenable, it'd be good to know why, as that would inform just how ProFTPD code might be changed (and why).  For example, this request _could_ be addressed by implementing some configuration directive that provides a regular expression that acts as a filter, on each and every log message emitted to the configured `ExtendedLog`, to accept/reject that particular line.  Such a filter would preserve existing compatibility/formats, and possibly allow the desired "don't include these log lines" -- at the cost of increased CPU usage.

offsides:
[...]

PR #1821, opened 2024-07-18
Change SFTP READ/WRITE commands to READFILE/WRITEFILE to remove the naming collision for log filtering purposes

offsides:
Change the SFTP READ and WRITE commands to READFILE and WRITEFILE so that they no longer clash with the command classes READ and WRITE (per https://github.com/proftpd/proftpd/issues/1813). This allows SFTP read/write operations to be filtered out of logs without filtering out the entire class. This also avoids breaking anything with regards to class filtering, as READFILE remains part of the READ class and WRITEFILE remains part of the write class, so that however logs were configured previously will continue to work exactly the same, it just allows these 2 specific commands to be explicitly excluded (or included) in log statements that do not already include/exclude the READ and WRITE classes.

There are several other instances of "READ" and "WRITE" in fxp.c, but it's unclear to me if they all need to be changed to "READFILE"/"WRITEFILE" or not. But if they do, that's easy enough to do, just let me know.

### DIFF
--- a/contrib/mod_sftp/fxp.c
+++ b/contrib/mod_sftp/fxp.c
@@ -1097,10 +1097,10 @@ static const char *fxp_get_request_type_desc(unsigned char request_type) {
       return "CLOSE";
 
     case SFTP_SSH2_FXP_READ:
-      return "READ";
+      return "READFILE";
 
     case SFTP_SSH2_FXP_WRITE:
-      return "WRITE";
+      return "WRITEFILE";
 
     case SFTP_SSH2_FXP_LSTAT:
       return "LSTAT";
@@ -4314,7 +4314,7 @@ static int fxp_handle_ext_check_file(struct fxp_packet *fxp, char *digest_list,
   }
 
   cmd = fxp_cmd_alloc(fxp->pool, "SITE_DIGEST", pstrdup(fxp->pool, path));
-  if (!dir_check(fxp->pool, cmd, "READ", path, NULL)) {
+  if (!dir_check(fxp->pool, cmd, "READFILE", path, NULL)) {
     xerrno = EACCES;
 
     (void) pr_log_writefile(sftp_logfd, MOD_SFTP_VERSION,
@@ -4721,7 +4721,7 @@ static int fxp_handle_ext_copy_file(struct fxp_packet *fxp, char *src,
 
     cmd2 = pr_cmd_alloc(fxp->pool, 3, "SITE_COPY", src, dst);
     cmd2->arg = pstrdup(fxp->pool, args);
-    limit_allow = dir_check(fxp->pool, cmd2, "WRITE", dst, NULL);
+    limit_allow = dir_check(fxp->pool, cmd2, "WRITEFILE", dst, NULL);
 
     if (!overwrite ||
         (allow_overwrite == NULL ||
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/fxp.c
```diff
@@ -4314,7 +4314,7 @@ static int fxp_handle_ext_check_file(struct fxp_packet *fxp, char *digest_list,
   }
 
   cmd = fxp_cmd_alloc(fxp->pool, "SITE_DIGEST", pstrdup(fxp->pool, path));
-  if (!dir_check(fxp->pool, cmd, "READ", path, NULL)) {
+  if (!dir_check(fxp->pool, cmd, "READFILE", path, NULL)) {
```
Castaglia, line 4317: This means that `<Limit READ>` configurations, for SFTP sessions, would no longer match/work as expected.

Castaglia:
Hmm.  While I understand the desire here, I'm not sure that the backward compatibility issues such a change would cause warrant this.  I'm wondering if there's a different way to achieve the functionality desired in Issue #1813.

contrib/mod_sftp/fxp.c
```diff
@@ -1097,10 +1097,10 @@ static const char *fxp_get_request_type_desc(unsigned char request_type) {
       return "CLOSE";
 
     case SFTP_SSH2_FXP_READ:
-      return "READ";
+      return "READFILE";
```
Castaglia, line 1100: This will change the names as logged, which in turn will have consequences for any log parsers/regexes that folks may be relying upon.

contrib/mod_sftp/fxp.c
offsides: Hmm...  OK I can see that.  But this feels like it's one of those things that should have been done differently from the beginning since there was already a class named READ and WRITE, and there may not be a way to resolve the issue without a breaking change.  The one thing I can think of would be to either have a compile-time directive of whether to use READ or READFILE (and WRITE or WRITEFILE), or maybe even a runtime directive, which would allow configurable backward compatibility until people were ready to make the switch.  And it would give you time to officially deprecate the READ/WRITE command names and ultimately remove them in a future version.  Because as it stands right now, you cannot control the READ/WRITE commands individually in anything that also accepts command classes, because using them there affects the entire class rather than just the command...

Actually, I just realized that the documentation on Limit explicitly refers to the classes and not the commands:
```
    What about the READ and WRITE command groups?

    <Limit READ>
    Applies to the SYMLINK and READ SFTP requests

    <Limit WRITE>
    Applies to the LOCK, SETSTAT, SYMLINK, and WRITE SFTP requests
```
So based on the current documentation, this change would not actually break anything, because the SFTP READ(FILE) command would still be affected by a limit on the READ class, and the same for WRITE(FILE) and the WRITE class.

Your argument below about logfile parsers is definitely more impactful, though based on what the READ and WRITE commands currently log I'm not sure how useful they are for parsing vs the rest of the logs, and depending on what they are looking for if all they look for is READ and WRITE they might still match since only SFTP uses those commands in the logs.  I don't know what the right way to make a transition would be, but it feels pretty solid that the commands READ and WRITE need to somehow be differentiated from the classes READ and WRITE, and since the classes came first (and encompass the relevant commands), its the commands that should be renamed and not the classes.

offsides:
Any further thoughts on this?

offsides:
superseded by #1862.

Issue #1288, opened 2021-08-06
extremly slow upload speed, using mod_sftp


bdimitrovITGix:
### What I Did

Hello,

We are have a strange issue regarding upload speeds. Our proFTPD server is installed on a CentOS7 machine and is configured to use LDAP authentication and mod_sftp. It is running on a custom port- 2222.
Downloading is as expected, around 10MB/s when testing with 1GB file.
Uploading is no where near that, running at 300-500kb/s when testing with the same 1GB file.

At the same time, trying the standart SFTP on port 22 doe not have this problem. Upload and Download there is fast, as expected. This should mean that the issue is not in our Network.
No transfer limits have been set and enforcing such did not improve the UPLOAD speed.

I have tried it using FileZilla and Linux CLI as client connections.
### What I Expected/Wanted
I have been banging my head for days now. Any help towards troubleshooting this will be much appreciated.

### ProFTPD Version and Configuration

**VERSION:**

Compile-time Settings:
  Version: 1.3.5e (maint)
  Platform: LINUX [Linux 3.10.0-862.14.4.el7.x86_64 x86_64]
  Built: Wed Jan 31 2018 16:32:29 UTC
  Built With:
[...]

PR #1844, opened 2024-10-30
Issue #1288: Optimize handling of SFTP packets, and thus hopefully up…

Castaglia:
…loads, by avoiding more unnecessary `memmove(2)` calls.

### DIFF
--- a/contrib/mod_sftp/fxp.c
+++ b/contrib/mod_sftp/fxp.c
@@ -3231,16 +3231,20 @@ static void fxp_packet_set_packet(struct fxp_packet *pkt) {
   curr_pkt = pkt;
 }
 
-static void fxp_packet_clear_cache(void) {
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/fxp.c
```diff
@@ -14010,7 +14015,7 @@ static int fxp_send_display_login_file(uint32_t channel_id) {
 int sftp_fxp_handle_packet(pool *p, void *ssh2, uint32_t channel_id,
     unsigned char *data, uint32_t datalen) {
   struct fxp_packet *fxp;
-  int have_cache, res;
+  int res;
```
Castaglia, line 14018: Rather than having this `have_cache` variable here, reset to "false" whenever we handle a new packet, we need to keep that state elsewhere, so that it persists properly _across_ multiple SFTP packets.

PR #1862, opened 2025-01-03
Add compile-time option to change the name of SFTP READ and WRITE commands by appending a suffix to them

offsides:
To resolve the issue where there is a namespace collision between the mod_sftp READ and WRITE commands and the proftpd READ and WRITE classes, I propose that a suffix (default of 'FILE') be appended to the SFTP READ and WRITE commands.  However, to avoid it being a mandatory breaking change, this patch adds a configure directive, --with_sftp_readwrite_cmdsuffix to allow the builder the option of changing the name of the mod_sftp commands.  This directive takes an optional argument of the suffix to use, or will use the suffix of 'FILE' if no argument is provided changing the command names to READFILE and WRITEFILE.  Since these commands are still in the READ and WRITE classes, any limit or log directive that use the READ/WRITE classes will still apply to them, but they will now also be able to have those directives applied to just the mod_sftp read and write operations.  This allows, e.g., not logging SFTP read and write records to the log file by excluding just those specific lines rather than the entire class of commands.  If this directive is not given, the command names default to READ and WRITE, maintaining full compatibility with the current implementation.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -1110,6 +1110,12 @@
 /* Define if largefile support is desired.  */
 #undef PR_USE_LARGEFILES
 
+/* Command name to use for SFTP reads. */
+#undef PR_SFTP_READCMD
+
+/* Command name to use for SFTP writes. */
+#undef PR_SFTP_WRITECMD
+
 /* Define if Memcache support is desired.. */
 #undef PR_USE_MEMCACHE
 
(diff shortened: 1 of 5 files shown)


offsides:
I just realized that this won't actually work, as the log filtering only accepts command classes and not command names.  So I'm cancelling this PR and will look at alternate approaches.

PR #1863, opened 2025-01-08
Add command class CL_SFTPRW to allow SFTP READ & WRITE commands to be filtered

offsides:
In addition to the common (FTP & SFTP) RETR/STOR command, SFTP logs READ/WRITE commands for every chunk of data sent/received.  This can create large amounts of less-than-helpful log lines making it more difficult to quickly scan the log file.  This adds a class CL_SFTPRW that can be used to filter out the SFTP READ/WRITE commands in ExtendedLog config lines (e.g., ALL,!SFTPRW) without impacting other logging, and since it doesn't remove the commands from their existing classes it makes no change to existing proftpd logging configurations.  I believe that this is both a minimalist solution and avoids any backwards incompatibility.

### DIFF
--- a/contrib/mod_sftp/fxp.c
+++ b/contrib/mod_sftp/fxp.c
@@ -10373,7 +10373,7 @@ static int fxp_handle_read(struct fxp_packet *fxp) {
 #endif
 
   cmd = fxp_cmd_alloc(fxp->pool, "READ", name);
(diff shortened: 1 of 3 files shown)


Castaglia:
I'm going to close this, in favor of https://github.com/proftpd/proftpd/pull/1919, now that it has been merged.  Thanks!

PR #1870, opened 2025-02-21
Normal parsing causes config directives to be appended to the end of …

Castaglia:
…their container set.  <Directory> configs are prepended, though.

To preserve both sets of behaviors better, mod_ifsession needs to use the appropriate merge flags for the different directive types, rather than always prepending, as it does now.

See https://github.com/proftpd/proftpd/commit/29507be38c0598313b158540e65b77227fae07dd for where the current behavior originated, and why (_i.e._ too much development focus on <Directory> sections without keeping other configs in mind).

### DIFF
--- a/contrib/mod_ifsession.c
+++ b/contrib/mod_ifsession.c
@@ -1,7 +1,7 @@
 /*
  * ProFTPD: mod_ifsession -- a module supporting conditional
  *                            per-user/group/class configuration contexts.
- * Copyright (c) 2002-2023 TJ Saunders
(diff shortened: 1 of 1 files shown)


Castaglia:
Should this behavior be configurable?  Consider that, at present, mod_ifsession would merge a config directive via prepending to the set.  And that means that that directive would be found _first_, before the previously-existing directives.  Which might be the expected/desired behavior of such mod_ifsession-managed configurations.

To better decide, I'll need to do some testing with mod_ifsession using order-sensitive directives, such as those from mod_rewrite.

Issue #1876, opened 2025-03-04
SFTP channel allocations can lead to high memory utilization over time


andreaszervas:
Hello @Castaglia! 

There is a memory utilization issue discovered in the case that, within a single sftp session, a client repeatedly performs:
- channel open, 
- perform some operations,
- channel close

In order for this memory issue to become noticeable this opening-closing operation needs to be performed thousands of times.

The version used for testing was 1.3.8c, but this does not look to be a regression.

After looking into the code it looks like this is related to the fact that the pool allocated for the sftp channel (alloc_channel function of channel.c from mod_sftp) is not deallocated in channel close.

By adding the destruction of the channel sub pool in the destroy_channel function of channel.c like:

destroy_pool(chans[i]->pool);

the problem seems to be resolved.

Limited testing has been done, and I would really appreciate your help on this one.

Thanks a lot,

Andreas

Castaglia:
Good find!  We do indeed need to destroy the channel-specific pool as well.   In addition, the particular use case you highlight:
* channel open,
* perform some operations,
* channel close

done thousands of times, for a single session/connection, can lead to other (less severe) means of allocated-and-not-freed memory.  Let's see if I can provide a PR to address both of these issues.

Castaglia:
Could you provide logs (SFTPLog, TraceLog, debug logging) for your segfault?  I'm not seeing that locally, so I'm wondering what might be different.

andreaszervas:
Hi @Castaglia just tested with the latest additions to the branch and the SIGSEGV is fixed!

Also memory utilization is now stable in my test of performing continuous opening-closing of the channel within the same sftp session. 

Thanks a lot!!

Castaglia:
These changes have been merged to master, and backported to the 1.3.8 branch.  Thanks!

andreaszervas:
Thanks a lot!!

PR #1877, opened 2025-03-04
Issue #1876: Ensure we free up the channel-specific memory pool when …

Castaglia:
…destroy channels, and also implement an upper bounds on the memory used for tracking all channels across the lifetime of the connection.

### DIFF
--- a/contrib/mod_sftp/channel.c
+++ b/contrib/mod_sftp/channel.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD - mod_sftp channels
- * Copyright (c) 2008-2024 TJ Saunders
+ * Copyright (c) 2008-2025 TJ Saunders
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/channel.c
```diff
@@ -121,7 +123,26 @@ static struct ssh2_channel *alloc_channel(const char *type,
     channel_list = make_array(channel_pool, 1, sizeof(struct ssh2_channel *));
   }
 
-  *((struct ssh2_channel **) push_array(channel_list)) = chan;
+  /* Look for an empty slot in the list, from an already-destroyed channel,
+   * first.
+   */
+  chans = channel_list->elts;
+  for (i = 0; i < channel_list->nelts; i++) {
+    if (chans[i] == NULL) {
```
Castaglia, line 131: By reusing an existing (and empty) slot in the channel list, we avoid expanding the list on every new channel.  The [`SFTPMaxChannels`](http://www.proftpd.org/docs/contrib/mod_sftp.html#SFTPMaxChannels) limit places an upper bound on the number of concurrent items in this list, so as long as we reusing empty slots, there is no excess memory allocation across thousands of opened/closed channels.

Issue #1886, opened 2025-03-17
SCP transfers fail for files with spaces in their names


silentbogo:
Hello. 
After upgrading to version 1.3.8.b I've started encountering strange errors, and some of my files from the same directory with the same owner/rights would or would not download. After a brief look - all files that failed had at least one space in the file name. 
I'm using the latest total commander with SFTP plugin. The error only occurs when I have SCP enabled for file transfer. If I disable it(use SFTP for transfer) - no more problems with spaces, no more errors. 
Haven't had this issue in the past on older versions of proftpd. I'm using the same config as before, only removed some older ciphers from the list.
My co-worker tested on WinSCP - and everything works fine.

Castaglia:
What was the previous of ProFTPD that you had been using, prior to upgrading to 1.3.8b, so that I might start comparing/examining the list of changes between the versions?

silentbogo:
I believe it was 1.3.5e. We were migrating from Ubuntu 18.04 to 24.04.
Config is very minimal, no ClientMatch used, no global directory settings (just .ftpaccess for individual folders):
`<IfModule mod_sftp.c>
<VirtualHost 0.0.0.0>
  SFTPEngine ON
  Port ****
  SFTPHostKey /etc/proftpd/keys/ssh_host_rsa_key
  ...
  SFTPHostKey /etc/proftpd/keys/ssh_host_ecdsa_key
  SFTPCiphers aes128-cbc blowfish-cbc cast128-cbc arcfour128 \
              arcfour256 aes192-cbc aes256-cbc aes128-ctr aes192-ctr aes256-ctr \
              aes128-gcm@openssh.com aes256-gcm@openssh.com
  SFTPLog /var/log/proftpd/sftp.log
  SFTPCompression off
  SFTPAuthMethods password
  SFTPClientAlive 10 30
  AuthUserFile /etc/proftpd/ftpd.passwd
  AuthGroupFile /etc/proftpd/ftpd.group
  AuthOrder mod_auth_file.c
  SFTPPAMEngine off
  DefaultRoot ~
  RequireValidShell off
  DirFakeUser on ~
  DirFakeGroup on ~
  SFTPOptions IgnoreSCPUploadTimes IgnoreSFTPSetTimes
  DelayEngine off
</VirtualHost>
</IfModule>`

Castaglia:
Thanks, that info helps.

[...]

PR #1887, opened 2025-03-18
While working on Issue #1886 on a newer Mac, I encountered issues due…

Castaglia:
… to lack of up-to-date support for the Mac's `getopt(3)` implementations.


Castaglia:
The lack of this update mostly causes issues for SCP transfers on Mac OSX servers, hence why it was interfering with my ability to re-create the reported behaviors.  One yak shaved; let's see how many remain.

PR #1900, opened 2025-03-27
Bug #4512: Fix the compilation of the `mod_wrap2` module when the `--…

Castaglia:
…enable-wrapper-options` configure option is used.

### DIFF
--- a/contrib/mod_wrap2/mod_wrap2.c
+++ b/contrib/mod_wrap2/mod_wrap2.c
@@ -23,7 +23,7 @@
 
 #include "mod_wrap2.h"
 
-#ifdef WRAP2_USE_NIS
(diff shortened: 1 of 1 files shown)


contrib/mod_wrap2/mod_wrap2.c
```diff
@@ -1075,7 +1076,7 @@ static int wrap2_handle_opts(array_header *options, wrap2_conn_t *conn) {
     /* Separate the option into name and value parts. For backwards
      * compatibility we ignore exactly one '=' between name and value.
      */
-    curropt = wrap2_opt_trim_string(curr_opt);
+    curr_opt = wrap2_opt_trim_string(curr_opt);
```
Castaglia, line 1079: This is the actual change that fixes the build.

Issue #1896, opened 2025-03-24
Uploads using `MODE Z` sometimes result in corrupted files or broken transfers


R4to0:
Hello!

I've noticed that some files I try to upload to my server are failing when mod_deflate is enabled on the server.

Not all of my files triggers this issue, from the set I'm testing only one doesn't trigger from a set of six targa (.tga) files. Each of them are 196652 bytes in size, but during transfer they're inflated to different file sizes:

```
2025-03-24 18:10:43,208 mod_deflate/0.6[1027962]: STOR: inflated 8976 bytes to 189353 bytes (95.26% compression)
2025-03-24 18:10:44,221 mod_deflate/0.6[1027962]: STOR: inflated 9047 bytes to 190232 bytes (95.24% compression)
2025-03-24 18:10:45,248 mod_deflate/0.6[1027962]: STOR: inflated 9351 bytes to 189300 bytes (95.06% compression)
2025-03-24 18:10:46,707 mod_deflate/0.6[1027962]: STOR: inflated 66293 bytes to 196626 bytes (66.28% compression)
2025-03-24 18:10:47,687 mod_deflate/0.6[1027962]: STOR: inflated 8520 bytes to 190728 bytes (95.53% compression)
2025-03-24 18:10:48,249 mod_deflate/0.6[1028097]: wrote 2034 (of 2034) bytes of compressed of data to socket 17
2025-03-24 18:10:48,262 mod_deflate/0.6[1028097]: wrote 1553 (of 1553) bytes of compressed of data to socket 17
2025-03-24 18:10:48,274 mod_deflate/0.6[1028097]: wrote 1504 (of 1504) bytes of compressed of data to socket 17
2025-03-24 18:10:48,288 mod_deflate/0.6[1028097]: wrote 1463 (of 1463) bytes of compressed of data to socket 17
2025-03-24 18:10:48,302 mod_deflate/0.6[1028097]: wrote 1491 (of 1491) bytes of compressed of data to socket 17
2025-03-24 18:10:48,317 mod_deflate/0.6[1028097]: wrote 1531 (of 1531) bytes of compressed of data to socket 17
2025-03-24 18:10:48,331 mod_deflate/0.6[1028097]: wrote 1438 (of 1438) bytes of compressed of data to socket 17
2025-03-24 18:10:48,346 mod_deflate/0.6[1028097]: wrote 1527 (of 1527) bytes of compressed of data to socket 17
[...]

PR #1902, opened 2025-03-27
Issue #1896: Fix an edge case in the `mod_deflate` module's handling …

Castaglia:
…of compressed uploads, where it was improperly overwriting an internal buffer, which leads to corrupted files and/or broken transfers in some cases.

### DIFF
--- a/contrib/mod_deflate.c
+++ b/contrib/mod_deflate.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD: mod_deflate -- a module for supporting on-the-fly compression
- * Copyright (c) 2004-2024 TJ Saunders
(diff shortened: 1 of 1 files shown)


contrib/mod_deflate.c
```diff
@@ -513,30 +508,6 @@ static int deflate_netio_read_cb(pr_netio_stream_t *nstrm, char *buf,
     switch (deflate_zerrno) {
       case Z_OK:
       case Z_STREAM_END:
-        copylen = deflate_zbufsz - zstrm->avail_out;
```
Castaglia, line 516: With proper handling of the input and output buffers, we no longer need to handle the edge case of dynamically detecting (and wrongly in this case) of when we need to increase the size of the output buffer.

Issue #1904, opened 2025-03-28
mod_sftp fails to parse authorized user/host public keys with CRLF line endings


Castaglia:
A mod_sftp user recently reported privately that an RSA publickey authentication was failing unexpectedly.  The ensuing discussion led us to more closely examine the actual publickey text that had been configured in the server via the `SFTPAuthorizedUserKeys` directive:
```
$ od -c authorized_keys 
0000000    -   -   -   -       B   E   G   I   N       S   S   H   2    
0000020    P   U   B   L   I   C       K   E   Y       -   -   -   -  \r
0000040   \n   C   o   m   m   e   n   t   :       "   2   0   4   8   -
```
Notice how, for this file, the expected start-of-key marker text, `---- BEGIN SSH2 PUBLIC KEY ----` is followed by a `\r` (CR, Carriage Return) character, then an `\n` (LF, Line Feed) character, to indicate end-of-line.  Such CRLF line ending sequences are more commonly seen in Windows environments; Unix text files traditionally use just LF as the line ending character.  But SSH public keys generated on Windows platforms can legitimately end up using CRLF sequences, and thus mod_sftp should handle them properly.

In fact, there _is_ existing code in mod_sftp that checks for such CRLF line ending text in the authorized keys file; see:
* https://github.com/proftpd/proftpd/blob/master/contrib/mod_sftp/rfc4716.c#L137

The problem is that that function checks for _either_ CR _or_ LF _first_, earlier in the function:
* https://github.com/proftpd/proftpd/blob/master/contrib/mod_sftp/rfc4716.c#L85

This means that for the above public key, mod_sftp detects that the last character is the LF, trims it off -- but leaves the CR there.  Later, when mod_sftp goes to see if this line of text matches the epxected start-of-key marker, it fails to match -- because of that lingering CR character.  This, in turn, causes the rest of the public key to be ignored, and the user authentication using the publickey method will unexpectedly fail.

Castaglia:
Interestingly, this particular issue, of CRLF first, then CR or LF, does _not_ happen in the `mod_sftp_sql` module's code for parsing such keys; see:
* https://github.com/proftpd/proftpd/blob/master/contrib/mod_sftp_sql.c#L116

gjaekel:
@Castaglia: Thank you for your immediate assistance in investigating this issue and providing a patch to trace it down to the root. The problem was solved within hours!

Castaglia:
[...]

PR #1905, opened 2025-03-28
Issue #1904: When parsing SSH authorized key files, make sure we chec…

Castaglia:
…k for CRLF line endings first, and CR-or-LF line endings next.


gjaekel:
This will fix the bug for the first sprint. But to my eyes, there is to much duplicated code inside the `then` and `else` part of the `if` at https://github.com/proftpd/proftpd/blob/bc0cdda2a10b25923e667c4fdf7d625c9990ac6f/contrib/mod_sftp/rfc4716.c#L91 that might be moved below.

Issue #164, opened 2015-10-12
Improve integration with systemd when running ProFTPD in standalone mode


zabbal:
Right now proftpd supports both standalone and inetd modes. Having the latter it should be rather easy to add systemd's socket activation support so it can be nicely integrated with modern distributions.

Castaglia:
FYI, this is also being tracked in the project's Bugzilla as [Bug#3661](http://bugs.proftpd.org/show_bug.cgi?id=3661).

Castaglia:
I am contemplating adding this functionality via a module, _e.g._ a `mod_systemd` module that can listen for the relevant internal ProFTPD events, and communicate/notify systemd as needed.

Given the OpenSSH/xzutils/libsystemd fiasco of last year:
* https://en.wikipedia.org/wiki/XZ_Utils_backdoor

I am not inclined to use libsystemd for this work, but instead to do what OpenSSH did, which is to provide their own systemd notification mechanism/implementation:
* https://github.com/openssh/openssh-portable/commit/08f579231cd38a1c657aaa6ddeb8ab57a1fd4f5c

@pghmcfc @hpreusse do you see requests/desires from your users for this systemd "socket activation" support in ProFTPD?  I'm curious.

pghmcfc:
@Castaglia nobody has ever asked me for this feature, but it does seem to be a nice-to-have and might even be worth making the default mode of operation.

Castaglia:
I hope that having proper systemd support can help to address questions that have arisen in the past, such as:
* https://github.com/proftpd/proftpd/issues/1651#issuecomment-1493053469
* https://github.com/proftpd/proftpd/issues/1666#issuecomment-1537165310

Castaglia:
Support for systemd's `Type=notify-reload` is now in master, enabled by building the new `mod_systemd` module.

Depending on uptake/feedback on this module, I may have it be built by default by the `configure` script.

PR #1908, opened 2025-04-05
Issue #164: Implement support for systemd socket activation via a new…

Castaglia:
… `mod_systemd` module.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -2,7 +2,7 @@
  * ProFTPD - FTP server daemon
  * Copyright (c) 1997, 1998 Public Flood Software
  * Copyright (c) 1999, 2000 MacGyver aka Habeeb J. Dihu <macgyver@tos.net>
- * Copyright (c) 2001-2024 The ProFTPD Project team
+ * Copyright (c) 2001-2025 The ProFTPD Project team
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -1043,7 +1043,7 @@ void restart_daemon(void *d1, void *d2, void *d3, void *d4) {
   init_config();
   init_dirtree();
 
-#ifdef PR_USE_NLS
+#if defined(PR_USE_NLS)
   encode_free();
 #endif /* PR_USE_NLS */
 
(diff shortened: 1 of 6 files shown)


src/main.c
```diff
@@ -1093,6 +1093,7 @@ void restart_daemon(void *d1, void *d2, void *d3, void *d4) {
   }
 
   pr_event_generate("core.postparse", NULL);
+  pr_event_generate("core.restarted", NULL);
```
Castaglia, line 1096: This new `core.restarted` event is the main reason why this module requires ProFTPD 1.3.10rc1, and not earlier versions (as they did not generate this event).  This event is necessary for notifying systemd of when ProFTPD is reloading, per the requirements of the `Type=notify-reload` directive in the unit file.

pghmcfc:
Just tried this with 1.3.10rc1 and noticed that the sample proftpd.service file now has both "Type=notify-reload" and "ExecReload=...". The documentation at https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html suggests that "implementing this reload protocol is an efficient alternative to providing an ExecReload= command for reloading of the service's configuration", so I tried deleting the "ExecReload=" line but that just resulted in the reload timing out. I notice in the system log that proftpd did receive a SIGHUP but never completed the reload to systemd's satisfaction. Reinstating the ExecReload= resulted in working reloads. I don't understand why the ExecReload= is still necessary here.

Castaglia:
@pghmcfc Excellent question.  I'm not sure, not being an systemd expert myself.  As this module is a work-in-progress, I was expecting there to be some wrinkles.  I'll try to take a look into what might be happening here as soon as I can.  Thanks for the feedback/comments, they'll help me polish this module into working/expected order.

Castaglia:
@pghmcfc Hmm.  Interestingly, later in that same document, when discussing `ExecReload`, we see:
> This setting can be combined with Type=notify-reload, in which case the service main process is signaled after all specified command lines finish execution. Specially, if "RELOADING=1" notification is received before ExecReload= completes, the signaling is skipped and the service manager immediately starts listening for "READY=1".

It would indeed be nice to know though why, if this "optional" `ExecReload=` is omitted from the `.service` file, why `mod_systemd` wouldn't complete the reloading to systemd's requirements.

Castaglia:
For anyone curious, this issue/interaction with `ExecReload` was addressed by https://github.com/proftpd/proftpd/pull/2061

PR #1915, opened 2025-04-08
Issue #1288: Optimize SFTP downloads a little more by removing an unn…

Castaglia:
…ecessary `memcpy(2)` call.


contrib/mod_sftp/channel.c
```diff
@@ -1760,7 +1760,7 @@ static int channel_write_data(pool *p, uint32_t channel_id,
 
       /* In addition to the data itself, we need to allocate room in the
        * outgoing packet for the type (1 byte), the channel ID (4 bytes),
-       * a possible data type ID (4 bytes),  and for the data length (4 bytes).
+       * a possible data type ID (4 bytes), and for the data length (4 bytes).
```
Castaglia, line 1763: This is just a textual nit that was bothering me, removing an unnecessary space character.  No functional change.

Issue #1916, opened 2025-04-08
Make the SSH key fingerprint logged by mod_sftp configurable


Castaglia:
As demonstrated by Issue #1803 some admins would like mod_sftp to log SSH key fingerprints in a manner more similar to what OpenSSH does; see _e.g._
* https://github.com/proftpd/proftpd/issues/1803#issuecomment-2269126124

OpenSSH itself has an [`FingerprintHash`](https://man.openbsd.org/sshd_config#FingerprintHash) directive in its `sshd_config` for just this sort of feature.

The scope of this ticket, then, is to implement similar support in mod_sftp, such as an `SFTPKeyFingerprints` directive which might take the encoding (backward-compatible hex, or OpenSSH-compatible baes64) as well as the digest algorithm (MD5, SHA256).  Thus for logging key fingerprints similar to what newer OpenSSH does, one might use:
```
SFTPKeyFingerprints sha256+base64
```
For now, to retain compatibility with what mod_sftp does now, the default will probably be:
```
SFTPKeyFingerprints sha256+hex
```
One might argue that, since we're at the start of new 1.3.10 release cycle, now might be the best time to change the default to indeed be OpenSSH-compatible, and sites that wish mod_sftp to use its previous behavior would need to explicitly configure it to be so.

FYI @cshug @offsides

cschug:
@Castaglia Excellent, thanks a lot for your work! I was able to apply the diff on the 1.3.9 release as well and it seems good to me, SHA256 fingerprint are now logged in the same format as `ssh-keygen` does.

PR #1917, opened 2025-04-09
Issue #1916: Implement an `SFTPKeyFingerprints` directive, to support…

Castaglia:
… configuring the specific of the hashing and encoding used for logging public key fingerprints during authentication.

### DIFF
--- a/contrib/mod_sftp/keys.c
+++ b/contrib/mod_sftp/keys.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD - mod_sftp key mgmt (keys)
- * Copyright (c) 2008-2024 TJ Saunders
+ * Copyright (c) 2008-2025 TJ Saunders
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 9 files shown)


contrib/mod_sftp/keys.c
```diff
@@ -2606,21 +2606,107 @@ const char *sftp_keys_get_fingerprint(pool *p, unsigned char *key_data,
...
+    *algo = c->argv[1];
+    *fmt_id = *((int *) c->argv[2]);
+
+  } else {
+    /* Use default key fingerprint algorithms/encodings. */
+    *fmt_id = SFTP_KEYS_FP_FMT_BASE64;
+
+#if defined(HAVE_SHA256_OPENSSL)
+    *algo_id = SFTP_KEYS_FP_DIGEST_SHA256;
+    *algo = "SHA256";
```
Castaglia, line 2688: Here is where we set the defaults.  Note that, unlike previous versions, we now default to using SHA256 and base64 (rather than colon-delimited hexadecimal), which makes this compatible with newer `ssh-keygen -l -f` outputs.

PR #1919, opened 2025-04-09
Issue #1813: Modify the ExtendedLog directive syntax to allow "siftin…

Castaglia:
…g", i.e. more control over logged commands that allow command-specific inclusion/exclusion of the commands filtered by class.

### DIFF
--- a/src/jot.c
+++ b/src/jot.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD - FTP server daemon
- * Copyright (c) 2017-2024 The ProFTPD Project team
+ * Copyright (c) 2017-2025 The ProFTPD Project team
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -33,6 +33,9 @@ struct jot_filters_rec {
   int included_classes;
   int excluded_classes;
   array_header *cmd_ids;
+
+  /* Additional, refining sifting rules. */
+  array_header *cmd_sifts;
 };
 
 /* For tracking the size of deleted files. */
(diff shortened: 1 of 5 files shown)


src/jot.c
```diff
@@ -2205,22 +2208,86 @@ static int is_jottable_cmd(cmd_rec *cmd, int *cmd_ids, size_t ncmd_ids) {
...
+    int negated = FALSE;
+
+    cmd_name = cmd_names[i];
+
+    if (cmd_name[0] == '!') {
+      negated = TRUE;
+      cmd_name++;
+    }
+
+    if (strcmp(cmd_name, (const char *) cmd->argv[0]) == 0) {
```
Castaglia, line 2230: With this implementation, we do a direct comparison of the configured commands in the "sifting" list against the command being logged/processed.  This way, the approach applies to any/all commands/requests names across protocols, and Just Works(tm) for any future names as well.

offsides:
How exactly does this work?  I'd like to play with it and see if it does what I need rather than the custom patch I currently am using.  From how you describe it, conceptually it sounds like a good approach.

Castaglia:
> How exactly does this work? I'd like to play with it and see if it does what I need rather than the custom patch I currently am using. From how you describe it, conceptually it sounds like a good approach.

See _e.g._ https://github.com/proftpd/proftpd/issues/1813#issuecomment-2791014379

offsides:
OK, so a hypothetical log line with READ+!READ would log everything in the READ class except for the actual READ command?   I can't fault your methodology of making it generic so that it can be used to filter pretty much everything, even if that particular construct is a little clunky.  I'll grab the patch and give it a whirl and let you know how it goes.  Thanks!

offsides:
I just tried it and after thinking it wasn't working I realized I was looking at the wrong log file, and looking at the correct file shows that it works exactly as I want it to.  Thank you very much!  Now to figure out all the other patches you've merged over the past month or so... :)

PR #1925, opened 2025-04-16
The core API for doing RFC 1413-style lookup was moved into the `mod_…

Castaglia:
…ident` module, and removing from the core build, years ago.


src/ident.c
```diff
@@ -1,236 +0,0 @@
...
-
-  /* Set up our timer before going any further. */
-  timerno = pr_timer_add(PR_TUNABLE_TIMEOUTIDENT, -1, NULL,
-    (callback_t) ident_timeout_cb, "ident lookup");
-  if (timerno <= 0) {
-    destroy_pool(tmp_pool);
-    return pstrdup(p, ret);
-  }
-
-  ident_conn = pr_inet_create_connection(tmp_pool, NULL, -1, c->local_addr,
```
Castaglia, line 74: I noticed this old file as I was reworking some of the Inet API's connection functions; a `grep` turned up this particular reference to a `pr_inet_create_connection` function, which no longer exists.  And that, in turn, indicates that this file is never compiled.

PR #1926, opened 2025-04-16
Bug #3802: While looking into occurrences of `EADDRINUSE` when attemp…

Castaglia:
…ting to create local sockets for active data transfers, I did some refactoring to reduce the verbosity (and confusion) of the messages logged in such cases.

### DIFF
--- a/include/inet.h
+++ b/include/inet.h
@@ -2,7 +2,7 @@
  * ProFTPD - FTP server daemon
  * Copyright (c) 1997, 1998 Public Flood Software
  * Copyright (c) 1999, 2000 MacGyver aka Habeeb J. Dihu <macgyver@tos.net>
- * Copyright (c) 2001-2022 The ProFTPD Project team
+ * Copyright (c) 2001-2025 The ProFTPD Project team
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


Castaglia:
Note that as part of this investigation, I did verify that yes, in fact, the `data_active_open()` function _does_ retry failed `bind(2)` calls when `EADDRINUSE` occurs (as when the local ephemeral port range is exhausted by many other data transfers).  This happens already; I don't feel the need to increase the number of retry attempts, as posted/requested in [Bug #3802](http://bugs.proftpd.org/show_bug.cgi?id=3802).

include/inet.h
```diff
@@ -139,8 +139,17 @@ int pr_inet_reverse_dns(pool *, int);
...
 conn_t *pr_inet_copy_conn(pool *, conn_t *);
-conn_t *pr_inet_create_conn(pool *, int, const pr_netaddr_t *, int, int);
-conn_t *pr_inet_create_conn_portrange(pool *, const pr_netaddr_t *, int, int);
+
+conn_t *pr_inet_create_conn(pool *p, int fd, const pr_netaddr_t *bind_addr,
+  int port, int retry_bind);
+conn_t *pr_inet_create_conn2(pool *p, int fd, const pr_netaddr_t *bind_addr,
+  int port, int flags);
+#define PR_INET_CREATE_CONN_FL_RETRY_BIND	0x001
+#define PR_INET_CREATE_CONN_FL_LOG_ERRORS	0x002
```
Castaglia, line 148: With these new flags, particularly the `LOG_ERRORS` flag, callers can explicitly say whether they want the verbose (confusing?) messages logged, or not.

src/inet.c
```diff
@@ -509,9 +512,10 @@ static conn_t *init_conn(pool *p, int fd, const pr_netaddr_t *bind_addr,
         pr_signals_unblock();
       }
 
-      if (reporting) {
-        pr_log_pri(PR_LOG_ERR, "Failed binding to %s, port %d: %s",
-          pr_netaddr_get_ipstr(&na), port, strerror(hold_errno));
+      pr_log_pri(PR_LOG_ERR, "Failed binding to %s, port %d: %s",
+        pr_netaddr_get_ipstr(&na), port, strerror(hold_errno));
```
Castaglia, line 516: This message is now logged all of the time (as was happening most of the time previously); the messages about checking `ServerType` are only relevant for the daemon listening sockets, not for session-specific data transfer sockets.

src/inet.c
```diff
@@ -649,25 +665,18 @@ conn_t *pr_inet_create_conn_portrange(pool *p, const pr_netaddr_t *bind_addr,
...
 
-      c = init_conn(p, -1, bind_addr, ports[i], FALSE, FALSE);
-      if (c == NULL &&
-        /* Note that on Solaris, bind(2) might fail with EACCES if the
-         * randomly selected port for e.g. passive transfers is used by
-         * NFS.  Thus, for Solaris only, we treat EACCES as the same as
-         * EADDRINUSE.  Silly Solaris.
-         */
-#ifdef SOLARIS2
-          (inet_errno != EADDRINUSE && inet_errno != EACCES)) {
```
Castaglia, line 660: This block of comment+code already happens in `init_conn()` anyway, so I feel it should be removed.  As should the abrupt "kill this session if there's any other error" behavior.  That's unnecessarily rude to clients.

PR #1928, opened 2025-04-17
Bug #4037: Adding regression tests that reproduce the provided config…

Castaglia:
…uration; the test behavior shows uploads as expected via `.ftpaccess` configurations.


tests/tests.pl
```diff
@@ -205,7 +205,9 @@
     t/config/ftpaccess/empty.t
     t/config/ftpaccess/merging.t
     t/config/ftpaccess/retr.t
+    t/config/ftpaccess/stor.t
     t/config/limit/anonymous.t
+    t/config/limit/list.t
```
Castaglia, line 210: I forgot to add this test from a previous commit, so this is a catch-up change.

Issue #1931, opened 2025-04-18
0-byte file uploads due to TLS 1.3 Session Ticket expiry


ValentinSteiner:
I've observed an issue where some file uploads result in 0-byte files on the server (when using TLS 1.3). The problem occurs when uploading multiple files, and only after several hours, at a transition point between two consecutive file transfers.

I'm able to reproduce this consistently on ProFTPD 1.3.9. The client I've experienced this with is FileZilla.

Here's what I've found out.

* TLS 1.3 uses session tickets with a specific lifetime (maximum age), set by the server.
* Clients are expected to use the same TLS context between the control and the data connection(s).
* Since the control connection is first, clients are supposed to reuse the control connection's TLS context for all data connections that are established during the life of the control connection.
* That means that at the beginning of each file transfer, there is a chance that the session ticket has expired.
* Assuming the ticket has expired, the client now opens a data connection, either with the expired session ticket, or without a session ticket at all (FileZilla). Either way, a new TLS handshake occurs, the TLS session is not resumed.
* ProFTPD rejects the connection, but writes a 0-byte file.
* FileZilla prints: "Error:  TLS session resumption on data connection failed. Closing control connection to start over."
* Closing the data and control connection is a total reset, resulting in a clean sheet, and new TLS context.
* Assuming default settings, instead of overwriting, FileZilla will ask what to do with the 0-byte file that now exists on the server.

I believe that both could behave better, client and server. I've already filed a [bug against FileZilla](https://trac.filezilla-project.org/ticket/13235). When I wrote that report, I assumed that FileZilla sends the expired session ticket in its ClientHello. I've since found out that it does not, but the result is the same. The data connection won't reuse the TLS context.

ProFTPD uses a default ticket lifetime of 15840 seconds (4h 24min). In my real-life scenario, this is when the failures happened.

### Config

```
ServerType          standalone
Port                21
DefaultServer       on

User                proftpd
Group               nogroup

TimeoutNoTransfer   600
TimeoutStalled      20
TimeoutIdle         1200

<IfModule mod_tls.c>
  TLSEngine                        on
  TLSLog                           /var/log/proftpd/tls.log
  TLSProtocol                      TLSv1.3
  TLSRequired                      on
  TLSRSACertificateFile            /etc/proftpd/cert/cert.pem
  TLSRSACertificateKeyFile         /etc/proftpd/cert/key.pem

  # Short session ticket lifetime to demonstrate
  TLSSessionCache                  internal: 15
[...]

PR #1932, opened 2025-04-18
Issue #1931: Explicitly request new TLSv1.3 session tickets prior to …

Castaglia:
…FTPS uploads in order to avoid handling expired tickets that cause the uploads to be aborted.

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -6802,7 +6802,7 @@ static int tls_ticket_key_cb(SSL *ssl, unsigned char *key_name,
    * to their ciphers, digests?
    */
   const EVP_CIPHER *cipher = EVP_aes_256_cbc();
-# ifdef OPENSSL_NO_SHA256
+# if defined(OPENSSL_NO_SHA256)
   const EVP_MD *md = EVP_sha1();
 # else
   const EVP_MD *md = EVP_sha256();
@@ -6922,14 +6922,14 @@ static int tls_ticket_key_cb(SSL *ssl, unsigned char *key_name,
         "ticket renewal", key_name_str, (unsigned long) key_age,
         key_age != 1 ? "secs" : "sec", (unsigned long) newest_age,
         newest_age != 1 ? "secs" : "sec");
-      return 2;
     }
 
 # if defined(TLS1_3_VERSION)
     /* If we're a TLSv1.3 session, aim for single-use tickets, and indicate
-     * that the client should get a new ticket.  Will the FTPS client
-     * Do The Right Thing(tm), and use this new ticket for future data
-     * transfers?
+     * that the client should get a new ticket -- but only for control
+     * connection sessions, not data transfer (see Issue #1931).  Will the
+     * FTPS client Do The Right Thing(tm), and use this new ticket for future
+     * data transfers?
      */
     if (SSL_version(ssl) == TLS1_3_VERSION) {
       return 2;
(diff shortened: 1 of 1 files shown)


contrib/mod_tls.c
```diff
@@ -12916,20 +12917,18 @@ MODRET tls_any(cmd_rec *cmd) {
...
-        tls_log("SSL/TLS required but absent on control channel, "
-          "denying %s command", (char *) cmd->argv[0]);
-        pr_response_add_err(R_550,
-          _("SSL/TLS required on the control channel"));
+    if (tls_authenticated != NULL &&
+        *tls_authenticated == TRUE) {
+      tls_log("SSL/TLS required but absent on control channel, "
+        "denying %s command", (char *) cmd->argv[0]);
+      pr_response_add_err(R_550,
+        _("SSL/TLS required on the control channel"));
```
Castaglia, line 12927: Minor syntax changes; no functional changes.

contrib/mod_tls.c
```diff
@@ -13005,6 +13004,40 @@ MODRET tls_any(cmd_rec *cmd) {
...
+  ssl_session = SSL_get_session(ctrl_ssl);
+  ssl_version = SSL_SESSION_get_protocol_version(ssl_session);
+
+  if (ssl_version != TLS1_3_VERSION) {
+    return PR_DECLINED(cmd);
+  }
+
+  pr_trace_msg(trace_channel, 19,
+    "requesting new TLSv1.3 session ticket before upload");
+  SSL_new_session_ticket(ctrl_ssl);
```
Castaglia, line 13035: This required `SSL_new_session_ticket` API only appeared in OpenSSL 3.x, which means that OpenSSL-1.x sites which encounter this issue won't have a good solution, just the workaround of increasing the ticket/cache lifetime.

contrib/mod_tls.c
```diff
@@ -13005,6 +13004,49 @@ MODRET tls_any(cmd_rec *cmd) {
...
+  ssl_version = SSL_SESSION_get_protocol_version(ssl_session);
+
+  if (ssl_version != TLS1_3_VERSION) {
+    return PR_DECLINED(cmd);
+  }
+
+# if defined(OPENSSL_VERSION_MAJOR) && \
+     defined(OPENSSL_VERSION_MINOR) && \
+     OPENSSL_VERSION_MAJOR >= 3
+  if (SSL_new_session_ticket(ctrl_ssl) == 1) {
```
Castaglia, line 13036: The required `SSL_new_session_ticket` API only appeared in OpenSSL 3.x; sites using OpenSSL 1.x that encounter this issue will unfortunately need to resort to the workaround of configuring a longer ticket lifetime.

contrib/mod_tls.c
```diff
@@ -13005,6 +13024,78 @@ MODRET tls_any(cmd_rec *cmd) {
...
+   * before proceeding with the data transfer (Issue #1931).
+   */
+  if (sess_remaining <= 10) {
+    request_new_ticket = TRUE;
+  }
+
+# if defined(OPENSSL_VERSION_MAJOR) && \
+     OPENSSL_VERSION_MAJOR >= 3
+  if (request_new_ticket == TRUE) {
+    if (SSL_new_session_ticket(ctrl_ssl) == 1) {
```
Castaglia, line 13083: I will experiment with using [`SSL_key_update`](https://docs.openssl.org/master/man3/SSL_key_update/) instead here, as is done [elsewhere](https://github.com/proftpd/proftpd/blob/master/contrib/mod_tls.c#L4140), to see if that also fixes things.   Why?  `SSL_key_update` first appeared in OpenSSL 1.1.x, meaning that this fix would work for older OpenSSL installations as well.

contrib/mod_tls.c
```diff
@@ -13005,6 +13002,81 @@ MODRET tls_any(cmd_rec *cmd) {
...
+  if (sess_remaining <= 10) {
+    request_new_ticket = TRUE;
+  }
+
+  if (request_new_ticket == TRUE) {
+    int res;
+
+# if defined(OPENSSL_VERSION_MAJOR) && \
+     OPENSSL_VERSION_MAJOR >= 3
+    res = SSL_new_session_ticket(ctrl_ssl);
```
Castaglia, line 13057: I recently discovered that there is another solution possible, when using OpenSSL-1.x _or_ LibreSSL: we can extend the lifetime of the current TLS session object, such that it will be valid (not expired) when the subsequent data transfer TLS handshake occurs, such that `SSL_session_reused` will return true.

So this PR is not quite done yet.

However, that still leaves some edge cases to be explored: OpenSSL-3.x and non-TLSv1.3 sessions, with and without session tickets.

Castaglia:
I'm going to merge this PR as-is, to at least make forward progress on this issue for TLSv1.3, OpenSSL-3.x sites.  I'll keep the associated issue open for a bit longer, as I work on other combinations/scenarios.

Issue #831, opened 2019-09-14
UserAlias outside of existing <Anonymous> sections are handled as anonymous logins


deAtog:
### What I Did

Upgraded from ProFTPD 1.3.5 to 1.3.6. Configuration contains a UserAlias directive in the Global section. Login is successful in 1.3.5 but is handled as anonymous in 1.3.6.

Another user reported the issue here 11 months ago:

https://serverfault.com/questions/932083/proftpd-useralias-not-working-on-centos-7-possibly-pam-related

### What I Expected/Wanted

Login via UserAlias works as expected.

### ProFTPD Version and Configuration
```
Compile-time Settings:
  Version: 1.3.6 (stable)
  Platform: LINUX [Linux 3.10.0-957.5.1.el7.x86_64 x86_64]
  Built: Tue Jul 23 2019 18:20:14 UTC
  Built With:
[...]

PR #1935, opened 2025-04-21
Issue #831: Handle `UserAlias` directives outside of `<Anonymous>` se…

Castaglia:
…ctions properly.

### DIFF
--- a/modules/mod_auth.c
+++ b/modules/mod_auth.c
@@ -2,7 +2,7 @@
  * ProFTPD - FTP server daemon
  * Copyright (c) 1997, 1998 Public Flood Software
  * Copyright (c) 1999, 2000 MacGyver aka Habeeb J. Dihu <macgyver@tos.net>
- * Copyright (c) 2001-2024 The ProFTPD Project team
+ * Copyright (c) 2001-2025 The ProFTPD Project team
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


modules/mod_auth.c
```diff
@@ -4113,15 +4112,7 @@ MODRET set_useralias(cmd_rec *cmd) {
...
-  c = add_config_param_str(cmd->argv[0], 2, alias, real_user);
-
-  /* Note: only merge this directive down if it is not appearing in an
-   * <Anonymous> context.
-   */
-  if (!check_context(cmd, CONF_ANON)) {
-    c->flags |= CF_MERGEDOWN_MULTI;
-  }
-
+  add_config_param_str(cmd->argv[0], 2, alias, real_user);
```
Castaglia, line 4115: The "fix" here is removing the use of the `CF_MERGEDOWN_MULTI` flags on this config setting.

Why does this fix things?  When `CF_MERGEDOWN_MULTI` is used, it means that this `UserAlias` config_rec will be merged down into _e.g._ `<Anonymous>` sections, which may have their own separate `UserAlias` config_recs.  That's OK.  However, these merged-down config recs are found in `pr_auth_get_anon_config` [here](https://github.com/proftpd/proftpd/blob/master/src/auth.c#L1570) -- and _then_ treated as `<Anonymous>` logins [here](https://github.com/proftpd/proftpd/blob/master/src/auth.c#L1636) _because_ these non-anonymous `UserAlias` config_recs were merged down into the `<Anonymous>` section.  Oops.

So then I went digging, to find out why the `UserAlias` config_recs had been merged down in the first place.  Was it really necessary in the past for some reason?  I came across https://github.com/proftpd/proftpd/commit/1385d5393de87447e56291b5735044ace1f8a8d0#diff-6764960d8c9e5a7960e04fe3b41e5a4936802b4b79fc39918f28a2df8bb84298 So no, not really needed, just overdone work, when I was still very new to the ProFTPD code base.

PR #1940, opened 2025-05-15
Using `clang-tidy` to start polishing the code base, and using mod_tl…

Castaglia:
…s as the canary.

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -529,7 +529,7 @@ typedef struct tls_pkey_obj {
   int dsa_passlen;
   void *dsa_pkey_ptr;
 
-#ifdef PR_USE_OPENSSL_ECC
+#if defined(PR_USE_OPENSSL_ECC)
   char *ec_pkey;
   int ec_passlen;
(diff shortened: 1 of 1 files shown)


contrib/mod_tls.c
```diff
@@ -758,76 +758,77 @@ static X509_STORE *tls_crl_store = NULL;
 static array_header *tls_tmp_dhs = NULL;
 static RSA *tls_tmp_rsa = NULL;
 
-static void tls_exit_ev(const void *, void *);
-static int tls_sess_init(void);
```
Castaglia, line 762: The declaration for `tls_sess_init` also exists already on line 797; we don't need it here as well.

contrib/mod_tls.c
```diff
@@ -758,76 +758,77 @@ static X509_STORE *tls_crl_store = NULL;
...
+static void tls_exit_ev(const void *event_data, void *user_data);
```
Castaglia, line 761: Start naming the arguments, even in function declarations, for legibility.

PR #1944, opened 2025-05-19
Address stylistic nits found by `clang-tiday` in the mod_rewrite modu…

Castaglia:
…le; no functional changes/regressions, per the tests.

### DIFF
--- a/contrib/mod_rewrite.c
+++ b/contrib/mod_rewrite.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD: mod_rewrite -- a module for rewriting FTP commands
- * Copyright (c) 2001-2024 TJ Saunders
+ * Copyright (c) 2001-2025 TJ Saunders
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


contrib/mod_rewrite.c
```diff
@@ -539,12 +564,11 @@ static unsigned char rewrite_match_cond(cmd_rec *cmd, config_rec *cond) {
       rewrite_log("rewrite_match_cond(): checked lexical LT cond: %s > %s: %d",
         cond_str, (char *) cond->argv[1], res);
 
-      if (!negated) {
-        return (res < 0 ? TRUE : FALSE);
-
-      } else {
+      if (negated == TRUE) {
```
Castaglia, line 567: The human brain has a much easier time processing positives _vs_ negatives, thus I rewrote these conditionals to use the positive case first, and the negative case as the `else` branch.

contrib/mod_rewrite.c
```diff
@@ -1895,6 +1931,10 @@ static int rewrite_open_fifo(config_rec *c) {
     return -1;
   }
 
+  if (pr_fs_get_usable_fd2(&fd) < 0) {
+    return -1;
+  }
```
Castaglia, line 1936: Without this, the fd for a FIFO script is usually 2 (_i.e._ the same as stderr).  Not ideal.  And adding this bit o' code made the accompanying regression test pass.

PR #1954, opened 2025-06-03
use $(LIBTOOL) to install proftpd

orbea:
With slibtool `make install` will install the slibtool wrapper script instead of the actual executable file. This is because with slibtool the executable is compiled into the `.libs` directory while the file in the root directory is a wrapper script where with GNU libtool the executable will be placed into the root directory. Using `$(LIBTOOL)` to install the executable will ensure that both implementations install the correct file.

Gentoo-Issue: https://bugs.gentoo.org/953968


orbea:
On second look I think only `proftpd` itself needs this since other files aren't compiled with `$(LIBTOOL)`.

Castaglia:
Looks OK to me.  Thanks!

orbea:
> For my edification, "slibtool" here refers to https://github.com/midipix-project/slibtool

The github repo is a mirror, the upstream is here. https://dev.midipix.org/cross/slibtool

There is also a Gentoo wiki page that has some more info. https://wiki.gentoo.org/wiki/Slibtool

PR #1955, opened 2025-06-03
don't include attr/xattr.h

orbea:
There is no reason to include `attr/xattr.h` when `sys/xattr.h` is included.

This silences a very verbose warning.
```
In file included from ../include/fsio.h:41,
                 from ../include/conf.h:105,
                 from ftpdctl.c:27:
/usr/include/attr/xattr.h:5:2: warning: #warning "Please change your <attr/xattr.h> includes to <sys/xattr.h>" [-Wcpp]
    5 | #warning "Please change your <attr/xattr.h> includes to <sys/xattr.h>"
      |  ^~~~~~~
```

### DIFF
--- a/include/fsio.h
+++ b/include/fsio.h
@@ -37,10 +37,9 @@
 #  include <sys/extattr.h>
 # elif defined(HAVE_SYS_XATTR_H)
 #  include <sys/xattr.h>
-#  if defined(HAVE_ATTR_XATTR_H)
+# elif defined(HAVE_ATTR_XATTR_H)
 #   include <attr/xattr.h>
-#  endif /* HAVE_ATTR_XATTR_H */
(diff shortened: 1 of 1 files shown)


Castaglia:
Hmm.  I wonder how OS/distro-specific this is; I'll poke around on some of the BSDs, and other distros, to see if they also display this warning.

Castaglia:
Interesting.  FreeBSD (at least) uses `sys/extattr.h`, so it wouldn't be impacted by this change.  And in some versions of the `attr-dev` package I'm seeing on various Alpine images, there is `sys/xattr.h`, but no `attr/xattr.h` -- so those installations wouldn't be necessarily impacted.  So, in general, I think this change is OK -- we'll deal with any potential fallout in subsequent RCs.  Thanks!

PR #1956, opened 2025-06-13
Issue #1288: Experiment with disabling the Nagle algorith for SSH con…

Castaglia:
…nections.

### DIFF
--- a/contrib/mod_sftp/mod_sftp.c
+++ b/contrib/mod_sftp/mod_sftp.c
@@ -262,6 +262,12 @@ static void sftp_cmd_loop(server_rec *s, conn_t *conn) {
   sftp_conn = conn;
   pr_session_set_protocol("ssh2");
 
+  /* Disable the Nagle algorithm for SSH connections. */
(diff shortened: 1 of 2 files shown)


contrib/mod_sftp/mod_sftp.c
```diff
@@ -262,6 +262,12 @@ static void sftp_cmd_loop(server_rec *s, conn_t *conn) {
   sftp_conn = conn;
   pr_session_set_protocol("ssh2");
 
+  /* Disable the Nagle algorithm for SSH connections. */
+  if (pr_inet_set_proto_nodelay(sftp_pool, conn, 1) < 0) {
+    pr_trace_msg(trace_channel, 1,
+      "error enabling TCP_NODELAY for SSH session: %s", strerror(errno));
+  }
```
Castaglia, line 269: Hrm.  This may not actually be changing anything (thus not necessary); looks like we _already_ disable Nagle by default at https://github.com/proftpd/proftpd/blob/master/src/main.c#L1391

Issue #1960, opened 2025-06-27
Handling of CRLs in mod_tls is incorrect, leading to confusing errors


nicoden81:
**Summary**
When using mod_tls with a PEM-formatted CRL, ProFTPD fails to validate the CRL signature and logs:

invalid signature on CRL:
```
  (1) error:02000077:rsa routines::wrong signature length
  (2) error:1C880004:Provider routines::RSA lib
  (3) error:06880006:asn1 encoding routines::EVP lib
```
However, the same CRL is successfully verified using `openssl verify` with `-crl_check`.

**Environment**
* ProFTPD version: 1.3.8 (install from debian package)
* OpenSSL version: OpenSSL 3.0.x (system default)
* OS: Debian 12 (bookworm)
* TLS module used: mod_tls (with  mod_proxy_protocol, etc.)

**Steps to reproduce**

Enable CRL support in ProFTPD:
```
TLSCACertificateFile /etc/proftpd/ssl/certs/CA.crt.pem
TLSCARevocationFile  /etc/proftpd/ssl/certs/CA.crl.pem
TLSVerifyClient on
```
Provide a CRL signed by the same CA used for client certificates.

Use a certificate listed in the CRL.

Observe the error in tls.log:
```
 invalid signature on CRL: wrong signature length
```
**What works**

Running OpenSSL directly:
```
openssl verify \
  -CAfile /etc/proftpd/ssl/certs/CA.crt.pem \
  -CRLfile /etc/proftpd/ssl/certs/CA.crl.pem \
  -crl_check /path/to/client-cert.pem
```
returns:
```
error 23 at 0 depth lookup: certificate revoked
error /path/to/client-cert.pem: verification failed
```
This means the CRL is correct and the signature is valid.

**Expected behavior**
mod_tls should reject the client certificate without reporting a signature error on the CRL if it is valid and correctly signed.

**Additional info**
* The CRL is in proper PEM format (`-----BEGIN X509 CRL-----`).
* The issue persists even with a fresh CRL.
* The `TLSLog` output is identical every time.

Let me know if you need a redacted copy of the CRL and certificate for testing.

Thanks!

Castaglia:
This looks vaguely familiar; I may have encountered similar errors in the past, when working with CRLs on other projects.

[...]

PR #1968, opened 2025-07-14
Issue #1960: Fix the handling of CRLs in mod_tls.

Castaglia:


contrib/mod_tls.c
```diff
@@ -754,7 +754,6 @@ static size_t tls_data_ticket_appdatasz = 0, tls_data_ticket_appdata_len = 0;
 /* OpenSSL variables */
 static SSL *ctrl_ssl = NULL;
 static SSL_CTX *ssl_ctx = NULL;
-static X509_STORE *tls_crl_store = NULL;
```
Castaglia, line 757: We no longer need a separate `X509_STORE` object to hold CRLs.  A given `X509_STORE` can hold _both_ trusted CAs _and_ CRLs.  So rather than using a separate object, we will load our CRLs into the _existing_ `X509_STORE` used, in the `SSL_CTX`, for the `TLSCACertificateFile/Path` CAs.

contrib/mod_tls.c
```diff
@@ -10332,29 +10325,6 @@ static int tls_verify_cb(int ok, X509_STORE_CTX *ctx) {
...
-          crl_verify_err = X509_STORE_CTX_get_error(ctx);
-#else
-          crl_verify_err = ctx->error;
-#endif /* OpenSSL-1.1.x and later */
-
-         /* If we use the wrong public key to verify the CRL (as for an empty
-          * CRL, or a CRL signed by a different CA/key than the offered cert),
-          * then this could fail in an expected manner (Bug #4468).
-          */
-          if (crl_verify_err == X509_V_ERR_CRL_SIGNATURE_FAILURE) {
```
Castaglia, line 10348: This `X509_V_ERR_CRL_SIGNATURE_FAILURE` would occur because of the overly-complex previous implementation, where we could (and would) use the wrong public key to verify the signature on a CRL.  Since that implementation has been redone, this particular error/code path should not occur going forward.

contrib/mod_tls.c
```diff
@@ -10428,38 +10381,20 @@ static int tls_verify_cb(int ok, X509_STORE_CTX *ctx) {
...
         break;
 
-      case X509_V_ERR_INVALID_PURPOSE: {
-        register int i;
-        int count;
-
-        tls_log("client certificate failed verification: %s",
-          X509_verify_cert_error_string(ctx_error));
-
-        count = X509_PURPOSE_get_count();
```
Castaglia, line 10451: This particular scenario, of `X509_V_ERR_INVALID_PURPOSE`, can happen for multiple reasons.  (The one I encountered was from using `nsComment` in my OpenSSL config file, from a legacy configuration; I suspect that using other such legacy Netscape, or other, extensions could result in the same error, for no reason.)

In this particular case, logging the _possible_ purpose count (and what they are) is not useful, for such logging does _not_ indicate which particular purpose (or, most likely, _extension_) in the certificate caused OpenSSL such indigestion.  So I remove this logging entirely.

contrib/mod_tls.c
```diff
@@ -10472,109 +10407,79 @@ static int tls_verify_cb(int ok, X509_STORE_CTX *ctx) {
...
-   */
-  store_ctx = X509_STORE_CTX_new();
-#if OPENSSL_VERSION_NUMBER > 0x000907000L
-  if (X509_STORE_CTX_init(store_ctx, tls_crl_store, NULL, NULL) <= 0) {
-    tls_log("error initializing CRL store context: %s", tls_get_errors());
-    X509_STORE_CTX_free(store_ctx);
-    return ok;
-  }
-#else
-  X509_STORE_CTX_init(store_ctx, tls_crl_store, NULL, NULL);
```
Castaglia, line 10554: Now that we load the configured CRLs into the `SSL_CTX`, we no longer need a separate `X509_STORE_CTX`; we just use the one provided in the callback.  And that means we no longer need the complex implementation from the previous comments.

Less is more.

contrib/mod_tls.c
```diff
@@ -18340,22 +18210,21 @@ static int tls_ctx_set_ciphers(SSL_CTX *ctx) {
...
+  if (store == NULL) {
+    tls_log("error getting SSL_CTX store: %s", tls_get_errors());
     return -1;
   }
 
   PRIVS_ROOT
-  if (X509_STORE_load_locations(tls_crl_store, tls_crl_file,
-      tls_crl_path) != 1) {
-
+  if (X509_STORE_load_locations(store, tls_crl_file, tls_crl_path) != 1) {
```
Castaglia, line 18227: Here, we load the configured CRLs into the existing `SSL_CTX`.  Much simpler (and better!) this way.

PR #1979, opened 2025-09-13
fix C23, functions unprototyped

PPN-SD:
>./pr_fnmatch_loop.c:65:1: error:
>a function definition without a prototype is deprecated in all versions of C and
>      is not supported in C23 [-Werror,-Wdeprecated-non-prototype]
>   65 | FCT (pattern, string, string_end, no_leading_period, flags, ends)
>      | ^
>pr_fnmatch.c:241:14: note: expanded from macro 'FCT'
>  241 | # define FCT    internal_fnmatch
>In file included from mod_core.c:29:
>pr_fnmatch.c:358:1: error:
>a function definition without a prototype is deprecated in all versions of C and is not
>      supported in C23 [-Werror,-Wdeprecated-non-prototype]

>memcache.c:330:12: error:
>assigning to 'memcached_server_instance_st' (aka 'struct memcached_instance_st *') from
>      'const memcached_instance_st *' (aka 'const struct memcached_instance_st *') discards qualifiers
>      [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
>  330 |     server = memcached_server_instance_by_position(clone, i);
>      |            ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/lib/pr_fnmatch.c
+++ b/lib/pr_fnmatch.c
@@ -355,10 +355,7 @@ is_char_class (const wchar_t *wcs)
 
 
 int
-pr_fnmatch (pattern, string, flags)
-     const char *pattern;
-     const char *string;
-     int flags;
+pr_fnmatch (const char *pattern, const char *string, int flags)
 {
 # if HANDLE_MULTIBYTE
   if (__builtin_expect (MB_CUR_MAX, 1) != 1)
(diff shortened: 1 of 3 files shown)


Castaglia:
Could you provide some details for reproducing this, such as platform, compiler, compiler version used to encounter these issues?

I should also mention that the `libmemcached` library, used currently for the `--enable-memcache` implementation, is no longer well supported.  However, https://github.com/awesomized/libmemcached is a replacement for it; if you are using this library, that too would be useful to know.  Thanks!

Issue #1984, opened 2025-09-29
mod_quotatab_ldap interactions can lead to segfault due to stale pointer


a270443177:
# FTP File Deletion Exception and Technical Details
There is a user with permissions to access and write to all users' home directories. When the user **ftpuser** (user group: `ftp`) uploads the file `test.txt` to the user **user1** (user groups: `user1`, `ftp`), an abnormal interruption occurs when **user1** attempts to delete this file.

Through source code analysis and AI detection, it is identified that the exception occurs during the execution of the `ldaptab_lookup` method, specifically when the following code segments are run:
```c
ldap_cmd = pr_cmd_alloc(tmp_pool, 1, name);
ldap_res = pr_module_call(ldap_cmdtab->m, ldap_cmdtab->handler, ldap_cmd);
```

## Relevant Configurations and Logs
(Please fill in the specific configuration content and log information in this section, e.g., user permission configuration, FTP service configuration, system error logs, etc.)

**./proftpd  -V**
```c
Compile-time Settings:
  Version: 1.3.9 (stable)
  Platform: LINUX [Linux 3.10.0-1160.el7.x86_64 x86_64]
  OS/Release:
    NAME="Red Hat Enterprise Linux Server"
    VERSION="7.9 (Maipo)"
    ID="rhel"
    ID_LIKE="fedora"
    VARIANT="Server"
    VARIANT_ID="server"
    VERSION_ID="7.9"
    PRETTY_NAME="Red Hat Enterprise Linux Server 7.9 (Maipo)"
    CPE_NAME="cpe:/o:redhat:enterprise_linux:7.9:GA:server"
    
    REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
    REDHAT_BUGZILLA_PRODUCT_VERSION=7.9
    REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
    REDHAT_SUPPORT_PRODUCT_VERSION="7.9"
  Built: Mon Sep 29 2025 14:24:36 CST
  Built With:
    configure  '--prefix=/usr/local/proftpd' '--enable-nls' '--enable-openssl' '--enable-shadow' '--with-modules=mod_quotatab:mod_quotatab_file:mod_quotatab_ldap:mod_ldap:mod_tls:mod_sftp:mod_sftp_ldap:mod_sftp_pam' 'LDFLAGS=-ldl'

[...]

PR #1987, opened 2025-10-01
Issue #1984: Make sure that mod_ldap clears out the `cached_quota` po…

Castaglia:
…inter when its parent pool is destroyed, to avoid any stale pointers and related segfaults.

### DIFF
--- a/contrib/mod_ldap.c
+++ b/contrib/mod_ldap.c
@@ -1257,15 +1257,20 @@ static struct group *pr_ldap_group_lookup(pool *p, char *filter_template,
   return gr;
 }
 
+static void cached_quota_cleanup(void *event_data) {
+  cached_quota = NULL;
+}
+
 static void parse_quota(pool *p, const char *replace, char *str) {
   char **elts, *token;
 
   if (cached_quota == NULL) {
     cached_quota = make_array(p, 9, sizeof(char *));
+    register_cleanup2(p, NULL, cached_quota_cleanup);
   }
 
   elts = (char **) cached_quota->elts;
-  elts[0] = pstrdup(session.pool, replace);
+  elts[0] = pstrdup(p, replace);
   cached_quota->nelts = 1;
 
   (void) pr_log_writefile(ldap_logfd, MOD_LDAP_VERSION,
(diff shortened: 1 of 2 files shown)


contrib/mod_ldap.c
```diff
@@ -1257,23 +1257,28 @@ static struct group *pr_ldap_group_lookup(pool *p, char *filter_template,
...
   char **elts, *token;
 
   if (cached_quota == NULL) {
     cached_quota = make_array(p, 9, sizeof(char *));
+    register_cleanup2(p, NULL, cached_quota_cleanup);
   }
 
   elts = (char **) cached_quota->elts;
-  elts[0] = pstrdup(session.pool, replace);
+  elts[0] = pstrdup(p, replace);
```
Castaglia, line 1273: I'm not sure why we were allocating the _contents_ of this structure out of the long-lived `session.pool`, yet using the given pool for the structure memory.  This is both _a)_ confusing, and _b)_ possibly causing small amounts of memory leakage across the lifetime of a single session.

contrib/mod_ldap.c
```diff
@@ -1257,23 +1257,28 @@ static struct group *pr_ldap_group_lookup(pool *p, char *filter_template,
...
+static void cached_quota_cleanup(void *event_data) {
+  cached_quota = NULL;
+}
+
 static void parse_quota(pool *p, const char *replace, char *str) {
   char **elts, *token;
 
   if (cached_quota == NULL) {
     cached_quota = make_array(p, 9, sizeof(char *));
+    register_cleanup2(p, NULL, cached_quota_cleanup);
```
Castaglia, line 1269: At some point, I start to wonder whether having this `cached_quota` pointer makes sense at all.  The code flow would certainly be simpler if it was removed entirely.  And the mod_quotatab module does not look up the quota limits for users that often, I feel.

contrib/mod_quotatab_ldap.c
```diff
@@ -65,17 +65,19 @@ static unsigned char ldaptab_lookup(quota_table_t *ldaptab, void *ptr,
   ldap_cmd = pr_cmd_alloc(tmp_pool, 1, name);
   ldap_res = pr_module_call(ldap_cmdtab->m, ldap_cmdtab->handler, ldap_cmd);
 
-  destroy_pool(tmp_pool);
-
```
Castaglia, line 69: This _might_ be destroying this temporary pool too soon; the returned `ldap_res` structure may have used this pool, and we reference that structure _after_ the temporary pool is destroyed.  So let's destroy the pool only after we've done everything else.

Issue #2020, opened 2026-01-19
Out-of-bounds read in pr_netio_telnet_gets2 when input buffer starts with LF (UB/DoS)


yuyz-cyber:
*Summary*
A 1-byte out-of-bounds read exists in the FTP control-channel line reader: if a freshly read buffer chunk starts with `\n`, `pr_netio_telnet_gets2()` may access `buf[-1]`, triggering undefined behavior (UB) and potentially a crash (DoS).

*Details*
In `netio.c`, `pr_netio_telnet_gets2()` uses the following condition in its read/scan loop:
```
(*pbuf->current == '\n' && *(pbuf->current - 1) != '\r')
```
If `pbuf->current == pbuf->buf` and the first byte of the newly read data is `\n`, the expression dereferences `*(pbuf->current - 1)` _i.e._ `buf[-1]` (OOB read / UB).
This code path is reachable via the control connection command read loop: pr_cmd_read() → pr_netio_telnet_gets2().

*PoC*
This is a 1-byte OOB read, so whether it immediately crashes depends on allocator/memory layout/build options; lack of a crash does not mean the UB is absent.
Start ProFTPD (any normal config; `-X` helps for a single-session repro):
```
./proftpd -n -X -c /path/to/proftpd.conf
```
Connect to the control port and send a single LF byte (`\n`, without `\r`):
```
python3 - <<'PY'
import socket
s = socket.create_connection(("127.0.0.1", 21))
s.recv(4096)  # banner
s.sendall(b"\n")
PY
```
Observe server behavior/logs (disconnect, crash, or no obvious symptom are all possible due to UB).

*Impact*
Undefined behavior via 1-byte out-of-bounds read on the FTP control connection parsing path; can lead to process crash (DoS) under some memory layouts/implementations.

PR #2023, opened 2026-01-20
Issue #2020: Guard against inadvertently accessing bytes before the a…

Castaglia:
…llocated memory range when processing FTP control connection characters.

In certain conditions, we might use a -1 index, thus leading to a single byte read of out-of-bounds memory.

### DIFF
--- a/src/netio.c
+++ b/src/netio.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD - FTP server daemon
- * Copyright (c) 2001-2022 The ProFTPD Project team
+ * Copyright (c) 2001-2026 The ProFTPD Project team
  *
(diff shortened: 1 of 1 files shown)


src/netio.c
```diff
@@ -1836,7 +1840,8 @@ int pr_netio_telnet_gets2(char *buf, size_t bufsz,
        * turning the copied data from Telnet CRLF line termination to
        * Unix LF line termination.
        */
-      if (*(bp-1) == '\r') {
+      if (bp > buf &&
```
Castaglia: This added `bp > buf` check asserts the same as above, namely that the pointer is at least one byte larger/ahead of the buffer start, thus preventing the -1 out-of-bounds memory index/access.

PR #2026, opened 2026-01-28
`src/encode.c`: Fix call to `iconv` to fix the CI

hartwork:
Neither latest FreeBSD nor latest Solaris want `const char **` input.

On FreeBSD, [the symptom](https://github.com/proftpd/proftpd/runs/60830959156) was:
```
encode.c: In function 'str_convert':
encode.c:71:25: error: passing argument 2 of 'libiconv' from incompatible pointer type [-Wincompatible-pointer-types]
   71 |     nconv = iconv(conv, &inbuf, inbuflen, &outbuf, outbuflen);
      |                         ^~~~~~
      |                         |
      |                         const char **
In file included from encode.c:32:
/usr/local/include/iconv.h:84:43: note: expected 'char **' but argument is of type 'const char **'
   84 | extern size_t iconv (iconv_t cd,  char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
      |                                   ~~~~~~~~^~~~~
```

CC @Castaglia

### DIFF
--- a/contrib/mod_sftp/utf8.c
+++ b/contrib/mod_sftp/utf8.c
@@ -51,17 +51,7 @@ static int utf8_convert(iconv_t conv, const char *inbuf, size_t *inbuflen,
 
     pr_signals_handle();
 
-    /* Solaris/FreeBSD's iconv(3) takes a const char ** for the input buffer,
-     * whereas Linux/Mac OSX/GNU iconv(3) use char ** for the input buffer.
-     */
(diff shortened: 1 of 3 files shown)


Castaglia:
FYI, I'm inclined to use https://github.com/proftpd/proftpd/pull/2027 instead.

For compiler flags of this nature, you might experiment with _e.g._ `-Wincompatible-pointer-types`.

Issue #2047, opened 2026-04-14
FTP sessions could be prematurely closed during updates of system time/date


wiwid84:
I have noticed that the FTP session gets closed when there's simultaneous system time update via PTP or NTP.
Here's a log snippet:
    `FTP session opened`
    `Login timeout exceeded, disconnected`
    `FTP session closed.`

I believe the reason for that is alarm handler in _src/timers.c_ use standard "wall clock" time (CLOCK_REALTIME),
which can jump forwards or backwards if the user changes the system time or if an NTP/PTP synchronizes it.

A potential solution would be to use CLOCK_MONOTONIC which is guaranteed to only move forward.

Castaglia:
This should now be addressed in master.  Thank you!

PR #2048, opened 2026-04-15
Issue #2047: Use CLOCK_MONOTONIC in alarm handler

wiwid84:
FTP session can get get closed when there's simultaneous system time update via PTP or NTP.
Unlike the standard "wall clock" time (CLOCK_REALTIME), which can jump forwards or backwards if the user changes the system time or if an PTP/NTP synchronizes it, CLOCK_MONOTONIC is guaranteed to only move forward.
Fixes issue #2047.

### DIFF
--- a/src/timers.c
+++ b/src/timers.c
@@ -231,8 +231,12 @@ static RETSIGTYPE sig_alarm(int signo) {
   /* Reset the alarm */
   total_time += current_timeout;
   if (current_timeout) {
+#if defined(HAVE_CLOCK_GETTIME) && defined(HAVE_DECL_CLOCK_MONOTONIC)
     clock_gettime(CLOCK_MONOTONIC, &ts);
     alarmed_time = ts.tv_sec;
+#else
+    alarmed_time = time(NULL);
+#endif /* HAVE_CLOCK_GETTIME && HAVE_DECL_CLOCK_MONOTONIC*/
     alarm(current_timeout);
   }
 }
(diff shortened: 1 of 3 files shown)


src/timers.c
```diff
@@ -230,7 +231,8 @@ static RETSIGTYPE sig_alarm(int signo) {
   /* Reset the alarm */
   total_time += current_timeout;
   if (current_timeout) {
-    alarmed_time = time(NULL);
+    clock_gettime(CLOCK_MONOTONIC, &ts);
```
Castaglia, line 235: I've run into some systems (usually older ones) that don't define/include `CLOCK_MONOTONIC`; thus this may need some Autoconf support, using the existing `time(3)` usage as a fallback for when `CLOCK_MONOTIC` support is lacking.

I'm thinking we'd want `configure` to check for the `clock_gettime(3)` function, _and_ that `CLOCK_MONOTONIC` is defined.

Castaglia:
Thanks for this!  I think it'll be a great refinement to the core code.

src/timers.c
wiwid84: I can do that if you'd like

Castaglia:
Mind if I push some polishes to this branch of yours?

wiwid84:
I do not mind, go ahead ;)

śr., 15 kwi 2026, 21:47 użytkownik TJ Saunders ***@***.***>
napisał:

Castaglia:
Closing this PR per https://github.com/proftpd/proftpd/pull/2051  Thanks again for your work on this!

Issue #490, opened 2017-04-23
Support syntax checks on AuthUserFiles, AuthGroupFiles on startup


Castaglia:
To help catch any possible syntax errors in configured `AuthUserFiles` and `AuthGroupFiles`, ProFTPD should perform, on startup/restart, a basic syntax check of each entry in these files.  This can help to debug/detect any badly formatted files early.

These files could possibly be large, so such syntax checks should be configurable via some `Option` directive.

Castaglia:
If syntax errors *are* detected in the configured files, `mod_auth_file` should _log_ about them (_e.g._ at the `WARNING` log level), but *not* stop the server.  Consider the case where ProFTPD is regularly restarted via `SIGHUP`, by a program like `logrotate`.  It would be unexpected if the server died "randomly" due to such a syntax error.  So the best (and most appropriate behavior in this case, IMHO) is to detect and log the issue.

Castaglia:
Part of this should involve refactoring the existing [lib/pwgrent.c](https://github.com/proftpd/proftpd/blob/master/lib/pwgrent.c) code into `mod_auth_file` proper; those routines are very similar to ones we'd want to use for performing such syntax checks, albeit with built-in error accumulators for logging.

Castaglia:
See also this [forums thread](https://forums.proftpd.org/smf/index.php/topic,12270.0.html), where a syntax check of the `AuthGroupFile` on startup would have helped provide more specific info earlier in the debugging process.

Castaglia:
What if the `-t/--configtest` command-line option functionality could be extended to perform these `AuthUserFile`, `AuthGroupFile` syntax checks?

deba12:
Sounds really good. I like it.

Castaglia:
I ended going with an `AuthFileOption` for this, given that `-t/--configtest` isn't easily conveyed to modules.

To perform a syntax check on the `AuthUserFile`, `AuthGroupFile`, then, one would use:
```
AuthFileOptions SyntaxCheck
AuthUserFile /path/to/ftpd.passwd
AuthGroupFile /path/to/ftpd.group
```
Any syntax issues found will cause ProFTPD to log the issues, and fail to start up.

Castaglia:
Fixed in master.

PR #2055, opened 2026-04-30
Issue #490: Properly detect and handle the edge case of an AuthGroupF…

Castaglia:
…ile with a completely missing/absent members field.

### DIFF
--- a/modules/mod_auth_file.c
+++ b/modules/mod_auth_file.c
@@ -1,7 +1,7 @@
 /*
  * ProFTPD: mod_auth_file - file-based authentication module that supports
  *                          restrictions on the file contents
- * Copyright (c) 2002-2022 The ProFTPD Project team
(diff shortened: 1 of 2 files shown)


modules/mod_auth_file.c
```diff
@@ -494,7 +494,7 @@ static struct group *af_parse_grp(const char *buf, unsigned int lineno,
     }
   }
 
-  if (i < (NGRPFIELDS - 1)) {
+  if (i != NGRPFIELDS) {
```
Castaglia, line 497: Why was this comparison erroneously allowing a missing members field?  I'm not sure; `git blame` shows that this line of code has not changed in 24 years, which means it may predate my maintainership.  Perhaps the code was being lenient to handle accidentally malformed AuthGroupFile entries?

PR #2061, opened 2026-05-05
Adhere more closely to the behavior expected by systemd, by providing…

Castaglia:
… a monotonic timestamp in our "reloading" message.

### DIFF
--- a/modules/mod_systemd.c
+++ b/modules/mod_systemd.c
@@ -1,6 +1,6 @@
 /*
  * ProFTPD: mod_systemd -- provides systemd "socket activation" support
- * Copyright (c) 2025 TJ Saunders
+ * Copyright (c) 2025-2026 TJ Saunders
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -23,7 +23,7 @@
 #include "conf.h"
 #include "privs.h"
 
-#define MOD_SYSTEMD_VERSION	"mod_systemd/0.1"
+#define MOD_SYSTEMD_VERSION	"mod_systemd/0.2"
 
 /* Make sure the version of ProFTPD is as necessary. */
 #if PROFTPD_VERSION_NUMBER < 0x0001030A01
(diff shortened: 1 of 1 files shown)


Castaglia:
@pghmcfc Might this help, in your testing of when `ExecReload=` is omitted from the service file?  Unfortunately I don't have a (sufficiently enabled) systemd system for verifying this; I'm currently constrained to Docker containers, which don't play all that nicely with systemd for situations like this.

modules/mod_systemd.c
```diff
@@ -207,7 +224,13 @@ static void systemd_restart_ev(const void *event_data, void *user_data) {
     return;
   }
 
-  notify_systemd("RELOADING=1");
+  get_current_time(&now);
+
+  memset(text, '\0', sizeof(text));
+  pr_snprintf(text, sizeof(text)-1, "RELOADING=1\nMONOTONIC_USEC=%llu",
```
Castaglia, line 230: Per https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html a service is expected to provide a "monotonic" timestamp as well, for a `Type=notify-reload` configuration:
> When initiating the reload process the service is expected to reply with a notification message via [sd_notify(3)](https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html#) that contains the "RELOADING=1" field in combination with "MONOTONIC_USEC=" set to the current monotonic time (i.e. CLOCK_MONOTONIC in [clock_gettime(2)](https://man7.org/linux/man-pages/man2/clock_gettime.2.html)) in μs, formatted as decimal string. Once reloading is complete another notification message must be sent, containing "READY=1". Using this service type and implementing this reload protocol is an efficient alternative to providing an ExecReload= command for reloading of the service's configuration.

Castaglia:
Note to self: if/when this is merged, bump mod_systemd's version requirement to be on ProFTPD 1.3.10rc2, since that is where detection of `clock_gettime(2)`, `CLOCK_MONOTONIC` first appeared.

pghmcfc:
@Castaglia that seems to have fixed it, thanks (tried with just this change, so not testing the HAVE_CLOCK_MONOTONIC branch).

PR #2068, opened 2026-05-07
Issue #2036: Impose max message size limits for received Controls req…

Castaglia:
…uests and responses.

While clients are expected to be trusted, one could attempt to crash the daemon by sending a length prefix for a very large message, causing the daemon to allocate much memory, possibly leading to an out-of-memory condition. Similarly, an adversely large length prefix for a response could trigger undue memory allocations in the client.

### DIFF
--- a/src/ctrls.c
+++ b/src/ctrls.c
@@ -42,6 +42,9 @@
 
 #include "mod_ctrls.h"
 
+#define CTRLS_MAX_REQ_SIZE	1024
+#define CTRLS_MAX_RESP_SIZE	(1024 * 1024)
+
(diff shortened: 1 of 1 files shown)


src/ctrls.c
```diff
@@ -42,6 +42,9 @@
 
 #include "mod_ctrls.h"
 
+#define CTRLS_MAX_REQ_SIZE	1024
+#define CTRLS_MAX_RESP_SIZE	(1024 * 1024)
```
Castaglia, line 46: Yes, these limits (1K requests, 1M responses) are somewhat arbitrary.  However, I'm hoping that these limits will not hinder any existing use cases.

PR #2069, opened 2026-05-07
Try opting into use of Nodejs v24 for our CI workflow, per pending de…

Castaglia:
…precation of Nodejs v20 on the GitHub runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,9 +40,10 @@ jobs:
     environment: CI
 
     env:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -54,7 +55,6 @@ jobs:
           - gcc
         container:
           - almalinux:8
-          - alpine:3.15
```
Castaglia, line 57: Looks like this older Alpine image lacks the necessary libc support for Nodejs v24; let's retire it to the "legacy platforms" workflow.

PR #2070, opened 2026-05-07
Issue #2036: Add some unit tests for the "message too large" conditio…

Castaglia:
…ns just implemented in the Controls API.

### DIFF
--- a/src/ctrls.c
+++ b/src/ctrls.c
@@ -618,21 +618,22 @@ int pr_ctrls_recv_request(pr_ctrls_cl_t *cl) {
   xerrno = errno;
 
   if (nread < 0) {
+    pr_signals_unblock();
+
     pr_trace_msg(trace_channel, 3,
       "error reading %lu bytes of request message size: %s",
       sizeof(msglen), strerror(xerrno));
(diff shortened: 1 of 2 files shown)


src/ctrls.c
```diff
@@ -646,10 +647,11 @@ int pr_ctrls_recv_request(pr_ctrls_cl_t *cl) {
 
   /* Impose max request size limit here, for Issue #2036. */
   if (msglen > CTRLS_MAX_REQ_SIZE) {
+    destroy_pool(tmp_pool);
```
Castaglia, line 650: Make sure we don't leak memory here by properly destroying our temporary pool!  (Thank you ASan unit tests for detecting this!)

src/ctrls.c
```diff
@@ -618,21 +618,22 @@ int pr_ctrls_recv_request(pr_ctrls_cl_t *cl) {
   xerrno = errno;
 
   if (nread < 0) {
+    pr_signals_unblock();
```
Castaglia, line 621: I should be stylistically consistent (hopefully), and unblock signals as early as possible.

PR #2074, opened 2026-05-09
Properly check the response length provided by the SSH agent, not jus…

Castaglia:
…t the number of bytes read.

Thanks to Marco Lux of Mantodea Security for suggesting this.

### DIFF
--- a/contrib/mod_sftp/agent.c
+++ b/contrib/mod_sftp/agent.c
@@ -154,15 +154,10 @@ static unsigned char *agent_request(pool *p, int fd, const char *path,
     return NULL;
   }
 
-  /* Sanity check the returned length; we could be dealing with a buggy
-   * client (or something else is injecting data into the Unix domain socket).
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/agent.c
```diff
@@ -154,15 +154,10 @@ static unsigned char *agent_request(pool *p, int fd, const char *path,
     return NULL;
   }
 
-  /* Sanity check the returned length; we could be dealing with a buggy
-   * client (or something else is injecting data into the Unix domain socket).
-   * Best be conservative: if we get a response length of more than 256KB,
-   * it's too big.  (What about very long lists of keys, and/or large keys?)
-   */
-  if (res > AGENT_REPLY_MAXSZ) {
+  if (res < sizeof(uint32_t)) {
```
Castaglia, line 157: `res` here refers to the number of bytes as returned by `read(2)` from the socket; it does not refer to the length of response data that the agent will be sending.

Issue #1963, opened 2025-07-03
Resumed SSL_SESSION management in mod_tls leads to memory growth, infinite loop using newer OpenSSL versions


SmartSceptic:
## 1 — Summary

*Two name‑based SNI vhosts share the same IP (`ftp.vendor.com`, `ftp.other.com`). Only sessions to **ftp.vendor.com** leak.*

One external FTPS client (redacted as **Vendor\_FTP**) leaves its ProFTPD child process in **CLOSE\_WAIT** indefinitely; the child’s RSS climbs past **12 GB** until a watchdog kills it, after which the master process may be OOM‑killed.

The leak happens **only** when the client ACKs the server’s FIN and **immediately sends RST**, never sending a TLS **close\_notify**.

Similar or related issues already discussed:

* \#330 — memory growth with long‑running sessions
* \#1074 — SNI cleanup problems
* \#1106 — crash in `tls_sni_sess_tab_delete_cb()`
* \#795  — loop in `SSL_CTX_flush_sessions()` / `SSL_CTX_free()`

Our case combines these symptoms: infinite `CLOSE_WAIT` *and* unbounded RSS growth, always inside `tls_sni_sess_tab_delete_cb()`.

---

## 2 — Environment

* **OS / Kernel:** Oracle Linux 9.6 (kernel 5.14)
* **ProFTPD:** 1.3.8d (`proftpd-1.3.8d-1.el9.x86_64`)
* **mod\_tls:** 2.9.2
* **OpenSSL:** 3.2.2
* **Init:** systemd
* **Passive ports:** 65300–65334
* **Modules loaded:** `mod_sftp_pam`, `mod_sftp`, `mod_lang`, `mod_ctrls`, `mod_cap`, `mod_memcache`, `mod_tls`, `mod_pgsql`

---

## 3 — Minimal configuration (redacted)

```apache
<Global>
  ServerIdent "FTP Server ready."
  PassivePorts 65300 65334
  TLSLog /var/log/proftpd/tls.log
  TLSProtocol TLSv1.2 TLSv1.3
  TLSOptions  NoSessionReuseRequired ReleaseTokens
  TimeoutIdle 300
  TimeoutLinger 10
</Global>

<VirtualHost 0.0.0.0>
  ServerAlias ftp.vendor.com
  TLSEngine   on
  TLSRSACertificateFile   /path/vendor.pem
  TLSRSACertificateKeyFile /path/vendor.key
</VirtualHost>

[...]

PR #2076, opened 2026-05-10
Issue #2056: Initial testing with OpenSSL 3.5.5 shows that the prior …

Castaglia:
…changes for Issue #1963 were insufficient; adding a manual refcount increment works.


Castaglia:
I still want to run some tests of this with older OpenSSL versions (1.x, 3.0.x, 3.2.x, _etc_) to avoid more possible regressions.

PR #2081, opened 2026-05-12
Issue #2041: Implement support for using private keys from PKCS11 tok…

Castaglia:
…ens.

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -75,7 +75,11 @@
 # ifdef PR_USE_OPENSSL_OCSP
 #  include <openssl/ocsp.h>
 # endif /* PR_USE_OPENSSL_OCSP */
-#endif
+# if defined(PR_USE_OPENSSL_OSSL_PROVIDER_LOAD)
+#  include <openssl/provider.h>
+#  include <openssl/store.h>
+# endif /* PR_USE_OPENSSL_OSSL_PROVIDER_LOAD */
+#endif /* OpenSSL-0.9.7 and later */
 #ifdef PR_USE_OPENSSL_ECC
 # include <openssl/ec.h>
 # include <openssl/ecdh.h>
@@ -85,7 +89,7 @@
 # include <sys/mman.h>
 #endif
 
-#define MOD_TLS_VERSION		"mod_tls/2.9.2"
+#define MOD_TLS_VERSION		"mod_tls/2.9.3"
 
 /* Make sure the version of proftpd is as necessary. */
 #if PROFTPD_VERSION_NUMBER < 0x0001030602
@@ -757,6 +761,8 @@ static SSL_CTX *ssl_ctx = NULL;
 static array_header *tls_tmp_dhs = NULL;
 static RSA *tls_tmp_rsa = NULL;
 
+static int have_pkcs11_support = -1;
+
 static void tls_exit_ev(const void *event_data, void *user_data);
 
 /* SSL/TLS support functions */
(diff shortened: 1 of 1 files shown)


contrib/mod_tls.c
```diff
@@ -874,12 +880,227 @@ static int tls_data_need_init_handshake = TRUE;
 
 static const char *timing_channel = "timing";
 
+static void tls_use_pkcs11(void) {
+  int use_pkcs11 = FALSE;
+#if defined(PR_USE_OPENSSL_OSSL_PROVIDER_LOAD)
+  OSSL_PROVIDER *pkcs11;
+#endif /* PR_USE_OPENSSL_OSSL_PROVIDER_LOAD */
+
+  if (have_pkcs11_support != -1) {
```
Castaglia, line 889: This is to do runtime detection of PKCS11 provider presence/support.  We only need to do this detection once, so the static `have_pkcs11_support` variable caches the detection results.

contrib/mod_tls.c
```diff
@@ -14184,62 +14430,19 @@ MODRET set_tlsdsacertfile(cmd_rec *cmd) {
...
+  char *path, *errors = NULL;
 
   CHECK_ARGS(cmd, 1);
   CHECK_CONF(cmd, CONF_ROOT|CONF_VIRTUAL|CONF_GLOBAL);
 
   path = cmd->argv[1];
 
   PRIVS_ROOT
-
-  ctx = SSL_CTX_new(SSLv23_server_method());
```
Castaglia, line 14197: All of this duplicated code, in the `TLS{DSA,EC,RSA}CertificateKeyFile` handlers, has been refactored into the new `tls_keyfile_check` function.  Which made detection/handling of PKCS11 URIs _much_ easier.

contrib/mod_tls.c
```diff
@@ -874,12 +881,227 @@ static int tls_data_need_init_handshake = TRUE;
...
+
+  OSSL_STORE_close(store_ctx);
+  return pkey;
+#endif /* PR_USE_OPENSSL_OSSL_PROVIDER_LOAD */
+
+  errno = ENOSYS;
+  return NULL;
+}
+
+static EVP_PKEY *prompt_pkcs11_key(pool *p, const char *text,
```
Castaglia: I'm now wondering whether this separate "prompt for PKCS11 PIN with custom UI_METHOD" is even needed.  The custom UI prompt text isn't used, in my local tests; it seems as if the PKCS11 provider relies on the driver (SoftHSM2 here) to do the prompting.  In which case, this function isn't used at all.  Let's find out!

Castaglia: Answer: No, this `prompt_pkcs11_key()` function with a custom UI_METHOD was not actually needed.  Nice.

contrib/mod_tls.c
```diff
@@ -874,12 +880,187 @@ static int tls_data_need_init_handshake = TRUE;
...
+
+  store_info = OSSL_STORE_load(store_ctx);
+  if (store_info == NULL) {
+    *errors = pstrcat(p, "unable to load info for PKCS11 '", text, "': ",
+      tls_get_errors2(p), NULL);
+    OSSL_STORE_close(store_ctx);
+    return NULL;
+  }
+
+  pkey = OSSL_STORE_INFO_get1_PKEY(store_info);
```
Castaglia: Here we _assume_ that this "store" contains a private key, and not anything else.  A fair assumption for the given use case, but this bit of code may need some attention in the future.

PR #2087, opened 2026-05-17
Address some compiler warnings in mod_sftp; no functional change.

Castaglia:


contrib/mod_sftp/agent.c
```diff
@@ -154,7 +154,7 @@ static unsigned char *agent_request(pool *p, int fd, const char *path,
     return NULL;
   }
 
-  if (res < sizeof(uint32_t)) {
+  if ((size_t) res < sizeof(uint32_t)) {
```
Castaglia, line 157: We know that `res` cannot be negative here, due to the earlier check on line 146.

PR #2092, opened 2026-05-19
The mod_copy module was not properly detecting when the destination p…

Castaglia:
…ath is a dangling symlink, and thus was not enforcing the "AllowOverwrite off" configuration for such cases.

Thanks to skove for reporting this issue.

### DIFF
--- a/contrib/mod_copy.c
+++ b/contrib/mod_copy.c
@@ -339,7 +339,7 @@ static int copy_dir(pool *p, const char *src_dir, const char *dst_dir,
 }
 
 static int copy_paths(pool *p, const char *from, const char *to) {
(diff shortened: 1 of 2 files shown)


contrib/mod_copy.c
```diff
@@ -339,7 +339,7 @@ static int copy_dir(pool *p, const char *src_dir, const char *dst_dir,
 }
 
 static int copy_paths(pool *p, const char *from, const char *to) {
-  struct stat st;
+  struct stat from_st, to_st;
```
Castaglia, line 342: While not functionally necessary, this helps to make the code more explicit, and thus more legible.

PR #2112, opened 2026-05-30
A researcher reported a one-byte OOB write in `pr_fsio_gets()`, for w…

Castaglia:
…riting a single NUL byte past the end of the provided buffer, in situations where the data fills the buffer exactly.

### DIFF
--- a/src/fsio.c
+++ b/src/fsio.c
@@ -6682,33 +6682,35 @@ char *pr_fsio_getpipebuf(pool *p, int fd, long *bufsz) {
   return buf;
 }
 
-char *pr_fsio_gets(char *buf, size_t size, pr_fh_t *fh) {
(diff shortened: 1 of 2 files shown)


src/fsio.c
```diff
@@ -6762,7 +6764,17 @@ char *pr_fsio_gets(char *buf, size_t size, pr_fh_t *fh) {
     }
   }
 
-  *bp = '\0';
+  if (size > 0) {
```
Castaglia, line 6767: Without this bounds checking, the (updated) unit tests, run under ASan, would rightly report memory corruption from the OOB write, where `bp` points at the byte just past the end of the provided buffer.

Castaglia:
Specifically, this issue was reported by Marcello Maugeri of CybPass Limited.

PR #2113, opened 2026-05-30
Marcello Maugeri of CybPass Limited reported that malformed AuthGroup…

Castaglia:
…File entries could possibly leave stale data in the static `grent` struct for later reuse.

### DIFF
--- a/modules/mod_auth_file.c
+++ b/modules/mod_auth_file.c
@@ -448,39 +448,46 @@ static char **af_getgrmems(char *s) {
 
 static struct group *af_parse_grp(const char *buf, unsigned int lineno,
     int flags) {
-  unsigned int i;
+  register unsigned int i;
+  unsigned int sz;
   char *cp;
 
-  i = strlen(buf) + 1;
+  sz = strlen(buf) + 1;
 
   if (grpbuf == NULL) {
-    grpbufsz = i;
+    grpbufsz = sz;
     grpbuf = malloc(grpbufsz);
 
-  } else if (grpbufsz < (size_t) i) {
+  } else if (grpbufsz < (size_t) sz) {
     char *new_buf;
 
     pr_trace_msg(trace_channel, 19,
       "parsing group line '%s' (%lu bytes), allocating %lu bytes via "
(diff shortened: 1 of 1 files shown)


Castaglia:
In most/current use cases, this is not issue, since well-formed group entries which lead to successful authentication will populate that `grent` struct properly, and the proper values will be _copied_ during the authentication process.  Thus this change is a defense-in-depth measure, to ensure that future changes to workflows will not unintentionally use potentially-state group data.

modules/mod_auth_file.c
```diff
@@ -448,39 +448,46 @@ static char **af_getgrmems(char *s) {
 
 static struct group *af_parse_grp(const char *buf, unsigned int lineno,
     int flags) {
-  unsigned int i;
+  register unsigned int i;
+  unsigned int sz;
```
Castaglia, line 452: These new variables and names help differentiate which is used to indicate a buffer size, and which is used for the conventional iterator variable for `for` loops.

modules/mod_auth_file.c
```diff
@@ -1088,9 +1109,7 @@ static int af_check_group_syntax(pool *p, const char *path) {
     grp = af_getgrent(p, flags, &bad_entry_count);
   }
 
-  pr_fsio_close(af_group_file->af_file_fh);
-  af_group_file->af_file_fh = NULL;
-  af_group_file->af_lineno = 0;
+  af_endgrent();
```
Castaglia, line 1112: This helps avoid code duplication, and increases reuse of all the cleanup code.

modules/mod_auth_file.c
```diff
@@ -1133,9 +1152,7 @@ static int af_check_user_syntax(pool *p, const char *path) {
     pwd = af_getpwent(p, flags, &bad_entry_count);
   }
 
-  pr_fsio_close(af_user_file->af_file_fh);
-  af_user_file->af_file_fh = NULL;
-  af_user_file->af_lineno = 0;
+  af_endpwent();
```
Castaglia, line 1155: This helps avoid code duplication, and increases reuse of all the cleanup code.

PR #2118, opened 2026-05-30
Implement some sanity checks on the length of extended attributes (xa…

Castaglia:
…ttrs) that can be requested via custom SFTP extensions.

Bankde Eakasit posited that this could be another vector to triggering excessive memory allocations.


contrib/mod_sftp/fxp.c
```diff
@@ -1904,7 +1906,7 @@ static char *fxp_strattrs(pool *p, struct stat *st, uint32_t *attr_flags) {
 
       if (fxp_session->client_version >= 6) {
         flags |= SSH2_FX_ATTR_LINK_COUNT;
-#ifdef PR_USE_XATTR
+#if defined(PR_USE_XATTR)
```
Castaglia, line 1909: Stylistic nit.

PR #2128, opened 2026-06-03
Run through mod_sftp's support for using hostkeys in an SSH agent, ma…

Castaglia:
…king sure the cases work.

In particular, I found that SSH agent-hosted Ed25519 hostkeys did not work at all.  And the support for Ed448 hostkeys was slightly lacking.  All better now.

### DIFF
--- a/contrib/mod_sftp/mod_sftp.c
+++ b/contrib/mod_sftp/mod_sftp.c
@@ -1600,11 +1600,12 @@ MODRET set_sftphostkey(cmd_rec *cmd) {
     flags |= SFTP_HOSTKEY_FL_CLEAR_ED25519_KEY;
   }
 
-  if (strncmp(cmd->argv[1], "agent:", 6) != 0 &&
+  path = cmd->argv[1];
+
+  if (strncmp(path, "agent:", 6) != 0 &&
       flags == 0) {
     int res, xerrno;
 
-    path = cmd->argv[1];
     if (*path != '/') {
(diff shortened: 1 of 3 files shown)


contrib/mod_sftp/mod_sftp.c
```diff
@@ -1600,11 +1600,12 @@ MODRET set_sftphostkey(cmd_rec *cmd) {
     flags |= SFTP_HOSTKEY_FL_CLEAR_ED25519_KEY;
   }
 
-  if (strncmp(cmd->argv[1], "agent:", 6) != 0 &&
+  path = cmd->argv[1];
```
Castaglia, line 1603: This was causing a glaringly obvious null pointer dereference.  Sigh.  I guess use of SSH agent-hosted public keys for mod_sftp is not a common scenario.

contrib/mod_sftp/mod_sftp.c
```diff
@@ -2949,7 +2950,8 @@ static int sftp_sess_init(void) {
   if (sftp_keys_have_dsa_hostkey() < 0 &&
       sftp_keys_have_rsa_hostkey() < 0 &&
       sftp_keys_have_ecdsa_hostkey(sftp_pool, NULL) < 0 &&
-      sftp_keys_have_ed25519_hostkey()) {
+      sftp_keys_have_ed25519_hostkey() < 0 &&
+      sftp_keys_have_ed448_hostkey() < 0) {
```
Castaglia, line 2954: This fixes the case where the only hostkey configured was an Ed448 hostkey.  Oops.

Issue #2135, opened 2026-06-06
ExecEnviron values not passed due to regression since 1.3.8.d


poire-z:
It seems ExecEnviron no longer work with 1.3.9b as it did with 1.3.8.d, possibly since 4856d23 (which landed in 1.3.9 ?)
(I don't see ExecEnviron being tested in tests/t/lib/ProFTPD/Tests/Modules/mod_exec.pm, so possibly missed while moving to the Jot API?)

Relevant bits from my proftpd.conf:
```
<Global>
[...]
ExecEngine  on
ExecLog     /logs/proftpd/mod_exec.log
ExecOptions logStderr sendStdout 
ExecTimeout 600

# These are not passed:
ExecEnviron PROFTPD_IP %a
ExecEnviron PROFTPD_HOSTNAME %h
ExecEnviron PROFTPD_USERNAME %u

ExecEnviron PROFTPD_IP2 "%a"
ExecEnviron PROFTPD_HOSTNAME2 "%h"
ExecEnviron PROFTPD_USERNAME2 "%u"

ExecEnviron PROFTPD_IP3 "ip3"
ExecEnviron PROFTPD_HOSTNAME3 "hostname3"
ExecEnviron PROFTPD_USERNAME3 "username3"

ExecEnviron PROFTPD_IP4 ip4
ExecEnviron PROFTPD_HOSTNAME4 hostname4
ExecEnviron PROFTPD_USERNAME4 username4

# These are passed:
# Set via SetEnv VSERVER "SSH-myserver" in a Virtual host below
ExecEnviron VSERVER -

ExecEnviron SFTP  -
ExecEnviron SFTP_CLIENT_COMPRESSION_ALGO  -
ExecEnviron SFTP_SERVER_COMPRESSION_ALGO  -

[...]

<Directory ~/in/*/*>
    # If adding %u %f to these, they are passed.
    ExecOnCommand STOR,APPE /scripts/receiver_proftpd
    ExecOnError   STOR,APPE /scripts/recerror_proftpd
</Directory>

[...]

<VirtualHost myhostnname>
SetEnv VSERVER "myserver"
```
(Also using mod_vroot in case it's relevant.)

Output of 'env' in my receiver_proftpd script
```
PROFTPD_HOSTNAME2=
PROFTPD_HOSTNAME3=
PROFTPD_HOSTNAME4=
PROFTPD_HOSTNAME=
PROFTPD_IP2=
PROFTPD_IP3=
[...]

PR #2136, opened 2026-06-07
Issue #2135: Fix the regression with the `ExecEnviron` directive.

Castaglia:


contrib/mod_exec.c
```diff
@@ -1580,7 +1580,8 @@ MODRET set_execenviron(cmd_rec *cmd) {
     logfmt[logfmt_len] = '\0';
   }
 
-  c->argv[1] = pstrndup(c->pool, (char *) logfmt, logfmt_len);
+  c->argv[1] = pstrdup(c->pool, text);
+  c->argv[2] = pstrndup(c->pool, (char *) logfmt, logfmt_len);
```
Castaglia, line 1584: The regression occurred because https://github.com/proftpd/proftpd/commit/4856d23f16d8ebce4203deb0a8fa1340ab20f161 provided the environment key, and the Jot API-parsed logfmt string -- but not the original text.  And that text is _expected_ to be there by the Jot API resolving function; if not present, the resolution does not properly occur.  Hence the loss of the expected `ExecEnviron` values.

PR #2138, opened 2026-06-08
mod_ldap: Add configurable objectClass, group type support, and session tracking

zx97:
…on tracking

- Add LDAPUserObjectClass directive (default: posixAccount)
- Add LDAPGroupObjectClass directive (default: posixGroup)
- Add LDAPGroupMemberAttr directive (default: memberUid)
- Add LDAPGroupType directive (static|dynamic, default: static)
- Add LDAPSessionTracking directive (RFC 5646)

Enables support for non-POSIX LDAP schemas (e.g., inetOrgPerson, groupOfNames, groupOfURLs) and adds session tracking for debugging clients behind NAT/load-balancers.

Updated documentation for all new directives.

### DIFF
--- a/contrib/mod_ldap.c
+++ b/contrib/mod_ldap.c
@@ -215,8 +215,10 @@ static int ldap_group_type_static = TRUE;
  */
 static int ldap_session_tracking = FALSE;
 
-/* Session tracking control OID (RFC 5646) */
-#define LDAP_CONTROL_SESSION	"1.3.6.1.4.1.4203.666.5.8"
+/* Session tracking control OID (RFC 5646 / OpenLDAP)
+ * OID: 1.3.6.1.4.1.21008.108.63.1
+ */
+#define LDAP_CONTROL_SESSION	"1.3.6.1.4.1.21008.108.63.1"
 
 static char *ldap_dn, *ldap_dnpass, *ldap_sasl_mechs = NULL,
             *ldap_user_basedn = NULL, *ldap_user_name_filter = NULL,
--- a/contrib/mod_ldap.c
+++ b/contrib/mod_ldap.c
@@ -291,25 +291,21 @@ static void pr_ldap_unbind(void) {
   ld = NULL;
 }
 
-/* Generate a unique session identifier for LDAP session tracking.
- * This ID is sent with each LDAP request via the Session Tracking Control,
- * allowing the LDAP server to log/correlation client sessions.
- *
- * Format: ProFTPD-session-<pid>-<timestamp>-<random>
+/* Get session identifier from mod_unique_id if available.
+ * The UNIQUE_ID environment variable is set by mod_unique_id.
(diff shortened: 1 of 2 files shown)


Castaglia:
Would you mind opening a GitHub ticket/issue for this?  In particular, I would like to learn more about the use cases you might have for these additions.  Thanks!

zx97:
I opened issue https://github.com/proftpd/proftpd/issues/2140 at your request.

contrib/mod_ldap.c
```diff
@@ -251,6 +289,88 @@ static void pr_ldap_unbind(void) {
   ld = NULL;
 }
 
+/* Generate a unique session identifier for LDAP session tracking.
+ * This ID is sent with each LDAP request via the Session Tracking Control,
+ * allowing the LDAP server to log/correlation client sessions.
+ *
+ * Format: ProFTPD-session-<pid>-<timestamp>-<random>
```
Castaglia: Think it would make sense to make use of [mod_unique_id](https://github.com/proftpd/proftpd/blob/master/contrib/mod_unique_id.c#L155)'s ID here, since it looks to be similar in nature?  That is, I assume that this session ID text needs only to be unique, correct?

Castaglia:
Thanks for your work on this!  I am not that familiar with LDAP dynamic groups, at the moment, so it may take me a little time to read up on this feature, and assess the (well documented/commented!) code you provided.  I may also want to break this into separate PRs, to isolate the particular bits of code and functionality; for example, the session tracking bits might be easier to evaluate as a separate PR.

contrib/mod_ldap.c
```diff
@@ -185,6 +185,39 @@ static array_header *ldap_servers = NULL;
...
+
+/* LDAP Session Tracking Control (RFC 5646 / draft-wahl-ldap-session).
+ * When enabled, adds a session identifier to LDAP requests, allowing the
+ * LDAP server to log/correlation client sessions for debugging/troubleshooting.
+ * Useful when clients are behind NAT or load-balancers.
+ */
+static int ldap_session_tracking = FALSE;
+
+/* Session tracking control OID (RFC 5646) */
+#define LDAP_CONTROL_SESSION	"1.3.6.1.4.1.4203.666.5.8"
```
Castaglia: Where does this particular OID come from?  I ask because OpenLDAP uses a different OID (see [here](https://github.com/openldap/openldap/blob/master/include/ldap.h#L363)), the one that appears in the Draft.

zx97: Good catch, I have fixed this:
`#define LDAP_CONTROL_SESSION    "1.3.6.1.4.1.21008.108.63.1"
`

contrib/mod_ldap.c
zx97: Thanks, I've updated the code to make use of mod_unique_id's UNIQUE_ID

Issue #2146, opened 2026-06-12
Stack buffer overflow in MLSD/MLST handling for long path names


Castaglia:
The `facts_mlinfo_fmt()` function in the mod_facts module builds per-file MLSD/MLST fact strings (e.g. `type=file;size=1234; filename`) into a caller-supplied 1024-byte stack buffer. It accumulates the written length via `buflen += len`, where `len` is the `int` return value of `pr_snprintf()`.
  
ProFTPD's `pr_snprintf()` (implemented in `src/support.c`) deviates from POSIX: on truncation it returns −1 and sets `errno = ENOSPC`, rather than returning the would-be written size. When `buflen == 0` and the `type=` fact write is truncated, the expression `buflen += len` computes `size_t(0) + int(-1)`, which in C is `SIZE_MAX` (unsigned wraparound). The subsequent `ptr = buf + buflen` places the write pointer one byte **before** the start of `buf`. The next `pr_snprintf()` call — writing the space-separated filename and CRLF — then writes into the adjacent stack frame.

ASAN confirms the write: `WRITE of size 13` at `buf[-1]`, underflowing the 1024-byte stack buffer allocated at `facts_mlinfobuf_add()` line 460.

Exploiting this requires an authenticated FTP session. The sole configuration precondition is `FactsOptions UseSlink` (non-default, but commonly enabled for RFC 3659 symlink reporting). The attacker must be able to create or cause to exist a symlink whose target path is ≥ 1004 characters in a directory they can list.

**Affected Version:** 1.3.10rc3 (all prior versions with `FactsOptions UseSlink` expected to be affected)  
**Component:** `mod_facts` module — `modules/mod_facts.c`, function `facts_mlinfo_fmt`  
**Reporter:** Yiyi Wang, Tsinghua University
**Date:** 2026-06-12

Castaglia:
Fixed in master now, and backported to the 1.3.9 branch.

PR #2147, opened 2026-06-12
Issue #2146: Long path names, especially for symlinks, can lead to ov…

Castaglia:
…erflowing the fixed buffer used for `MLSD/MLST` facts formatting.

Make sure we properly handle the case where the buffer is full, before attempting to write more data into it.  In addition, increase the size of the fixed buffers used, to better accommodate such long path names.

### DIFF
--- a/modules/mod_facts.c
+++ b/modules/mod_facts.c
@@ -79,6 +79,12 @@ struct mlinfo {
   const char *real_path;
 };
 
+/* Allocate buffers large enough for long path names to appear twice, once
+ * as the symlink target for _e.g._ the UseSlink FactsOption, as well as
+ * additional facts.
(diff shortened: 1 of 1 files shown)


modules/mod_facts.c
```diff
@@ -551,7 +619,7 @@ static int facts_mlinfo_get(struct mlinfo *info, const char *path,
   info->tm = pr_gmtime(info->pool, &(info->st.st_mtime));
 
   if (!S_ISDIR(info->st.st_mode)) {
-#ifdef S_ISLNK
+#if defined(S_ISLNK)
```
Castaglia, line 628: Stylistic nit; no functional change.

PR #2151, opened 2026-06-15
When changing the ownership of the `ControlsSocket` path, use a small…

Castaglia:
…er window of privileges.

### DIFF
--- a/modules/mod_ctrls.c
+++ b/modules/mod_ctrls.c
@@ -754,14 +754,19 @@ static int ctrls_timer_cb(CALLBACK_FRAME) {
   }
 
   if (first_time == TRUE) {
-    /* Change the ownership on the socket to that configured by the admin */
+    int res, xerrno;
+
+    /* Change the ownership on the socket to that configured by the admin. */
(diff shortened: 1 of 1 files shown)


modules/mod_ctrls.c
```diff
@@ -754,14 +754,19 @@ static int ctrls_timer_cb(CALLBACK_FRAME) {
...
   if (first_time == TRUE) {
-    /* Change the ownership on the socket to that configured by the admin */
+    int res, xerrno;
+
+    /* Change the ownership on the socket to that configured by the admin. */
     PRIVS_ROOT
-    if (chown(ctrls_sock_file, ctrls_sock_uid, ctrls_sock_gid) < 0) {
+    res = chown(ctrls_sock_file, ctrls_sock_uid, ctrls_sock_gid);
+    xerrno = errno;
+    PRIVS_RELINQUISH
```
Castaglia, line 763: This pattern is more consistent with privilege use elsewhere; we don't need to do the logging while privileged.

PR #2153, opened 2026-06-15
Incrementally improve coverage of Redis and Scoreboard APIs via testing.

Castaglia:


src/scoreboard.c
```diff
@@ -331,11 +331,6 @@ static int wlock_entry(int fd) {
 static int write_entry(int fd) {
   int res;
 
-  if (fd < 0) {
-    errno = EINVAL;
-    return -1;
-  }
```
Castaglia, line 337: All of the callers of `write_entry()` already perform this check themselves; no need to have it here as well.

PR #2154, opened 2026-06-16
When iterating over a linked list of ctrls, the `pr_ctrls_reset()` fu…

Castaglia:
…nction can corrupt the list by freeing the wrong pointer, which leads to inconsistent handling of control requests.

Thanks to Kamil Leoniak and Delphos Labs for discovering and reporting this issue.

### DIFF
--- a/src/ctrls.c
+++ b/src/ctrls.c
@@ -1917,7 +1917,7 @@ int pr_run_ctrls(module *mod, const char *action) {
 }
 
 int pr_ctrls_reset(void) {
-  pr_ctrls_t *ctrl = NULL;
+  pr_ctrls_t *ctrl = NULL, *next_ctrl = NULL;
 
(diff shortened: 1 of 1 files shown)


src/ctrls.c
```diff
@@ -1937,7 +1937,9 @@ int pr_ctrls_reset(void) {
    *                             -1  processed, error (reset)
    */
 
-  for (ctrl = ctrls_active_list; ctrl; ctrl = ctrl->ctrls_next) {
+  for (ctrl = ctrls_active_list; ctrl; ctrl = next_ctrl) {
+    next_ctrl = ctrl->ctrls_next;
```
Castaglia, line 1941: Keep track of this ctrl's `ctrls_next` pointer _before_ calling `pr_ctrls_free()` on it, as that will clear/reset that pointer.

PR #2155, opened 2026-06-16
It is currently possible, although not actually observed, that excess…

Castaglia:
…ively long log messages could overflow the given log buffer.

The `pr_snprintf()` function returns -1 for such situations; the `log_write()` function needs to look for such return values, and act accordingly.

Thanks to Thanks to Kamil Leoniak and Delphos Labs for discovering and reporting this issue.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -526,11 +526,11 @@ static void log_write(int priority, int f, char *s, int discard) {
     remote_name = pr_netaddr_get_sess_remote_name();
 
     if (log_opts & PR_LOG_OPT_USE_VHOST) {
-      size_t len;
+      int len;
 
       len = pr_snprintf(serverinfo, sizeof(serverinfo)-1, "%s",
         main_server->ServerFQDN);
-      buflen += len;
+      buflen += (size_t) len;
     }
 
     serverinfo[sizeof(serverinfo)-1] = '\0';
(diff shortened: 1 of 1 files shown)


src/log.c
```diff
@@ -526,11 +526,11 @@ static void log_write(int priority, int f, char *s, int discard) {
     remote_name = pr_netaddr_get_sess_remote_name();
 
     if (log_opts & PR_LOG_OPT_USE_VHOST) {
-      size_t len;
+      int len;
 
       len = pr_snprintf(serverinfo, sizeof(serverinfo)-1, "%s",
         main_server->ServerFQDN);
-      buflen += len;
+      buflen += (size_t) len;
```
Castaglia, line 533: DNS hostnames cannot exceed 256 bytes (actually less), and at this time in the function, the `serverinfo` buffer is empty.  Thus no actual chance of overflowing the buffer, _i.e._ having `pr_snprintf()` return a negative value here.

src/log.c
```diff
@@ -583,16 +586,31 @@ static void log_write(int priority, int f, char *s, int discard) {
     }
 
     if (*serverinfo) {
+      int len;
+
       len = pr_snprintf(buf + buflen, sizeof(buf) - buflen,
         "%s[%u] %s: %s\n", process_label, (unsigned int) log_pid, serverinfo,
         s);
+      if (len < 0) {
```
Castaglia, line 594: These particular `pr_snprintf()` calls, where we are using the provided `s` text of the caller, are the most likely places for a potential overflow to occur, and thus it is here where we look for, and handle, a negative return value from `pr_snprintf()`.

Similarly for the other places in this `log_write()` function where this occurs.

PR #2157, opened 2026-06-18
Minor refactoring of some Redis API helper functions to reduce code d…

Castaglia:
…uplication.

### DIFF
--- a/src/redis.c
+++ b/src/redis.c
@@ -25,11 +25,11 @@
 
 #include "conf.h"
 
-#ifdef PR_USE_REDIS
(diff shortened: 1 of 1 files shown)


src/redis.c
```diff
@@ -25,19 +25,19 @@
 
 #include "conf.h"
 
-#ifdef PR_USE_REDIS
+#if defined(PR_USE_REDIS)
```
Castaglia, line 28: Stylistic nits; no functional change.  I lean toward being more verbose and explicit, rather than less.

PR #2162, opened 2026-06-23
It is possible for a client to say that it wishes to use the "sk-ssh-…

Castaglia:
…ed25519@openssh.com" algorithm for SSH user publickey authentication, but to _actually_ provide an "ssh-ed25519" public key.

The issue is that, when verifying the type of the public key provided by the client, we do not enforce that the provided key algorith matches the expected public key algorithm.


Castaglia:
This is mostly an issue for any sites that use:
```
SFTPAuthPublicKeys sk-ssh-ed25519@openssh.com
```
to require that _only_ this algorithm be used for user publickey authentication, for this issue allows the user to provide a (valid) "ssh-ed25519" key, without any of the touch/presence requirements that an _e.g._ FIDO security key might require.

PR #2168, opened 2026-06-23
When PCRE2 is enabled and used, it is possible that the given regular…

Castaglia:
… expression captures more groups than matches are provided by the caller.

### DIFF
--- a/src/regexp.c
+++ b/src/regexp.c
@@ -536,12 +536,22 @@ static int regexp_exec_pcre2(pr_regex_t *pre, const char *text,
     register uint32_t i;
     PCRE2_SIZE *ovector = NULL;
 
-    pr_trace_msg(trace_channel, 9,
-      "PCRE2 regex '%s' captured %lu groups in subject '%s'",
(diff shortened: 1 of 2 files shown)


src/regexp.c
```diff
@@ -536,12 +536,22 @@ static int regexp_exec_pcre2(pr_regex_t *pre, const char *text,
...
+        pr_regexp_get_pattern(pre), (unsigned long) ovector_count, text);
 
-    ovector = pcre2_get_ovector_pointer(match_data);
+    } else {
+      pr_trace_msg(trace_channel, 9,
+        "PCRE2 regex '%s' captured %lu groups in subject '%s', but %lu %s "
+        "provided", pr_regexp_get_pattern(pre),
+        (unsigned long) ovector_count, text, (unsigned long) nmatches,
+        nmatches != 1 ? "matches" : "match");
+      ovector_count = nmatches;
```
Castaglia, line 551: This is the fix: if there are more capture groups than `nmatches`, we clamp the value to `nmatches` for the subsequent iterator.

PR #2169, opened 2026-06-23
If the non-default (and undocumented) "xattr" SFTP extensions are ena…

Castaglia:
…bled, the "flistxattr" extension can return heap memory to the unsuspecting client.

### DIFF
--- a/contrib/mod_sftp/fxp.c
+++ b/contrib/mod_sftp/fxp.c
@@ -6351,8 +6351,6 @@ static int fxp_handle_ext_flistxattr(struct fxp_packet *fxp,
     sftp_msg_write_string(&buf, &buflen, name);
   }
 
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/fxp.c
```diff
@@ -6351,8 +6351,6 @@ static int fxp_handle_ext_flistxattr(struct fxp_packet *fxp,
     sftp_msg_write_string(&buf, &buflen, name);
   }
 
-  sftp_msg_write_data(&buf, &buflen, (const unsigned char *) names, res, TRUE);
```
Castaglia, line 6354: The "listxattr" and "llistxattr" extension implementations do The Right Thing(tm), and do not have this extra `sftp_msg_write_data()`; I suspect that its presence in the "flistxattr" implementation is an artifact of the original development time of this code, unintentionally left behind.

PR #2171, opened 2026-06-24
If the `NLST` command uses globs without recursion (`-R`), and one of…

Castaglia:
… the globbed paths is a directory, then any policy limits on that path are not being honored properly.

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -3242,7 +3242,9 @@ MODRET ls_nlst(cmd_rec *cmd) {
 
     /* Iterate through each matching entry */
     path = g.gl_pathv;
-    while (path && *path && res >= 0) {
+    while (path != NULL &&
+           *path &&
+           res >= 0) {
       struct stat st;
 
       pr_signals_handle();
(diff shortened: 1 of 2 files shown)


modules/mod_ls.c
```diff
@@ -3260,11 +3262,18 @@ MODRET ls_nlst(cmd_rec *cmd) {
 
           } else {
             /*...otherwise, just list the name. */
-            res = nlstfile(cmd, p);
+            if (ls_perms(cmd->tmp_pool, cmd, p, &hidden)) {
```
Castaglia, line 3265: The bug occurs because `ls_perms()` was not being called in this particular case.

modules/mod_ls.c
```diff
@@ -3260,11 +3262,18 @@ MODRET ls_nlst(cmd_rec *cmd) {
...
+                continue;
+              }
+
+              res = nlstfile(cmd, p);
+            }
           }
 
         } else if (S_ISREG(st.st_mode) &&
-            ls_perms(cmd->tmp_pool, cmd, p, &hidden)) {
+                   ls_perms(cmd->tmp_pool, cmd, p, &hidden)) {
```
Castaglia, line 3276: Stylistic nit; no functional change.

PR #2172, opened 2026-06-24
If the undocumented `UserDirRoot` directive is used for anonymous log…

Castaglia:
…ins, _and_ `UserAlias` is unwisely used with a wildcard matching, the constructed `chroot(2)` path for the login could be surprisingly different than expected.

Sanitize the client-provided `USER` name in such situations, to ensure that the constructed path is under the configured `<Anonymous>` chroot path.

### DIFF
--- a/doc/directives/index.html
+++ b/doc/directives/index.html
@@ -479,7 +479,6 @@ by the core <i>and</i> contrib modules in the ProFTPD source distribution.
   <li><a href="../modules/mod_xfer.html#UseSendfile">UseSendfile</a>
   <li><a href="../modules/mod_core.html#User">User</a>
   <li><a href="../modules/mod_auth.html#UserAlias">UserAlias</a>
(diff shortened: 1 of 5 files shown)


doc/directives/index.html
```diff
@@ -479,7 +479,6 @@
   <li><a href="../modules/mod_xfer.html#UseSendfile">UseSendfile</a>
   <li><a href="../modules/mod_core.html#User">User</a>
   <li><a href="../modules/mod_auth.html#UserAlias">UserAlias</a>
-  <li><a href="../modules/mod_auth.html#UserDirRoot">UserDirRoot</a>
```
Castaglia, line 482: This linked doc entry does not actually exist.  The `UserDirRoot` directive has been unofficially deprecated, downplayed, for years.  Perhaps now it's time to make it officially deprecated.

Issue #2173, opened 2026-06-24
RewriteMap unescape builtin use causes one-byte out-of-bounds write, fails to reject illegal characters


Castaglia:
Fabian Wahle of Hap Security disclosed the following issue for the `mod_rewrite` module.

**Summary**

When an admin configures:
```
<IfModule mod_rewrite.c>
  ...
  RewriteMap unescape int:unescape
  RewriteRule (.*) ${unescape:$1}
 ...
</IfModule>
```
and applies `${unescape:$1}` to client command arguments, the `rewrite_map_int_unescape()` allocates `strlen(key)` bytes and writes the NUL terminator at `value[i]`, causing a one-byte overflow for non-empty inputs.  The same code checks `key[i]` instead of decoded `value[i]`, allowing encoded `/` and NUL-style controls through the guard.

Castaglia:
Fixed in master, and backported to the 1.3.9 branch.

PR #2174, opened 2026-06-24
Issue #2173: Ensure that the `unescape` builtin `RewriteMap` allocate…

Castaglia:
…s enough memory to properly NUL-terminate the generated value without overflow.

In addition, make sure that we properly check the decoded characters, and reject unallowed `/`, NUL characters.

### DIFF
--- a/contrib/mod_rewrite.c
+++ b/contrib/mod_rewrite.c
@@ -1317,42 +1317,62 @@ static const char *rewrite_subst_maps(cmd_rec *cmd, const char *pattern) {
         /* Handle FIFO maps */
         if (strcmp(c->argv[1], "fifo") == 0) {
           lookup_value = rewrite_subst_maps_fifo(cmd, c, &map);
-          rewrite_log("rewrite_subst_maps(): fifo map '%s' returned '%s'",
-            map.map_name, lookup_value);
+          if (lookup_value != NULL) {
+            rewrite_log("rewrite_subst_maps(): fifo map '%s' returned '%s'",
+              map.map_name, lookup_value);
+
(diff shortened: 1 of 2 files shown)


contrib/mod_rewrite.c
```diff
@@ -1317,42 +1317,62 @@ static const char *rewrite_subst_maps(cmd_rec *cmd, const char *pattern) {
...
-          rewrite_log("rewrite_subst_maps(): fifo map '%s' returned '%s'",
-            map.map_name, lookup_value);
+          if (lookup_value != NULL) {
+            rewrite_log("rewrite_subst_maps(): fifo map '%s' returned '%s'",
+              map.map_name, lookup_value);
+
+          } else {
+            rewrite_log("rewrite_subst_maps(): fifo map '%s' returned NULL: %s",
+              map.map_name, strerror(errno));
+          }
```
Castaglia, line 1327: Most of these changes are to better handle the case where the map functions return a `NULL` value.

contrib/mod_rewrite.c
```diff
@@ -1889,7 +1909,7 @@ static const char *rewrite_map_int_unescape(pool *map_pool, char *key) {
   register int i, j;
   char *value;
 
-  value = pcalloc(map_pool, sizeof(char) * strlen(key));
+  value = pcalloc(map_pool, strlen(key) + 1);
```
Castaglia, line 1912: As `sizeof(char)` is largely redundant these days (_i.e._ it's almost always 1), just remove it.  We _do_ need to allocate one more byte, though, for the terminating NUL.

PR #2185, opened 2026-06-26
Address more memory leaks found while doing basic behavior tests with…

Castaglia:
… ASAN-instrumented builds.

### DIFF
--- a/contrib/mod_sftp/mod_sftp.c
+++ b/contrib/mod_sftp/mod_sftp.c
@@ -2358,7 +2358,7 @@ static void sftp_mod_unload_ev(const void *event_data, void *user_data) {
   destroy_pool(sftp_pool);
   sftp_pool = NULL;
 
-  close(sftp_logfd);
+  (void) close(sftp_logfd);
   sftp_logfd = -1;
 }
 #endif /* PR_SHARED_MODULE */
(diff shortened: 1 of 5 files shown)


contrib/mod_sftp/mod_sftp.c
```diff
@@ -2515,6 +2515,13 @@ static void sftp_restart_ev(const void *event_data, void *user_data) {
 
   /* Clear the client banner regexes. */
   sftp_interop_free();
+
+#if defined(HAVE_OSSL_PROVIDER_LOAD_OPENSSL)
+  if (legacy_provider != NULL) {
```
Castaglia, line 2520: This provider is loaded during the "postparse" event handling, which happens after a restart anyway.  Failing to unload it here leaks memory.

contrib/mod_shaper.c
```diff
@@ -2071,13 +2071,19 @@ MODRET shaper_pre_pass(cmd_rec *cmd) {
    * in the daemon process.
    */
 
-  PRIVS_ROOT
-  shaper_tabfd = open(shaper_tab_path, O_RDWR);
```
Castaglia, line 2075: If `shaper_tab_path` is NULL (as it will be with, with a configuration that does not enable/configure the `mod_shaper` module), then `open(2)` is called with a NULL argument, which rightly generates a runtime warning about calling it with an illegal (NULL) value.

contrib/mod_tls.c
```diff
@@ -15925,6 +15925,11 @@ static void tls_restart_ev(const void *event_data, void *user_data) {
   }
 #endif /* PR_USE_CTRLS */
 
+  if (ssl_ctx != NULL) {
```
Castaglia, line 15928: This `ssl_ctx` is allocated during the "postparse" event handling, which happens as part of a restart; failing to free it here leaks memory.

PR #2196, opened 2026-06-30
Addressing issues reported by Coverity.

Castaglia:


contrib/mod_auth_otp/otp.c
```diff
@@ -43,7 +43,7 @@ static int otp(pool *p, const EVP_MD *md,
     value[i] = counter;
   }
 
-  hash_len = EVP_MAX_MD_SIZE;
+  hash_len = sizeof(hash);
```
Castaglia, line 46: Stylistic nit; no functional change.

contrib/mod_sftp/channel.c
```diff
@@ -469,17 +471,16 @@ static struct ssh2_channel_databuf *get_databuf(uint32_t channel_id,
 static struct ssh2_channel_databuf *get_incoming_databuf(uint32_t channel_id,
     uint32_t buflen) {
   struct ssh2_channel_databuf *db;
-  struct ssh2_channel *chan;
+  struct ssh2_channel *chan = NULL;
 
-  db = get_databuf(channel_id, buflen);
+  db = get_databuf(channel_id, buflen, &chan);
```
Castaglia, line 476: Rather than doing a (redundant) `get_channel()` call later, we now have `get_databuf()` return that same channel itself.  Makes things less confusing for tools like Coverity.

contrib/mod_sftp/fxp.c
```diff
@@ -11973,8 +11975,6 @@ static int fxp_handle_realpath(struct fxp_packet *fxp) {
       uint32_t status_code;
       const char *reason;
 
-      xerrno = errno;
```
Castaglia, line 11976: We set `xerrno` to the appropriate value above; no need to do it here.

contrib/mod_sftp/keys.c
```diff
@@ -3246,7 +3246,9 @@ static int decrypt_openssh_data(pool *p, const char *path,
   /* Compute the decryption key using the KDF and the passphrase.  Note that
    * we derive the key AND the IV using this approach at the same time.
    */
-  passphrase_len = strlen(passphrase);
+  if (passphrase != NULL) {
+    passphrase_len = strlen(passphrase);
+  }
```
Castaglia, line 3251: Coverity rightly pointed out a case where `passphrase` could possibly be NULL; we need to handle that properly here.

contrib/mod_sftp/keys.c
```diff
@@ -4259,6 +4261,7 @@ static int load_file_hostkey(pool *p, const char *path) {
   }
 
   free_hostkey_bio(bio);
+  (void) close(fd);
```
Castaglia, line 4264: Fix a resource leak (ish) found by Coverity.

Loading of mod_sftp `SFTPHostKey` files happens on startup, and during restarts.  So over the (long) lifetime of the daemon, such an fd leak might be noticed.

contrib/mod_sftp/packet.c
```diff
@@ -552,7 +552,7 @@ static void read_packet_discard(int sockfd) {
      * to be read in.
      */
     flags = SFTP_PACKET_READ_FL_PESSIMISTIC;
-    sftp_ssh2_packet_sock_read(sockfd, buf, buflen, flags);
+    (void) sftp_ssh2_packet_sock_read(sockfd, buf, buflen, flags);
```
Castaglia, line 555: Try to indicate to Coverity (and to future readers/maintainers) that we _really_ don't care about the return value here, since we're just reading a random number of bytes from the socket before we close things anyway.  (Why?  So that we don't behave like an oracle in such error situations.)

contrib/mod_sftp/scp.c
```diff
@@ -2629,6 +2629,14 @@ int sftp_scp_set_params(pool *p, uint32_t channel_id, array_header *req) {
     }
   }
 
+  /* If we have no target path at this point, it's an error. */
+  if (target_path == NULL) {
+    (void) pr_log_writefile(sftp_logfd, MOD_SFTP_VERSION,
+      "'scp' request provided no target path, rejecting");
+    errno = EPERM;
+    return -1;
+  }
```
Castaglia, line 2638: It's unlikely that `target_path` _would_ be null in normal SCP usage, but we should be prepared for ill-behaved SCP clients anyway.

contrib/mod_radius.c
```diff
@@ -1254,7 +1254,7 @@ static int radius_process_quota_info_attribs(radius_packet_t *pkt) {
 
 static int radius_process_accept_packet(radius_packet_t *pkt,
     const unsigned char *secret, size_t secret_len) {
-  int attrib_count = 0, res;;
+  int attrib_count = 0, res;
```
Castaglia, line 1257: Typo cleanup; no functional change.

contrib/mod_radius.c
```diff
@@ -3084,9 +3084,10 @@ static unsigned int radius_get_terminate_cause(void) {
 }
 
 static int radius_stop_accting(void) {
-  int sockfd = -1, acct_status = 0, acct_authentic = 0, event_ts = 0,
-    now = 0, pid_len = 0, session_duration = 0;
-  unsigned int terminate_cause = 0;
+  int sockfd = -1, acct_status = 0, acct_authentic = 0;
+  size_t pid_len = 0;
+  time_t now = 0, event_ts = 0;
+  unsigned int terminate_cause = 0, session_duration = 0;
```
Castaglia, line 3090: Try to use more appropriate data types, especially `time_t` for time-related values.

contrib/mod_tls.c
```diff
@@ -5948,6 +5948,10 @@ static OCSP_RESPONSE *ocsp_request_response(pool *p, X509 *cert, SSL *ssl,
     OPENSSL_free(port);
     OPENSSL_free(uri);
 
+    if (ctx != NULL) {
+      SSL_CTX_free(ctx);
+    }
```
Castaglia, line 5953: Address possible resource leak in this error case if we have to use HTTPS to obtain OCSP responses.

contrib/mod_tls.c
```diff
@@ -10843,15 +10847,6 @@ static int tls_verify_crl(int ok, X509_STORE_CTX *ctx) {
...
 #endif /* OpenSSL 1.1.x and later */
 
-      if (res == 0) {
-        tls_log("CRL has invalid lastUpdate field: %s", tls_get_errors());
-
-        X509_STORE_CTX_set_error(ctx,
-          X509_V_ERR_ERROR_IN_CRL_LAST_UPDATE_FIELD);
-        sk_X509_CRL_free(crls);
-        return FALSE;
-      }
```
Castaglia, line 10853: Delete copy/pasta.

contrib/mod_tls.c
```diff
@@ -14545,6 +14540,7 @@ MODRET set_tlsecdhcurve(cmd_rec *cmd) {
 
     res = SSL_CTX_set1_curves_list(ctx, curve_names);
     if (res != 1) {
+      SSL_CTX_free(ctx);
```
Castaglia, line 14543: Minor resource leak in the error case; makes Coverity happier to have this addressed.

src/ascii.c
```diff
@@ -61,15 +61,16 @@ int pr_ascii_ftp_from_crlf(pool *p, char *in, size_t inlen, char **out,
         adj++;
         *dst++ = *src++;
 
+        break;
```
Castaglia, line 64: If `rem == 0`, then we do _not_ want to continue the `while` loop.  So make that explicitly clear here.  No functional change.

src/ctrls.c
```diff
@@ -943,14 +943,6 @@ int pr_ctrls_recv_response(pool *p, int fd, int *status, char ***respargv) {
   if (nread != sizeof(uint32_t)) {
     pr_signals_unblock();
 
-    if (nread < 0) {
```
Castaglia, line 946: The `nread < 0` case is already handled on line 931 above, so we can remove this redundant code.

src/inet.c
```diff
@@ -1177,6 +1177,11 @@ int pr_inet_set_reuse_port(pool *p, conn_t *c, int reuse_port) {
     return -1;
   }
 
+  if (c->listen_fd < 0) {
+    errno = EBADF;
+    return -1;
+  }
```
Castaglia, line 1183: The `setsockopt(2)` system call ideally doesn't take negative values, so be defensive and check for such things here.  Makes Coverity happier.

utils/ftptop.c
```diff
@@ -258,35 +258,35 @@ static int check_scoreboard_file(void) {
 static const char *show_ftpd_uptime(void) {
   static char buf[128] = {'\0'};
   time_t uptime_secs = time(NULL) - ftp_uptime;
-  int upminutes, uphours, updays;
+  unsigned long upminutes, uphours, updays;
```
Castaglia, line 261: These variables hold time-related values, so use a larger data type for them.

utils/ftpwho.c
```diff
@@ -276,30 +276,30 @@ static int check_scoreboard_file(void) {
 static const char *show_uptime(time_t uptime_since) {
   static char buf[128] = {'\0'};
   time_t uptime_secs = time(NULL) - uptime_since;
-  int upminutes, uphours, updays;
+  unsigned long upminutes, uphours, updays;
```
Castaglia, line 279: These variables hold time-related values, so use a larger data type for them.

PR #2204, opened 2026-07-02
Make sure that an SFTP `REALPATH` request uses one of the legal/expec…

Castaglia:
…ted flag values, and reject any others.


contrib/mod_sftp/fxp.c
```diff
@@ -12133,6 +12133,30 @@ static int fxp_handle_realpath(struct fxp_packet *fxp) {
   fxb->bufsz = buflen = FXP_RESPONSE_NAME_DEFAULT_SZ;
   fxb->ptr = buf = palloc(fxp->pool, fxb->bufsz);
 
+  if (realpath_flags != SSH2_FXRP_NO_CHECK &&
```
Castaglia, line 12136: Coverity complained of use of unitialized `st` in this funtion, which could _only_ happen if `realpath_flags` was not a value listed in a `switch` statement.  Address that valid complaint properly by adding validating of the `realpath_flags` value, to ensure it can only be one of those listed values.

PR #2215, opened 2026-07-08
Implement stricter sanity checks of the client-provided filesystem pa…

Castaglia:
…th in the `pr_ctrls_accept()` function, such that arbitrary paths will be rejected.

### DIFF
--- a/src/ctrls.c
+++ b/src/ctrls.c
@@ -42,6 +42,8 @@
 
 #include "mod_ctrls.h"
 
+#define CTRLS_SOCKET_PREFIX	"/tmp/ftp.cl"
+
 #define CTRLS_MAX_REQ_SIZE	1024
 #define CTRLS_MAX_RESP_SIZE	(1024 * 1024)
 
@@ -1332,7 +1334,7 @@ int pr_ctrls_connect(const char *socket_file) {
 
   cl_sock.sun_family = AF_UNIX;
   pr_snprintf(cl_sock.sun_path, sizeof(cl_sock.sun_path) - 1, "%s%05u",
-    "/tmp/ftp.cl", (unsigned int) getpid());
+    CTRLS_SOCKET_PREFIX, (unsigned int) getpid());
   len = sizeof(cl_sock);
 
   /* Make sure the file doesn't already exist.  If it does exist AND is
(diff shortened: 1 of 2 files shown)


src/ctrls.c
```diff
@@ -1648,13 +1650,14 @@ static int ctrls_get_creds_local(int fd, uid_t *uid, gid_t *gid,
...
-  char *tmp = NULL;
+  char *tmp = NULL, *endp = NULL;
   time_t stale_time;
   struct stat st;
+  unsigned long file_pid = 0;
 
   /* Check the path -- hmmm... */
   PRIVS_ROOT
-  while (stat(sock->sun_path, &st) < 0) {
+  while (lstat(sock->sun_path, &st) < 0) {
```
Castaglia, line 1660: Here, we check for possible symlinks.

src/ctrls.c
```diff
@@ -1665,7 +1668,6 @@ static int ctrls_get_creds_basic(struct sockaddr_un *sock, int cl_fd,
     PRIVS_RELINQUISH
     pr_trace_msg(trace_channel, 2, "error: unable to stat %s: %s",
       sock->sun_path, strerror(xerrno));
-    (void) close(cl_fd);
```
Castaglia, line 1668: Let the caller (which opened this `cl_fd`) close it on error, not us.

src/ctrls.c
```diff
@@ -1734,29 +1734,30 @@ static int ctrls_get_creds_basic(struct sockaddr_un *sock, int cl_fd,
...
-  }
+  cl_pid = (pid_t) file_pid;
 
-  if (pid != NULL) {
-    *pid = cl_pid;
-  }
+  /* Return the IDs of the caller */
+  *uid = st.st_uid;
+  *gid = st.st_gid;
+  *pid = cl_pid;
```
Castaglia, line 1760: We know (now) that the only caller of this function provides these pointers, so no need for the NULL pointer checks.

src/ctrls.c
```diff
@@ -1791,6 +1795,33 @@ int pr_ctrls_accept(int fd, uid_t *uid, gid_t *gid, pid_t *pid,
...
   sock.sun_path[sizeof(sock.sun_path)-1] = '\0';
 
+  if (strncmp(sock.sun_path, CTRLS_SOCKET_PREFIX,
+      strlen(CTRLS_SOCKET_PREFIX)) != 0) {
+    pr_trace_msg(trace_channel, 6, "refusing unexpected client path '%s'",
+      sock.sun_path);
+    (void) close(cl_fd);
+    errno = EINVAL;
+    return -1;
+  }
```
Castaglia, line 1805: This lets us check for, and avoid, using any Unix socket address path that does not start with the expected `/tmp/ftp.cl`.

PR #2219, opened 2026-07-09
ASAN detected a few memory leaks in the `mod_ldap` module; address them.

Castaglia:


contrib/mod_ldap.c
```diff
@@ -65,28 +65,28 @@ static const char *libtrace_channel = "ldap.library";
...
 static int ldap_port = LDAP_PORT;
-#endif
+#endif /* HAS_LDAP_UNBIND_EXT_S */
 
 /* On some systems LDAP_OPT_DIAGNOSTIC_MESSAGE isn't there (e.g. OpenLDAP-2.3.x)
  * but LDAP_OPT_ERROR_STRING is.
  */
-#ifndef LDAP_OPT_DIAGNOSTIC_MESSAGE
-# ifdef LDAP_OPT_ERROR_STRING
+#if !defined(LDAP_OPT_DIAGNOSTIC_MESSAGE)
```
Castaglia, line 85: Stylistic nits; no functional change.

contrib/mod_ldap.c
```diff
@@ -1144,7 +1146,11 @@ static struct passwd *pr_ldap_user_lookup(pool *p, char *filter_template,
   /* If we're doing auth binds, save the DN of this entry so we can
    * bind to the LDAP server as it later.
    */
-  if (user_dn) {
+  if (user_dn != NULL) {
+    if (*user_dn != NULL) {
+      ldap_memfree(*user_dn);
+    }
```
Castaglia, line 1152: Previously, we would _always_ overwrite any existing `*user_dn` value with a new value, leaking small bits of memory.  Now, last one wins.

contrib/mod_ldap.c
```diff
@@ -2193,6 +2199,7 @@ static void server_info_get_ssl_defaults(struct server_info *info) {
         pr_trace_msg(trace_channel, 17,
           "using default 'ssl-ca' value: %s", ssl_val);
         info->ssl_ca_file = ldap_strdup(ssl_val);
+        ldap_memfree(ssl_val);
```
Castaglia, line 2202: Since we already make a copy of the returned value for ourselves via `ldap_strdup`, we should free the value returned by by `ldap_get_option` here and not leak the memory.

PR #2222, opened 2026-07-10
Be a little more careful when examining passwords in the LDAP "AuthBi…

Castaglia:
…nds off" scenario, as the passwords could be an empty string, or not have expected "{scheme}" prefaces.

### DIFF
--- a/contrib/mod_ldap.c
+++ b/contrib/mod_ldap.c
@@ -1869,7 +1869,8 @@ MODRET ldap_auth_auth(cmd_rec *cmd) {
  */
 MODRET ldap_auth_check(cmd_rec *cmd) {
   char *pass, *cryptpass, *hash_method, *crypted;
-  int encname_len, res;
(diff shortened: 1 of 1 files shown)


contrib/mod_ldap.c
```diff
@@ -1937,20 +1938,31 @@ MODRET ldap_auth_check(cmd_rec *cmd) {
...
   }
 
-  /* Get the length of "scheme" in the leading {scheme} so we can skip it
-   * in the password comparison.
+  /* Check to see how the password is encrypted, and check accordingly.
+   *
+   * Get the length of "scheme" in the potential leading {scheme} so we can
+   * skip it in the password comparison.
    */
-  encname_len = strcspn(cryptpass + 1, "}");
```
Castaglia, line 1943: In the case where `cryptpass` is the `""` empty string, this previous use of `cryptpass+1` is a one-byte OOB read.

PR #2223, opened 2026-07-10
Handle edge cases where the UID/GID attribute values obtained from LD…

Castaglia:
…AP are not properly numeric.

### DIFF
--- a/contrib/mod_ldap.c
+++ b/contrib/mod_ldap.c
@@ -1058,7 +1058,23 @@ static struct passwd *pr_ldap_user_lookup(pool *p, char *filter_template,
         }
 
       } else {
-        pw->pw_uid = (uid_t) strtoul(LDAP_VALUE(values, 0), NULL, 10);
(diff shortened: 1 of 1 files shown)


contrib/mod_ldap.c
```diff
@@ -1058,7 +1058,23 @@ static struct passwd *pr_ldap_user_lookup(pool *p, char *filter_template,
         }
 
       } else {
-        pw->pw_uid = (uid_t) strtoul(LDAP_VALUE(values, 0), NULL, 10);
+        const char *uid_text;
+        uid_t uid;
+
+        uid_text = LDAP_VALUE(values, 0);
+        if (pr_str2uid(uid_text, &uid) == 0) {
```
Castaglia, line 1065: At the time this mod_ldap code was written, this `pr_str2uid()` function did not exist.  Thus this is more of a backporting of this new functionality to this older code path.

PR #2224, opened 2026-07-10
Handle a potential use-after-free scenario when a group lookup is mis…

Castaglia:
…sing required attributes.

### DIFF
--- a/contrib/mod_ldap.c
+++ b/contrib/mod_ldap.c
@@ -869,7 +869,7 @@ static struct passwd *pr_ldap_user_lookup(pool *p, char *filter_template,
           (void) pr_log_writefile(ldap_logfd, MOD_LDAP_VERSION,
             "no %s attribute for DN %s found, and LDAPDefaultUID not "
             "configured", ldap_attr_uidnumber, dn);
-          free(dn);
+          ldap_memfree(dn);
           return NULL;
         }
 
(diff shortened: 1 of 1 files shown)


contrib/mod_ldap.c
```diff
@@ -869,7 +869,7 @@ static struct passwd *pr_ldap_user_lookup(pool *p, char *filter_template,
           (void) pr_log_writefile(ldap_logfd, MOD_LDAP_VERSION,
             "no %s attribute for DN %s found, and LDAPDefaultUID not "
             "configured", ldap_attr_uidnumber, dn);
-          free(dn);
+          ldap_memfree(dn);
```
Castaglia, line 872: Let's be consistent, and use `ldap_memfree` as prescribed by the `ldap_get_dn` docs.

contrib/mod_ldap.c
```diff
@@ -1261,13 +1261,13 @@ static struct group *pr_ldap_group_lookup(pool *p, char *filter_template,
         continue;
       }
 
-      ldap_msgfree(result);
```
Castaglia, line 1264: This is the potential use-after-free site: `ldap_msgfree()` is called _before_ we obtain the DN from the (now-freed) result for logging.  So instead, we free the message _after_ we've logged what we want.  Easy.

PR #2227, opened 2026-07-11
Implement stricter sanity checks when resolving numeric/index referen…

Castaglia:
…ces in SQLNamedQuery statements.

### DIFF
--- a/tests/t/lib/ProFTPD/Tests/Modules/mod_sql_sqlite.pm
+++ b/tests/t/lib/ProFTPD/Tests/Modules/mod_sql_sqlite.pm
@@ -477,6 +477,17 @@ my $TESTS = {
     order => ++$order,
     test_class => [qw(forking bug)],
   },
+
+  sql_sqllog_freeform_ignore_numeric_tag => {
+    order => ++$order,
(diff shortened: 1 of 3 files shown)


tests/t/lib/ProFTPD/Tests/Modules/mod_sql_sqlite.pm
```diff
@@ -16108,4 +16085,356 @@ EOS
...
+
+      'mod_sql.c' => {
+        SQLEngine => 'log',
+        SQLBackend => 'sqlite3',
+        SQLConnectInfo => $db_file,
+        SQLLogFile => $setup->{log_file},
+
+        # Yes, this "rename_from" column is not accurate.  I'm lazy, and
+        # am reusing existing code for this test.
+        SQLNamedQuery => 'log_download INSERT "\'%u\', \'%L\', \'%{999999}\'" ftpsessions',
```
Castaglia, line 16341: Triggering this bug requires that the site configuration have such an out-of-bounds reference.  Which amounts to a "shooting yourself in the foot" kind of situation; this is not reachable remotely.

PR #2229, opened 2026-07-11
Clean up a few ratio-related code paths that did not properly handle …

Castaglia:
…cases where the database was unavailable, and "SQLOptions noDisconnectOnError" was in effect.

### DIFF
--- a/contrib/mod_ratio.c
+++ b/contrib/mod_ratio.c
@@ -169,8 +169,6 @@ _dispatch (cmd_rec * cmd, char *match)
 
   cr = _make_cmd (cmd->tmp_pool, 0);
   mr = _dispatch_ratio (cr, match);
-  if (cr->tmp_pool)
-    destroy_pool (cr->tmp_pool);
   return mr;
 }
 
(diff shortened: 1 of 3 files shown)


contrib/mod_ratio.c
```diff
@@ -169,8 +169,6 @@ _dispatch (cmd_rec * cmd, char *match)
 
   cr = _make_cmd (cmd->tmp_pool, 0);
   mr = _dispatch_ratio (cr, match);
-  if (cr->tmp_pool)
-    destroy_pool (cr->tmp_pool);
```
Castaglia, line 173: ASAN properly complained that destroying the pool here leads to a use-after-free, since the returned `mr` is allocated out of the pool being destroyed.

As this function uses the provided `cmd->tmp_pool` anyway, that `tmp_pool` is cleared up/destroyed by the core code, so there's no need to do the destroy here.

contrib/mod_sql.c
```diff
@@ -4607,17 +4607,23 @@ MODRET sql_auth_getgroups(cmd_rec *cmd) {
...
 
   if (!cmap.sql_fstor) {
     return PR_DECLINED(cmd);
   }
 
-  usrwhere = pstrcat(cmd->tmp_pool, cmap.usrfield, " = '", sql_realuser(cmd),
+  real_user = sql_realuser(cmd);
+  if (real_user == NULL) {
+    sql_log(DEBUG_FUNC, "%s", "<<< cmd_getstats");
+    return PR_ERROR_MSG(cmd, MOD_SQL_VERSION, "database error");
```
Castaglia, line 4623: If we can't retrieve the real user here, there's no point in continuing on.

PR #2230, opened 2026-07-11
Ensure that any generated home directory for LDAP users is sanitized,…

Castaglia:
… and that that path still falls within the configured `LDAPGenerateHomedirPrefix`.

### DIFF
--- a/contrib/mod_ldap.c
+++ b/contrib/mod_ldap.c
@@ -1171,6 +1171,9 @@ static struct passwd *pr_ldap_user_lookup(pool *p, char *filter_template,
 
         } else {
           LDAP_VALUE_T **canon_username;
+          char *home_dir, sanitized_home_dir[PR_TUNABLE_PATH_MAX+1];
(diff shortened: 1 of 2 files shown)


contrib/mod_ldap.c
```diff
@@ -1182,9 +1185,43 @@ static struct passwd *pr_ldap_user_lookup(pool *p, char *filter_template,
             return NULL;
           }
 
-          pw->pw_dir = pstrcat(session.pool, ldap_genhdir_prefix, "/",
+          home_dir = pstrcat(p, ldap_genhdir_prefix, "/",
             LDAP_VALUE(canon_username, 0), NULL);
           LDAP_VALUE_FREE(canon_username);
```
Castaglia, line 1190: The issue here is that the "canonical username" provided by the LDAP server _might_ contain path characters, such as "../../".  We thus sanitize/canonicalize the generated path, and then ensure it is within/under the `LDAPGenerateHomedirPrefix` directory.

PR #2234, opened 2026-07-14
When working with the `ftpQuotaProfileDN` values for quota limit limi…

Castaglia:
…t lookups, make sure we handle recursion cases with appropriate limits.

### DIFF
--- a/contrib/mod_ldap.c
+++ b/contrib/mod_ldap.c
@@ -51,6 +51,7 @@ static int ldap_logfd = -1;
 static pool *ldap_pool = NULL;
 
 static const char *trace_channel = "ldap";
+
 #if defined(LBER_OPT_LOG_PRINT_FN)
 static const char *libtrace_channel = "ldap.library";
 #endif /* LBER_OPT_LOG_PRINT_FN */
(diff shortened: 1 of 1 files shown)


contrib/mod_ldap.c
```diff
@@ -906,6 +909,9 @@ static LDAPMessage *pr_ldap_search(const char *basedn, const char *filter,
       if (ctrls[0] != NULL) {
         ldap_control_free(ctrls[0]);
       }
+      if (result != NULL) {
+        ldap_msgfree(result);
+      }
```
Castaglia, line 914: These defensive checks help address (minor) memory leaks, as found by ASAN builds.

contrib/mod_ldap.c
```diff
@@ -1644,11 +1664,22 @@ static array_header *pr_ldap_quota_lookup(pool *p, char *filter_template,
...
     orig_scope = ldap_search_scope;
     ldap_search_scope = LDAP_SCOPE_BASE;
-    quota = pr_ldap_quota_lookup(p, NULL, replace, LDAP_VALUE(values, 0));
+
+    new_basedn = LDAP_VALUE(values, 0);
+    pr_trace_msg(trace_channel, 19,
+      "following %s attribute value '%s' (depth %u)",
+      ldap_attr_ftpquota_profiledn, new_basedn, depth);
+
+    quota = pr_ldap_quota_lookup(p, filter_template, replace, new_basedn,
```
Castaglia: The presence/use of custom `ftpQuotaProfileDN` attributes is unusual, but supported in the code, so we need to handle them (as opposed to remove support for such possible recursion).

Castaglia: Hmm.  Should we pass in our provided `filter_template` to this recursion call, or not?  The original/existing code does not, on the _assumption_, I presume, that the obtained DN needs no filtering.  Providing that filter template is s change of existing behavior; perhaps I should change it back.

PR #2236, opened 2026-07-14
More iteration around the support and handling of possibly-recursive …

Castaglia:
…`ftpQuotaProfileDN` attributes.

### DIFF
--- a/contrib/mod_ldap.c
+++ b/contrib/mod_ldap.c
@@ -1677,7 +1677,16 @@ static array_header *pr_ldap_quota_lookup(pool *p, char *filter_template,
       "following %s attribute value '%s' (depth %u)",
       ldap_attr_ftpquota_profiledn, new_basedn, depth);
 
-    quota = pr_ldap_quota_lookup(p, NULL, replace, new_basedn, ++depth);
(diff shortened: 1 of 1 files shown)


contrib/mod_ldap.c
```diff
@@ -1677,7 +1677,16 @@ static array_header *pr_ldap_quota_lookup(pool *p, char *filter_template,
...
-    quota = pr_ldap_quota_lookup(p, NULL, replace, new_basedn, ++depth);
+    /* It is possible that, recursively, the DN found points to an entry that
+     * itself has a DN reference.  In order to support such cases, we do need
+     * to provide the same filter template as we were given.
+     *
+     * NOTE: If folks later want to be able to customize that filter template,
+     * we can re-create an "LDAPQuotas" directive for providing such things.
+     * But not unless and until we have a concrete use case for it.
+     */
+    quota = pr_ldap_quota_lookup(p, filter_template, replace, new_basedn,
```
Castaglia, line 1688: The passing of a filter template is necessary for supporting such recursion, due to the check on line 1650.  If there is no `filter`, and no `ftpQuota` attribute, then we will not look further via recursion.

PR #2239, opened 2026-07-17
When populating the names of the quotatab limit/tally tables, use our…

Castaglia:
… own `sstrncpy()` function.


contrib/mod_quotatab.c
```diff
@@ -1012,7 +1012,7 @@ unsigned char quotatab_lookup_default(quota_tabtype_t tab_type, void *ptr,
      *  files_{in,out,xfer}_avail
      */
 
-    memmove(limit->name, name, strlen(name) + 1);
+    sstrncpy(limit->name, name, sizeof(limit->name)-1);
```
Castaglia, line 1015: Using `sstrncpy` ensures that we respect the size of the `name` field.

Issue #2242, opened 2026-07-20
SSH channel open request from authenticated client with max packet size of zero leads to infinite loop


afldl:
# ProFTPD mod_sftp `max_packet_size=0` channel-send infinite loop (server-side)

## Identity
- **Target:** ProFTPD 1.3.9b with `mod_sftp` (`contrib/mod_sftp/channel.c`)
- **Vulnerable sites:**
  - `contrib/mod_sftp/channel.c:120` — `chan->remote_max_packetsz = remote_max_packetsz;` stores the peer's value **raw, no `>= 1` floor**
  - `contrib/mod_sftp/channel.c:355-356` — send-burst clamp `if (payload_len > chan->remote_max_packetsz) payload_len = chan->remote_max_packetsz;`
  - send loop `~335-410`: `buflen -= payload_len; remote_windowsz -= payload_len; outgoing_len -= payload_len;` — all `-= 0` when `payload_len == 0`
- **Class:** CWE-835 (infinite loop) — the SSH `max_packet_size=0` cluster (RFC 4254 §5.1)
- **Reachability:** post-auth; a malicious SSH **client** advertises `max_packet_size=0` in
  `SSH_MSG_CHANNEL_OPEN`; the server wedges when it next sends channel data (e.g. an SFTP response)
- **Impact:** server DoS — one mod_sftp child process pinned at ~100% CPU / unbounded output buffer

## Verification level
**Source-confirmed** against ProFTPD 1.3.9b (the exact code chain below). A live E2E
was attempted: ProFTPD 1.3.9b + mod_sftp was built, but on this build host mod_sftp compiled to
`contrib/mod_sftp/mod_sftp.a` yet did not link into the `proftpd` binary (the `-l` compiled-module
list stayed empty) — a build-environment quirk. The `malicious_client.py` + `build-and-run.sh`
complete the live path once mod_sftp links cleanly. The vulnerability stands on source evidence
(the same evidence the rest of the maxpkt0 cluster was verified on).

## Root cause
```c
// contrib/mod_sftp/channel.c  (channel write/send loop)
while (db != NULL && db->buflen > 0 && chan->remote_windowsz > 0) {
    payload_len = db->buflen;
    if (payload_len > chan->remote_max_packetsz)          // line ~355
        payload_len = chan->remote_max_packetsz;          //   -> 0 when peer sends 0
    if (payload_len > chan->remote_windowsz)
        payload_len = chan->remote_windowsz;
[...]

PR #2243, opened 2026-07-20
Issue #2242: Reject SSH `CHANNEL_OPEN` requests with max packet size …

Castaglia:
…of zero.

### DIFF
--- a/contrib/mod_sftp/channel.c
+++ b/contrib/mod_sftp/channel.c
@@ -535,6 +535,7 @@ static int read_channel_open(struct ssh2_packet *pkt, uint32_t *channel_id) {
     (void) pr_log_writefile(sftp_logfd, MOD_SFTP_VERSION,
       "maximum number of channels (%u) open, denying request to "
       "open '%s' channel", channel_count, channel_type);
+    errno = EPERM;
     return -1;
   }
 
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/channel.c
```diff
@@ -1502,17 +1533,35 @@ uint32_t sftp_channel_set_max_windowsz(uint32_t windowsz) {
...
+  int res, xerrno;
   uint32_t channel_id;
 
   switch (msg_type) {
     case SFTP_SSH2_MSG_CHANNEL_OPEN: {
       res = read_channel_open(pkt, &channel_id);
+      xerrno = errno;
+
       if (res < 0) {
+        int reason_code = SFTP_SSH2_CHANNEL_OPEN_UNKNOWN_CHANNEL_TYPE;
```
Castaglia: Providing better-matched reasons for rejecting bad channel requests makes for (slightly) less surprising error messages to the end user of the remote client.

contrib/mod_sftp/channel.c
```diff
@@ -1436,25 +1450,40 @@ static int write_channel_open_confirm(struct ssh2_packet *pkt,
   return 0;
 }
 
-static int write_channel_open_failed(struct ssh2_packet *pkt,
-    uint32_t channel_id) {
+static void write_channel_open_failed(struct ssh2_packet *pkt,
```
Castaglia, line 1453: Since we (currently) only ever return 0 (success) for this function, let's just make it `void` for now.  More legible for future human readers that way.

PR #2245, opened 2026-07-21
The handlers for the SFTP `hardlink@openssh.com` and `LINK` requests …

Castaglia:
…were not dispatching a cmd, for use by any/all modules as necessary, as is done by other handlers.

### DIFF
--- a/contrib/mod_sftp/fxp.c
+++ b/contrib/mod_sftp/fxp.c
@@ -5453,6 +5453,29 @@ static int fxp_handle_ext_hardlink(struct fxp_packet *fxp, char *src,
   buflen = bufsz = FXP_RESPONSE_DATA_DEFAULT_SZ;
   buf = ptr = palloc(fxp->pool, bufsz);
 
+  if (pr_cmd_dispatch_phase(cmd, PRE_CMD, 0) < 0) {
+    status_code = SSH2_FX_PERMISSION_DENIED;
+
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/fxp.c
```diff
@@ -9302,6 +9325,31 @@ static int fxp_handle_link(struct fxp_packet *fxp) {
   buflen = bufsz = FXP_RESPONSE_DATA_DEFAULT_SZ;
   buf = ptr = palloc(fxp->pool, bufsz);
 
+  if (is_symlink == FALSE) {
```
Castaglia, line 9328: If `is_symlink` is true, we'll handle this as a symlink request, which should be already covered.  Right?

contrib/mod_sftp/fxp.c
```diff
@@ -14170,7 +14218,7 @@ static int fxp_handle_symlink(struct fxp_packet *fxp) {
     status_code = SSH2_FX_PERMISSION_DENIED;
 
     (void) pr_log_writefile(sftp_logfd, MOD_SFTP_VERSION,
-      "SYMLINK of '%s' to '%s' blocked by '%s' handler: %sl", target_path,
+      "SYMLINK of '%s' to '%s' blocked by '%s' handler: %s", target_path,
```
Castaglia, line 14221: Fixing typo.

PR #2246, opened 2026-07-21
Running the mod_redis integration/regression tests with an ASAN build…

Castaglia:
… revealed two minor use-after-free bugs (but only with specialized developer builds).

### DIFF
--- a/modules/mod_redis.c
+++ b/modules/mod_redis.c
@@ -688,12 +688,12 @@ MODRET set_redislogoncommand(cmd_rec *cmd) {
         key_text, "': ", strerror(errno), NULL));
     }
 
-    destroy_pool(tmp_pool);
-
     key_fmtlen = jot_parsed->bufsz - jot_parsed->buflen;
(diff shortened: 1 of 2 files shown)


modules/mod_redis.c
```diff
@@ -688,12 +688,12 @@ MODRET set_redislogoncommand(cmd_rec *cmd) {
...
     }
 
-    destroy_pool(tmp_pool);
-
     key_fmtlen = jot_parsed->bufsz - jot_parsed->buflen;
     key_fmtbuf[key_fmtlen] = '\0';
     key_fmt = (unsigned char *) pstrndup(c->pool, (char *) key_fmtbuf,
       key_fmtlen);
+
+    destroy_pool(tmp_pool);
```
Castaglia, line 696: These particular use-after-free bugs would occur at parse/startup time, and only when `-DPR_DEVEL_NO_FREELIST` is defined, which only happens for specialized developer builds.

PR #2248, opened 2026-07-21
The `mod_ctrls_admin` module was completely lacking in integration/re…

Castaglia:
…gression tests, so I added them.

In doing so, there were some tidbits that needed minor tweaking, in terms of better error messages for invalid requests, etc.

### DIFF
--- a/contrib/mod_ctrls_admin.c
+++ b/contrib/mod_ctrls_admin.c
@@ -35,9 +35,9 @@
 # error "ProFTPD 1.3.6rc2 or later required"
 #endif
 
-#ifndef PR_USE_CTRLS
+#if !defined(PR_USE_CTRLS)
 # error "Controls support required (use --enable-ctrls)"
-#endif
+#endif /* PR_USE_CTRLS */
 
 /* Values for the stop flags */
 #define CTRL_STOP_DEFAULT     (1 << 0)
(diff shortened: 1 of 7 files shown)


contrib/mod_ctrls_admin.c
```diff
@@ -433,7 +433,15 @@ static int ctrls_handle_debug(pr_ctrls_t *ctrl, int reqargc,
...
 
     } else if (reqargc == 2) {
-      level = atoi(reqargv[1]);
+      char *ptr = NULL;
+
+      level = (int) strtol(reqargv[1], &ptr, 10);
+      if (ptr != NULL && *ptr) {
+        pr_ctrls_add_response(ctrl, "debug level must be numeric: '%s'",
+          reqargv[1]);
+        return PR_CTRLS_STATUS_WRONG_PARAMETERS;
```
Castaglia, line 442: With this, if the given level is _e.g._ "foo", then rather than interpreting that as zero (as `atoi(3)` would return), we properly notice the non-numeric nature, and return an error.

src/bindings.c
```diff
@@ -769,8 +786,12 @@ server_rec *pr_ipbind_get_server(const pr_netaddr_t *addr, unsigned int port) {
   /* Not found in binding list, and no DefaultServer, so see if it's the
    * loopback address
    */
-  if (ipbind_localhost_server &&
-      pr_netaddr_is_loopback(addr)) {
+  if (pr_netaddr_is_loopback(addr) &&
+      ipbind_localhost_server != NULL &&
+      ipbind_localhost_server->ib_isactive) {
```
Castaglia, line 791: In this particular code path, we were _not_ honoring the `ib_isactive` flag, for _e.g._ "downed" servers, as revealed by the new integration tests.

PR #2254, opened 2026-07-25
Address a potential use-after-free issue in SCP, when buffering up pa…

Castaglia:
…ckets for receiving control messages.

Thanks to Elijah Chia for reporting this issue.

### DIFF
--- a/contrib/mod_sftp/scp.c
+++ b/contrib/mod_sftp/scp.c
@@ -382,7 +382,7 @@ static int write_confirm(pool *p, uint32_t channel_id, int code,
 
 /* Functions for receiving files from the client. */
 
-static int recv_ctl(uint32_t channel_id, struct scp_path *sp,
+static int recv_ctl(pool *p, uint32_t channel_id, struct scp_path *sp,
     unsigned char *data, uint32_t datalen,
     unsigned char **ctl_data, uint32_t *ctl_datalen) {
   register int i;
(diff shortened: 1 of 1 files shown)


contrib/mod_sftp/scp.c
```diff
@@ -426,13 +426,15 @@ static int recv_ctl(uint32_t channel_id, struct scp_path *sp,
...
    */
   if (have_newline == TRUE) {
-    *ctl_data = sp->ctl_data;
     *ctl_datalen = sp->ctl_datalen;
+    *ctl_data = palloc(p, *ctl_datalen);
+    memcpy(*ctl_data, sp->ctl_data, sp->ctl_datalen);
 
     sp->ctl_data = NULL;
     sp->ctl_datalen = 0;
     destroy_pool(sp->ctl_pool);
```
Castaglia, line 435: The issue, accurately spotted, was that we would allocate the data to be returned to caller (via `*ctl_data`, `*ctl_datalen`) out of this `sp->ctl_pool` -- but we would _destroy_ that pool here, before actually returning to the caller.

The reason this isn't really an issue for normal builds is because of ProFTPD's memory pool "free list", where destroyed pools are not actually `free(3)`'d, but kept on an internal free list, for whenever the code next requests a pool.  This longer-than-normal lifetime means that, in practice, this use-after-free issue is seldom actually seen.

Issue #2256, opened 2026-07-27
Cross-filesystem RNTO fallback follows source and destination symlinks unexpectedly


UWCSZhou:
## Summary

I reproduced this on an unmodified build of current `master` at
[`ce4015cb3f70b85e2b3d9af4799e1733b70115ae`](https://github.com/proftpd/proftpd/commit/ce4015cb3f70b85e2b3d9af4799e1733b70115ae).

When `RNTO` crosses a filesystem boundary, the underlying `rename(2)` returns
`EXDEV`. ProFTPD then falls back to copying the source contents to the
destination and unlinking the source. This fallback follows source and
destination symlinks instead of preserving the symlink directory-entry
semantics of a rename.

A cross-filesystem operation could instead preserve those semantics or reject
the operation without changing either object.

## Reproduction results

The
[standalone reproducer](https://gist.github.com/UWCSZhou/4dc159d5f5a6687f9a68dee24ae3eb24)
exercises both affected forms and equivalent same-filesystem controls. The
gist contains only the README, Python reproducer, and wrapper script, and can
be downloaded as a ZIP.

| Case | Cross-filesystem result | Same-filesystem control |
| --- | --- | --- |
| Source is a symlink | `RNTO` returns `250 Rename successful`; the source symlink is removed and the destination becomes a regular file containing the symlink target's data | The symlink entry itself is moved and its target remains unchanged |
| Destination is a symlink | `RNTO` returns `250 Rename successful`; the destination symlink remains while its target is truncated and overwritten, and the source is removed | The destination symlink entry is replaced by the source file and its former target remains unchanged |

Both cross-filesystem forms reproduced, while both same-filesystem controls
preserved normal link-entry rename semantics.

ProFTPD's current same-filesystem tests explicitly require the same behavior:
the destination symlink must be replaced rather than followed, and a source
symlink must be moved as a link:

- [`RNTO.pm`, destination-symlink expectation](https://github.com/proftpd/proftpd/blob/ce4015cb3f70b85e2b3d9af4799e1733b70115ae/tests/t/lib/ProFTPD/Tests/Commands/RNTO.pm#L305-L318)
- [`RNTO.pm`, source-symlink expectation](https://github.com/proftpd/proftpd/blob/ce4015cb3f70b85e2b3d9af4799e1733b70115ae/tests/t/lib/ProFTPD/Tests/Commands/RNTO.pm#L570-L580)

Those tests use one filesystem and therefore do not enter the `EXDEV`
[...]

PR #2259, opened 2026-07-28
Issue #2256: If an FTP `RNTO` command requests renaming/moving a syml…

Castaglia:
…ink across filesystem mounts, the fallback code followed the source (and destination) symlinks unexpectedly.

Normal `mv(1)` semantics only move the source symlink without following it, and without following any symlink at the destination path.

### DIFF
--- a/modules/mod_core.c
+++ b/modules/mod_core.c
@@ -6303,8 +6303,127 @@ MODRET core_dele(cmd_rec *cmd) {
   return PR_HANDLED(cmd);
 }
 
+static int handle_cross_mount_rename(cmd_rec *cmd, const char *src_path,
+    const char *dst_path, struct stat *dst_st) {
+  int res, xerrno;
+  struct stat src_st;
+
+  pr_fs_clear_cache2(src_path);
+  res = pr_fsio_lstat(src_path, &src_st);
+  if (res < 0) {
+    xerrno = errno;
+
+    pr_log_debug(DEBUG4, "Unable to stat '%s' for cross filesystem mount "
+      "point rename: %s", src_path, strerror(xerrno));
+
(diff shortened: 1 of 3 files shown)


modules/mod_core.c
```diff
@@ -6365,23 +6484,28 @@ MODRET core_rnto(cmd_rec *cmd) {
...
    * rename file already exists.
    */
+  allow_overwrite = get_param_ptr(CURRENT_CONF, "AllowOverwrite", FALSE);
+
   pr_fs_clear_cache2(path);
-  if ((!allow_overwrite || *allow_overwrite == FALSE) &&
-      pr_fsio_stat(path, &st) == 0) {
-    pr_log_debug(DEBUG6, "AllowOverwrite denied permission for %s", path);
-    pr_response_add_err(R_550, _("%s: Rename permission denied"), cmd->arg);
+  res = pr_fsio_lstat(path, &st);
```
Castaglia, line 6493: Here, we switch from using `stat(2)` on this target path to `lstat(2)`.  It looks minor, but it _does_ mean that if the target path is a dangling symlink, now `AllowOverwrite on` is needed to overwrite it, whereas previously such dangling symlinks would be overwritten improperly.

tests/t/lib/ProFTPD/Tests/Commands/RNTO.pm
```diff
@@ -260,6 +594,8 @@ sub rnto_abs_dst_symlink {
     AuthGroupFile => $setup->{auth_group_file},
     AuthOrder => 'mod_auth_file.c',
 
+    AllowOverwrite => 'on',
+
```
Castaglia, line 598: This is now needed because of the (proper) use of `lstat(2)` on the target path, rather that `stat(2)`, as the former detects even dangling symlinks as existing.

Issue #2265, opened 2026-08-03
Possible use-after-free issue via FTP `STAT` command using `-C` option


Castaglia:
Abdullah Al Ishtiaq reported the following issue:

ProFTPD contains a heap use-after-free in the mod_ls module's handling of FTP `STAT` commands with the `-C` listing option. An authenticated FTP user with permission to create entries in a directory can cause mod_ls to retain a pointer to a command-owned memory pool after that pool has been destroyed. A second `STAT` command on the same control connection then allocates through the stale pool pointer.

AddressSanitizer reports a heap use-after-free in `pcalloc` called by the `addfile`. The affected session child process terminates before it can handle a subsequent command on the control connection.

Castaglia:
There are a couple of parts to addressing this issue.  First, to prevent the unintentional use-after-free scenario via proper memory pool management; this will be backported to the 1.3.9 branch as well.

The second part is to remove support/use of the `-C` option for FTP `STAT` commands; requesting columnar displays of file listings via the FTP `STAT` command does not really make as much sense.  However, this would be a change in behavior, and so this `-C` related fix will _not_ be backported to the 1.3.9 branch.

Castaglia:
Fixed in master as described, and backported fix to the 1.3.9 branch.

PR #2266, opened 2026-08-03
Issue #2265: Make sure that the memory allocated for sorting files is…

Castaglia:
… properly cleaned up, leaving no stale pointers, once its pool is destroyed.

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -938,7 +938,15 @@ struct sort_filename {
 static struct filename *head = NULL;
 static struct filename *tail = NULL;
 static array_header *sort_arr = NULL;
-static pool *fpool = NULL;
+static pool *sort_pool = NULL;
+
+static void sort_cleanup(void *user_data) {
+  sort_pool = NULL;
+  sort_arr = NULL;
+  head = tail = NULL;
+  colwidth = 0;
+  filenames = 0;
+}
 
 static void addfile(cmd_rec *cmd, const char *name, const char *suffix,
     time_t sort_time, off_t size) {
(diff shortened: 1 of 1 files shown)


modules/mod_ls.c
```diff
@@ -938,7 +938,14 @@ struct sort_filename {
 static struct filename *head = NULL;
 static struct filename *tail = NULL;
 static array_header *sort_arr = NULL;
-static pool *fpool = NULL;
+static pool *sort_pool = NULL;
```
Castaglia, line 941: While I'm here, I'm changing the name of the pool to be less terse, and more accurately reflect how it is used.

modules/mod_ls.c
```diff
@@ -1102,12 +1111,8 @@ static int outputfiles(cmd_rec *cmd) {
...
     }
 
-    destroy_pool(fpool);
-    fpool = NULL;
-    sort_arr = NULL;
-    head = tail = NULL;
-    colwidth = 0;
-    filenames = 0;
+    destroy_pool(sort_pool);
+    sort_pool = NULL;
```
Castaglia, line 1115: No need for the cleanup here, since we've now moved it to a destructor callback invoked when the pool is destroyed.

modules/mod_ls.c
```diff
@@ -1196,37 +1201,39 @@ static int outputfiles(cmd_rec *cmd) {
...
 
-      if (sendline(0, "%s%s", q->line, pad) < 0) {
-        return -1;
+      if (session.curr_cmd_id == PR_CMD_LIST_ID) {
+        res = sendline(0, "%s%s", q->line, pad);
+
+      } else {
+        pr_response_add(NULL, "%s%s", q->line, pad);
+        res = 0;
+      }
```
Castaglia, line 1210: There are two necessary fixes here.  First, ensure that for columnar listings for a `STAT` command, we add the line to the FTP response, and _not_ attempt to send it via a (nonexistent, for `STAT` commands) data channel.  Second, rather than returning early if there is an issue, we set a return value, break out of the loop, and thus ensure that the destruction of `sort_pool` still happens, success or failure, at the end of the function.

PR #2267, opened 2026-08-03
Issue #2265: Since the -C and -U options do not make sense for FTP `S…

Castaglia:
…TAT` commands, make it so that they are silently ignored.

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -108,8 +108,7 @@ static int
     opt_S = 0,
     opt_t = 0,
     opt_U = 0,
-    opt_u = 0,
-    opt_STAT = 0;
+    opt_u = 0;
 
 /* Determines which struct st timestamp is used for sorting, if any. */
 static int ls_sort_by = 0;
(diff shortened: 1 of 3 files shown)


modules/mod_ls.c
```diff
@@ -108,8 +108,7 @@ static int
     opt_S = 0,
     opt_t = 0,
     opt_U = 0,
-    opt_u = 0,
-    opt_STAT = 0;
```
Castaglia, line 112: The mod_ls module inconsistently used `opt_STAT` and `session.curr_cmd_id`; now we just pick one (`session.curr_cmd_id`) and use only that.

modules/mod_ls.c
```diff
@@ -1201,14 +1205,7 @@ static int outputfiles(cmd_rec *cmd) {
...
 
-      if (session.curr_cmd_id == PR_CMD_LIST_ID) {
-        res = sendline(0, "%s%s", q->line, pad);
-
-      } else {
-        pr_response_add(NULL, "%s%s", q->line, pad);
-        res = 0;
-      }
-
+      res = sendline(0, "%s%s", q->line, pad);
```
Castaglia, line 1208: With -C now ignored for STAT commands, the only time this code path is taken is for LIST commands, which will have a data connection for `sendline()` to use.

PR #2268, opened 2026-08-04
Make sure that _all_ listing options are properly cleared/reset, as i…

Castaglia:
…s the "sort by" flag.

### DIFF
--- a/modules/mod_ls.c
+++ b/modules/mod_ls.c
@@ -110,13 +110,21 @@ static int
     opt_U = 0,
     opt_u = 0;
 
-/* Determines which struct st timestamp is used for sorting, if any. */
+/* Determines which struct st timestamp is used for sorting, if any.  The
+ * default, when sorting, is by mtime.
+ */
 static int ls_sort_by = 0;
 #define LS_SORT_BY_MTIME	100
 #define LS_SORT_BY_CTIME	101
 #define LS_SORT_BY_ATIME	102
 
 static char cwd[PR_TUNABLE_PATH_MAX+1] = "";
+
+static void clear_list_opts(void) {
+  opt_1 = opt_A = opt_a = opt_B = opt_C = opt_c = opt_d = opt_F = opt_h =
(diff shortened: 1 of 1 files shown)


modules/mod_ls.c
```diff
@@ -1706,6 +1708,7 @@ static void parse_list_opts(char **opt, int *glob_flags, int handle_plus_opts) {
 
         case 't':
           opt_t = 1;
+          ls_sort_by = LS_SORT_BY_MTIME;
```
Castaglia, line 1717: This was always the implicit default; this makes it _explicit_.

modules/mod_ls.c
```diff
@@ -1714,7 +1717,7 @@ static void parse_list_opts(char **opt, int *glob_flags, int handle_plus_opts) {
         case 'U':
           if (session.curr_cmd_id != PR_CMD_STAT_ID) {
             opt_U = 1;
-            opt_c = opt_S = opt_t = 0;
+            opt_c = opt_S = opt_t = opt_u = 0;
```
Castaglia, line 1726: The `-U` option is for "no sorting", which means that the `-u` (sort by file atime) flag _should_ have been being cleared, and wasn't.  Minor bug.

modules/mod_ls.c
```diff
@@ -1916,8 +1928,9 @@ static int dolist(cmd_rec *cmd, const char *opt, const char *resp_code,
   ls_curtime = time(NULL);
 
   if (clear_flags) {
-    opt_1 = opt_A = opt_a = opt_B = opt_C = opt_d = opt_F = opt_h = opt_n =
-      opt_r = opt_R = opt_S = opt_t = opt_L = 0;
+    opt_1 = opt_A = opt_a = opt_B = opt_C = opt_c = opt_d = opt_F = opt_h =
+      opt_L = opt_l = opt_n = opt_R = opt_r = opt_S = opt_t = opt_U = opt_u = 0;
```
Castaglia: The prior list of options to be cleared/reset did not contain _all_ of the possible options.

Issue #2272, opened 2026-08-05
Passive FTP data transfers do not honor AllowForeignAddress policy properly


Castaglia:
securityraven@proton.me disclosed the following issue.

When accepting data connections for FTP passive transfers (_i.e._ FTP `EPSV` or `PASV` commands), ProFTPD does not verify whether the connecting peer address is "foreign" by default (meaning [`AllowForeignAddress`](http://www.proftpd.org/docs/modules/mod_core.html#AllowForeignAddress) has not been explicitly configured).  The expected behavior is that all data connections, active or passive, have the peer addresses verified as "foreign" or not; if "foreign", then such connections should be rejected, unless explicitly configured to allow them via `AllowForeignAddress`.

This is a regression caused by changes for Issue #1346, and specifically affect passive data transfers, but _not_ active data transfers.

FTPS sessions are usually not vulnerable to this, as they require that the control connection's TLS session be reused for data connections by default.

Castaglia:
Note that this regression happened because I copied/re-implemented the `<Class>`-specific handling of the `AllowForeignAddress` directive for FTP passive data transfers, rather than reusing the existing code that was done for _active_ data transfers, which Does The Right Thing(tm).

Thus the fix for this involves refactoring that `AllowForeignAddress` logic into a single function in the Inet API, and ensuring that all data transfers, active or passive, use that single function -- and that it behaves as expected whether `AllowForeignAddress` is explicitly configured or not.

Castaglia:
Fixed in master, and backported to the 1.3.9 branch.

PR #2273, opened 2026-08-05
Issue #2272: Restore the expected implicit behavior for FTP passive d…

Castaglia:
…ata transfers and foreign addresses.

### DIFF
--- a/modules/mod_core.c
+++ b/modules/mod_core.c
@@ -4052,14 +4052,14 @@ MODRET core_pasv(cmd_rec *cmd) {
 MODRET core_port(cmd_rec *cmd) {
   const pr_netaddr_t *listen_addr = NULL, *port_addr = NULL;
   char *port_info;
-#ifdef PR_USE_IPV6
+#if defined(PR_USE_IPV6)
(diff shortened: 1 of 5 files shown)


modules/mod_core.c
```diff
@@ -4052,14 +4052,14 @@ MODRET core_pasv(cmd_rec *cmd) {
 MODRET core_port(cmd_rec *cmd) {
   const pr_netaddr_t *listen_addr = NULL, *port_addr = NULL;
   char *port_info;
-#ifdef PR_USE_IPV6
+#if defined(PR_USE_IPV6)
```
Castaglia, line 4055: Opportunistic stylistic nit; no functional change.

modules/mod_core.c
```diff
@@ -4444,8 +4410,6 @@ MODRET core_eprt(cmd_rec *cmd) {
    */
   *tmp = '\0';
 
-  memset(&na, 0, sizeof(na));
```
Castaglia, line 4447: This was unnecessary, since the `na` struct is properly initialized via `pr_netaddr_clear()` earlier on line 4315, which does the same thing.

PR #2274, opened 2026-08-05
If the `stat(2)` calls fail while processing the UserOwner/GroupOw…

Castaglia:
…ner directives for FTP uploads, an uninitialized `struct stat` might be used, leading to unexpected results.

This was reported by securityraven@proton.me.

### DIFF
--- a/modules/mod_xfer.c
+++ b/modules/mod_xfer.c
@@ -851,41 +851,43 @@ static void stor_chown(pool *p) {
         pr_log_debug(DEBUG0,
           "'%s' stat(2) error during root chmod: %s", xfer_path,
           strerror(errno));
-      }
-
-      /* The chmod happens after the chown because chown will remove
-       * the S{U,G}ID bits on some files (namely, directories); the subsequent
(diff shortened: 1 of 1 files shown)


modules/mod_xfer.c
```diff
@@ -851,41 +851,43 @@ static void stor_chown(pool *p) {
...
-       * that the mode can be set (a file might be being "given away", and if
-       * root privs aren't used, the chmod() will fail because the old owner/
-       * session user doesn't have the necessary privileges to do so).
-       */
-      xerrno = 0;
-      PRIVS_ROOT
-      res = pr_fsio_chmod_with_error(p, xfer_path, st.st_mode, &err);
-      xerrno = errno;
-      PRIVS_RELINQUISH
+      } else {
```
Castaglia, line 855: The fix for this is to ensure that the `chmod(2)` calls, using the struct stat `st`,  only occur in the non-error-handling branches of the preceding `stat(2)` conditionals.

PR #2281, opened 2026-08-07
As fallout from Issue #2279, we can remove a now-redundant "is the te…

Castaglia:
…xt length greater than zero" check, as now it is guaranteed to be a non-empty string.


modules/mod_auth_file.c
```diff
@@ -1600,9 +1600,9 @@ MODRET authfile_chkpass(cmd_rec *cmd) {
     pr_log_debug(DEBUG0, MOD_AUTH_FILE_VERSION
       ": error using crypt(3) for user '%s': %s", user, strerror(xerrno));
 
-    if (ciphertxt_passlen > 0 &&
```
Castaglia, line 1603: This `ciphertxt_passlen > 0` check is now guaranteed to always be true, and thus is no longer needed.

Issue #2282, opened 2026-08-08
.ftpaccess file policy bypass possible in certain configurations


Castaglia:
[securityraven@proton.me](mailto:securityraven@proton.me) disclosed the following behavior with handling of `.ftpaccess` file policies.

Under the right conditions of parent/ancestor directories with `.ftpaccess` files, and combination of `CWD` commands, it is possible for clients to unexpectedly access files, which _should_ have been protected by the parent/ancestor `.ftpaccess` configurations, to be accessed in a manner blocked by possible `<Limit>` configurations in those files.  This unexpected behavior occurs for authenticated users, and for `<Anonymous>` sessions as well.

Note that this only pertains to `<Limit>` sections configured via `.ftpaccess` files; it does not affect `<Directory>` sections configured directly into the ProFTPD configuration file(s).

Castaglia:
Fixed in master, and backported to the 1.3.9 branch.

PR #2283, opened 2026-08-08
Issue #2282: When checking for any `<Limit>` policies on a given path…

Castaglia:
…, make sure to _always_ recurse upward through all parents looking for possible `.ftpaccess` files.

### DIFF
--- a/include/dirtree.h
+++ b/include/dirtree.h
@@ -217,7 +217,11 @@ int pr_config_get_xfer_bufsz2(int);
 int pr_config_get_server_xfer_bufsz(int);
 
 config_rec *dir_match_path(pool *, char *);
-void build_dyn_config(pool *, const char *, struct stat *, unsigned char);
+
+void build_dyn_config(pool *p, const char *path, struct stat *st,
+  unsigned char recurse);
+void build_dyn_config2(pool *p, const char *path, struct stat *st);
+
 unsigned char dir_hide_file(const char *);
 int dir_check_full(pool *, cmd_rec *, const char *, const char *, int *);
 int dir_check_limits(cmd_rec *, config_rec *, const char *, int);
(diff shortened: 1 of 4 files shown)


include/dirtree.h
```diff
@@ -217,7 +217,11 @@ int pr_config_get_xfer_bufsz2(int);
 int pr_config_get_server_xfer_bufsz(int);
 
 config_rec *dir_match_path(pool *, char *);
-void build_dyn_config(pool *, const char *, struct stat *, unsigned char);
+
+void build_dyn_config(pool *p, const char *path, struct stat *st,
```
Castaglia, line 221: This `build_dyn_config()` function is part of the public API, and _is_ used in some other third-party modules.  Thus we cannot simply remove the `unsigned char recurse` Boolean parameter, which would change the function signature.

So instead, we create a `build_dyn_config2()` wrapper which always uses recursion.  And that use of recursion addresses the buggy behavior.

src/dirtree.c
```diff
@@ -1972,7 +1997,7 @@ int dir_check(pool *pp, cmd_rec *cmd, const char *group, const char *path,
         (session.anon_config ? session.anon_config : NULL));
 
   if (c == NULL ||
-      strncmp(c->name, fullpath, strlen(c->name)) != 0) {
+      is_ancestor_path(c->name, fullpath) != TRUE) {
```
Castaglia, line 2000: The report from securityraven@proton.me also correctly noted that the current check, using:
```
  strncmp(c->name, fullpath, strlen(c->name))
```
could unexpectedly match, _e.g._ "/srv/tmp" could prefix-match "/srv/tmpA".  So I created a helper `is_ancestor_path()` function, which matches the prefix _and_ asserts that the next character is the path separator, to more explicitly implement the desired checking behavior here.

Issue #2293, opened 2026-08-20
Regression with custom user/group info SQLNamedQuery in 1.3.9d


Castaglia:
As reported in [Bug #4515](http://bugs.proftpd.org/show_bug.cgi?id=4515), the recent 1.3.9d (and 1.3.10rc4) ProFTPD releases shipped a regression in the use of [`SQLNamedQuery`](http://www.proftpd.org/docs/contrib/mod_sql.html#SQLNamedQuery) to provide custom user/group information, using _e.g._:
```
SQLNamedQuery get-group-by-name get-group-by-name SELECT "g.groupname,g.gid,u.username AS members FROM users u,  users_groups ug,groups g WHERE u.id = ug.users_id AND g.id = ug.groups_id and g.groupname = '%{0}'"
```

This regression was caused by changes to protect against use of numeric variables (_e.g._ `%{0}`) in `SQLNamedQuery` statements as vehicles for SQL injection and/or null pointer dereferences; see https://github.com/proftpd/proftpd/commit/cbcc9b41d5ab07ada2da72c4eec121e7abfea360.  Unfortunately, the implemented restrictions inadvertently impacted such custom user/group info constructions unintentionally.

@pghmcfc @hpreusse I should have a PR shortly for this regression.

Castaglia:
Fixed in master, and backported to the 1.3.9 branch.

hpreusse:
Thanks for notification + patch.

PR #2294, opened 2026-08-20
Issue #2293: Mark these internally-generated SQL lookups explicitly, …

Castaglia:
…so that they pass the implemented restrictions for resolving numeric variables in SQL statements.

This restores the previously working behavior of `SQLNamedQuery` statements using numeric variables to provide custom user/group information.

### DIFF
--- a/contrib/mod_sql.c
+++ b/contrib/mod_sql.c
@@ -1809,7 +1809,7 @@ static int sql_getuserprimarykey(cmd_rec *cmd, const char *username) {
     }
 
   } else {
-    mr = sql_lookup(sql_make_cmd(cmd->tmp_pool, 3, MOD_SQL_DEF_CONN_NAME, ptr,
+    mr = sql_lookup(sql_make_cmd(cmd->tmp_pool, 3, "sql_lookup", ptr,
       username));
     if (check_response(mr, 0) < 0) {
       return -1;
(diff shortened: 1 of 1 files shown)


contrib/mod_sql.c
```diff
@@ -1809,7 +1809,7 @@ static int sql_getuserprimarykey(cmd_rec *cmd, const char *username) {
     }
 
   } else {
-    mr = sql_lookup(sql_make_cmd(cmd->tmp_pool, 3, MOD_SQL_DEF_CONN_NAME, ptr,
+    mr = sql_lookup(sql_make_cmd(cmd->tmp_pool, 3, "sql_lookup", ptr,
```
Castaglia, line 1812: Why this magic-looking `"sql_lookup"` text for the generated `cmd_rec` name?  See https://github.com/proftpd/proftpd/blob/master/contrib/mod_sql.c#L968

contrib/mod_sql.c
```diff
@@ -2342,8 +2342,8 @@ static struct group *sql_getgroup(cmd_rec *cmd, struct group *g) {
 
     sql_log(DEBUG_WARN, "cache miss for GID '%s'", gidstr);
 
-    if (!cmap.groupcustombyid) {
-      if (cmap.grpgidfield) {
+    if (cmap.groupcustombyid == NULL) {
+      if (cmap.grpgidfield != NULL) {
```
Castaglia, line 2346: Stylistic nit; no functional change.

PR #2300, opened 2026-08-25
Refactor the construction of the key exchange and hostkey algorithm l…

Castaglia:
…ists for SSH KEXINIT so that all algorithm lists are in the same place.

Previously, these lists were spread between `kex.c` and `crypto.c`, leading to some unexpected duplication when adding new algorithms.

### DIFF
--- a/contrib/mod_sftp/crypto.c
+++ b/contrib/mod_sftp/crypto.c
@@ -23,6 +23,7 @@
 
 #include "mod_sftp.h"
 #include "crypto.h"
+#include "keys.h"
 #include "provider.h"
 #include "umac.h"
 
@@ -40,6 +41,8 @@
     defined(PR_USE_OPENSSL_ENGINE)
 static const char *crypto_engine = NULL;
 #endif /* PR_USE_OPENSSL_ENGINE */
+
+static const char *trace_channel = "ssh2";
 
 struct sftp_cipher {
   const char *name;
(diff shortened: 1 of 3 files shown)


contrib/mod_sftp/crypto.c
```diff
@@ -246,42 +249,66 @@ static const char *hostkey_algos[] = {
...
+ * REQUIRED algorithms is conspicuously absent:
+ *
+ *   diffie-hellman-group1-sha1
+ *
+ * This exchange has a weak hardcoded DH group, and will thus only be used
+ * if explicitly requested via SFTPKeyExchanges, or if the AllowWeakDH
+ * SFTPOption is used.
+ */
 static const char *key_exchanges[] = {
-  "diffie-hellman-group1-sha1",
```
Castaglia, line 250: The order of these key exchange algorithms, in `crypto.c`, now matches the preferred order that we send to clients, as implemented in `kex.c`.

contrib/mod_sftp/kex.c
```diff
@@ -1567,93 +1567,10 @@ static const char *get_preferred_name(pool *p, const char *names) {
...
-
-  } else {
-    register unsigned int i;
-
-    for (i = 0; kex_exchanges[i]; i++) {
-      res = pstrcat(p, res, *res ? "," : "", pstrdup(p, kex_exchanges[i]),
-        NULL);
-    }
+static const char *get_kexinit_key_exchange_list(pool *p) {
+  char *res;
```
Castaglia, line 1571: We still need a wrapper here in `kex.c` for handling the list of key exchange algorithms, because part of the list requires knowing about some internal kex-specific state.  So we obtain the normal algorithm list from `crypto.c`, then do a little post-processing here.

Issue #2301, opened 2026-08-26
Unauthenticated SNMP UDP packets kill SNMP agent due to underflow computation


Castaglia:
The Kimi Security Team disclosed the following issue with the mod_snmp module:

A pre-authentication remote denial of service exists in the `mod_snmp` module. A remote unauthenticated attacker can deterministically kill the standalone SNMP agent process with a few UDP datagrams; the agent is not restarted until the daemon is restarted, leaving SNMP monitoring silently dead.

The underflow computation occurs when parsing the ASN1 length fields of the received SNMP packets into a provided buffer; the client-provided length values could lead to excessively long length values that cause an underflow, leading to large memory allocations and eventually triggering a SIGSEGV.

Castaglia:
Fixed in master, and backported to the 1.3.9 branch.

PR #2302, opened 2026-08-26
Issue #2301: Ensure that we have sufficient remaining buffer space fo…

Castaglia:
…r copying data when reading the ASN.1 length fields in an SNMP message.

In addition, we now delay reading the SNMP message PDU (the "payload") bytes until after the message has been authenticated, and additionally enforce our maximum message size when reading other fields (e.g. text strings, integer, OIDs).

### DIFF
--- a/contrib/mod_snmp/asn1.c
+++ b/contrib/mod_snmp/asn1.c
@@ -223,6 +223,17 @@ static int asn1_read_len(pool *p, unsigned char **buf, size_t *buflen,
       return -1;
     }
 
+    /* Make sure the provided buffer is large enough for the amount of data
+     * received (Issue #2301).
+     */
+    if (byte > *buflen) {
+      (void) pr_log_writefile(snmp_logfd, MOD_SNMP_VERSION,
+        "ASN.1 error: provided ASN1 length value %u exceeds buffer (%lu bytes)",
+        (unsigned int) byte, (unsigned long) *buflen);
+      errno = EINVAL;
+      return -1;
+    }
+
     *asn1_len = 0;
     memmove(asn1_len, *buf, (size_t) byte);
     (*buf) += byte;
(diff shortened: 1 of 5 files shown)


contrib/mod_snmp/asn1.c
```diff
@@ -320,6 +331,16 @@ int snmp_asn1_read_int(pool *p, unsigned char **buf, size_t *buflen,
     return -1;
   }
 
+  /* Sanity check on the object length, to make sure it is not absurd. */
+  if (objlen > SNMP_ASN1_MAX_OBJECT_LEN) {
```
Castaglia, line 335: We do this check when reading ASN.1 headers; we should be consistent, and do it for other data types (such as integers, OIDs, and strings) as well.

contrib/mod_snmp/mib.c
```diff
@@ -44,7 +44,7 @@ static struct snmp_mib snmp_mibs[] = {
   /* Miscellaneous non-mod_snmp MIBs */
   { { SNMP_MGMT_SYS_OID_UPTIME, 0 },
     SNMP_MGMT_SYS_OIDLEN_UPTIME + 1,
-    0, TRUE, TRUE,
+    SNMP_DB_NOTIFY_F_SYS_UPTIME, TRUE, TRUE,
```
Castaglia, line 47: This addresses a minor nit I noticed while verifying this issue.

contrib/mod_snmp/mod_snmp.c
```diff
@@ -1405,12 +1405,12 @@ static int snmp_agent_handle_packet(int sockfd, pr_netaddr_t *agent_addr) {
     return -1;
   }
 
-  res = snmp_msg_read(pkt->pool, &(pkt->req_data), &(pkt->req_datalen),
-    &(pkt->community), &(pkt->community_len), &(pkt->snmp_version),
-    &(pkt->req_pdu));
+  /* Read the packet fields needed for authentication. */
+  res = snmp_msg_read_header(pkt->pool, &(pkt->req_data), &(pkt->req_datalen),
+    &(pkt->community), &(pkt->community_len), &(pkt->snmp_version));
```
Castaglia, line 1410: This is where we split reading the received datagram into two parts: first, just enough of the datagram fields are parsed for performing authentication (suupported SNMP versions + matching community string), and _then_ assuming authentication, we'll read the rest.

PR #2304, opened 2026-08-28
Once mod_dnsbl has determined whether to allow or reject the connecti…

Castaglia:
…on, it no longer needs to keep any `DNSBLLog` fd open any more, so close it.


Castaglia:
Many of the code changes here are stylistic nit cleanup and polish.  The main functional change is the earlier closing of the `DNSBLLog` fd.

PR #2310, opened 2026-09-06
When aggregating multiple MLSD entries into a buffer for sending to t…

Castaglia:
…he client, it is possible for an overly long entry to lead to an out-of-bounds write, due to failure to account for `sstrcat()` behavior.

Thanks to k3rn3lbr3ach3r of Team kBxAc for reporting this issue.

Note that this differs from Issue #2146, which is about dealing with a single overly long MSLD entry, while this is about dealing with buffering of multiple MLSD entries into the aggregation buffer.

### DIFF
--- a/modules/mod_facts.c
+++ b/modules/mod_facts.c
@@ -516,6 +516,17 @@ static void facts_mlinfobuf_init(void) {
   if (mlinfo_buf == NULL) {
     mlinfo_bufsz = pr_config_get_server_xfer_bufsz(PR_NETIO_IO_WR);
 
+    /* Make sure that the preferred network IO write buffer size is large
+     * enough; some admin may have configured a too-small SocketOptions setting
+     * for our needs.
+     */
+    if (mlinfo_bufsz < FACTS_MLINFO_BUFSZ) {
+      pr_trace_msg("data", 9, "configured network buffer size (%lu bytes) is "
+        "too small for MLSD entries (%lu bytes), ignoring",
+        (unsigned long) mlinfo_bufsz, (unsigned long) FACTS_MLINFO_BUFSZ);
+      mlinfo_bufsz = FACTS_MLINFO_BUFSZ;
+    }
+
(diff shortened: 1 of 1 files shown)


modules/mod_facts.c
```diff
@@ -516,6 +516,17 @@ static void facts_mlinfobuf_init(void) {
...
+    /* Make sure that the preferred network IO write buffer size is large
+     * enough; some admin may have configured a too-small SocketOptions setting
+     * for our needs.
+     */
+    if (mlinfo_bufsz < FACTS_MLINFO_BUFSZ) {
+      pr_trace_msg("data", 9, "configured network buffer size (%lu bytes) is "
+        "too small for MLSD entries (%lu bytes), ignoring",
+        (unsigned long) mlinfo_bufsz, (unsigned long) FACTS_MLINFO_BUFSZ);
+      mlinfo_bufsz = FACTS_MLINFO_BUFSZ;
+    }
```
Castaglia, line 528: This implements a defense-in-depth, for ensuring that the allocated aggregation buffer is always large enough for at least one entry.

modules/mod_facts.c
```diff
@@ -534,25 +545,51 @@ static void facts_mlinfobuf_init(void) {
...
     if (facts_mlinfobuf_flush() < 0) {
       return -1;
     }
+
+    /* Calculate the mlinfo_buf capacity again due to the flushing. */
+    mlinfo_capacity = mlinfo_bufsz - mlinfo_buflen;
   }
 
-  sstrcat(mlinfo_bufptr, buf, mlinfo_bufsz - mlinfo_buflen);
-  mlinfo_bufptr += buflen;
```
Castaglia, line 554: The previous bad behavior was here.  `sstrcat()` might not have written all of the `buflen` bytes, as when `buflen` exceeded the available capacity -- but here we act as if it did, and increment this pointer by (now) too much.

PR #2314, opened 2026-09-07
Fix memory leaks from rotating/generating TLS session ticket keys.

Castaglia:
Thanks to Bruno Henc for reporting this issue.

### DIFF
--- a/contrib/mod_tls.c
+++ b/contrib/mod_tls.c
@@ -4046,7 +4046,7 @@ static void tls_scrub_pkey(tls_pkey_t *k) {
     k->dsa_passlen = 0;
   }
 
-#ifdef PR_USE_OPENSSL_ECC
+#if defined(PR_USE_OPENSSL_ECC)
   if (k->ec_pkey != NULL) {
     pr_memscrub(k->ec_pkey, k->pkeysz);
     free(k->ec_pkey_ptr);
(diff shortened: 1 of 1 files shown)


contrib/mod_tls.c
```diff
@@ -4046,7 +4046,7 @@ static void tls_scrub_pkey(tls_pkey_t *k) {
     k->dsa_passlen = 0;
   }
 
-#ifdef PR_USE_OPENSSL_ECC
+#if defined(PR_USE_OPENSSL_ECC)
```
Castaglia, line 4049: Stylistic nit; no functional change.

contrib/mod_tls.c
```diff
@@ -4194,46 +4194,13 @@ static void tls_remove_pkey(tls_pkey_t *k) {
...
-    }
-#endif /* PR_USE_OPENSSL_ECC */
-
-    if (k->pkcs12_passwd != NULL &&
-        k->pkcs12_passlen > 0) {
-      passphrase_count++;
-    }
-  }
-
-  if (passphrase_count == 0) {
```
Castaglia, line 4228: Rather than counting the number of passphrases (which can be zero, for passphrase-less certs/keys), it's better to just free/scrub _all_ passphrase keys.  And it avoids memory leaks, too.

contrib/mod_tls.c
```diff
@@ -6898,12 +6865,14 @@ static int remove_expired_ticket_keys(void) {
...
        k;
-       k = k->next) {
+       k = knext) {
     time_t key_age;
 
+    knext = k->next;
     key_age = now - k->created;
     if (key_age > tls_ticket_key_max_age) {
       if (xaset_remove(tls_ticket_keys, (xasetmember_t *) k) == 0) {
+        destroy_ticket_key(k);
```
Castaglia, line 6875: Make sure we _destroy_ any TLS session ticket key we remove.

contrib/mod_tls.c
```diff
@@ -7037,14 +7007,14 @@ static void lock_ticket_keys(void) {
 }
 
 static void scrub_ticket_keys(void) {
-  struct tls_ticket_key *k, *next_k;
+  struct tls_ticket_key *k, *knext;
```
Castaglia, line 7010: This is a name change only, for consistency with other places where we use `knext`, not `next_k`.

PR #2322, opened 2026-09-08
Address a possible (albeit unlikely) out-of-bounds read/write for ove…

Castaglia:
…rly long async response lines.

Thanks to Bruno Henc for reporting this issue.

### DIFF
--- a/src/response.c
+++ b/src/response.c
@@ -325,7 +325,7 @@ void pr_response_add(const char *numeric, const char *fmt, ...) {
 
 void pr_response_send_async(const char *resp_numeric, const char *fmt, ...) {
   int res;
-  char buf[PR_TUNABLE_BUFFER_SIZE] = {'\0'};
+  char buf[PR_TUNABLE_BUFFER_SIZE] = {'\0'}, *ptr;
   va_list msg;
   size_t len, max_len;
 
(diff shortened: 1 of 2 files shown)


src/response.c
```diff
@@ -342,25 +342,40 @@ void pr_response_send_async(const char *resp_numeric, const char *fmt, ...) {
...
   va_end(msg);
 
-  buf[sizeof(buf) - 1] = '\0';
+  buf[sizeof(buf)-1] = '\0';
 
   resp_last_response_code = pstrdup(resp_pool, resp_numeric);
   resp_last_response_msg = pstrdup(resp_pool, buf + len + 1);
 
-  sstrcat(buf + res, "\r\n", max_len - res);
+  if (res < 0) {
```
Castaglia, line 366: The cause of the issue here was that we were not checking the `pr_vsprintf` return value to see whether it would have filled the buffer, and thus proceeded blithely on.  Now, we take such conditions into account.

tests/api/response.c
```diff
@@ -354,6 +355,60 @@ START_TEST (response_send_async_test) {
 }
 END_TEST
 
+START_TEST (response_send_async_too_large_test) {
+  char buf[PR_TUNABLE_BUFFER_SIZE * 2], expected[PR_TUNABLE_BUFFER_SIZE];
```
Castaglia, line 359: This input buffer size is guaranteed to exceed that of the `pr_response_send_async()` function deliberately, so that we test its boundary condition handling.

PR #2324, opened 2026-09-09
When reading cached TLS session and/or OCSP values from Memcached/Red…

Castaglia:
…is, apply more sanity checks to the obtained data.

In theory, these data sources should be secure, isolated and protected from manipulation, and thus trustworthy.  That said, a little defense-in-depth scrutiny to the obtained values never hurts.

Thanks to Bruno Henc for reporting these issues.

### DIFF
--- a/contrib/mod_tls_memcache.c
+++ b/contrib/mod_tls_memcache.c
@@ -393,6 +393,7 @@ static int sess_cache_entry_decode_json(pool *p, void *value, size_t valuesz,
   const char *key;
   char *entry, *text;
   double number = 0;
+  unsigned int sess_datalen = 0;
 
   entry = value;
   if (pr_json_text_validate(p, entry) == FALSE) {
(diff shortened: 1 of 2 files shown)


contrib/mod_tls_memcache.c
```diff
@@ -826,11 +850,13 @@ static int sess_cache_add_large_sess(tls_sess_cache_t *cache,
   struct sesscache_large_entry *entry = NULL;
 
   if (sess_len > TLS_MAX_SSL_SESSION_SIZE) {
-    const char *exceeds_key = sesscache_keys[SESSCACHE_KEY_EXCEEDS].key,
-      *max_len_key = sesscache_keys[SESSCACHE_KEY_MAX_LEN].key;
+    const char *exceeds_key, *max_len_key;
```
Castaglia, line 853: This is a stylistic nit, separating variable declaration from variable definition.

Issue #2319, opened 2026-09-07
Stale password-less authentication flag in mod_auth can allow wrong passwords to authenticate


Castaglia:
That is a session-wide flag in the mod_auth module that tracks whether the requested user can be allowed to authenticate without providing password.  This scenario currently only happens if the `AllowDotLogin` [`TLSOption`](http://www.proftpd.org/docs/contrib/mod_tls.html#TLSOptions), or [`TLSUserName`](http://www.proftpd.org/docs/contrib/mod_tls.html#TLSUserName), is used, _and_ the FTP `USER` command that successfully matches the available `~/.tlslogin` file; when this occurs, the mod_auth flag is set to _true_.

If that FTP `USER` command later fails for a different reason, mod_auth still retains that _true_ flag, even if a subsequent `USER` command from the same client requests a different username.

Thanks to Bruno Henc for responsibly disclosing this issue.

Castaglia:
Fixed in master, and backported to the 1.3.9 branch.

PR #2325, opened 2026-09-10
Issue #2319: In addition to properly resetting the module flag for pa…

Castaglia:
…sswordless authentication, we also need to remove the stashed configuration record as well.

Thanks to Bruno Henc for reporting this additional wrinkle to the original fix of the issue.

### DIFF
--- a/modules/mod_auth.c
+++ b/modules/mod_auth.c
@@ -66,10 +66,10 @@ static int auth_sess_init(void);
 /* auth_cmd_chk_cb() is hooked into the main server's auth_hook function,
  * so that we can deny all commands until authentication is complete.
  *
- * Note: Once this function returns true (i.e. client has authenticated),
- * it will ALWAYS return true.  At least until REIN is implemented.  Thus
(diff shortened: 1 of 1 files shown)


modules/mod_auth.c
```diff
@@ -170,7 +185,8 @@ static void auth_sess_reinit_ev(const void *event_data, void *user_data) {
 #if defined(PR_USE_LASTLOG)
   auth_use_lastlog = FALSE;
 #endif /* PR_USE_LASTLOG */
-  authenticated_without_pass = FALSE;
+  reset_authenticated_without_pass(NULL);
+  auth_have_authenticated = FALSE;
```
Castaglia, line 189: This ensures that any failed passwordless `USER` command does not leak across vhosts due to a `HOST` command.  (Although currently, this issue requires TLS before `HOST`, and `HOST` is rejected if TLS has been negotiated beforehand.  Still, best to be thorough, for future use cases might provide other avenues for `HOST` to be used, or other means of passwordless authentications.)

PR #2328, opened 2026-09-13
Make sure that the SFTP extensions which use paths also make use of `…

Castaglia:
…mod_procfs` and its checking for paths which should not be visible/used.


tests/t/lib/ProFTPD/Tests/Modules/mod_procfs/sftp.pm
```diff
@@ -2399,7 +2419,430 @@ sub procfs_sftp_ext_hardlink {
           print STDERR "# $line\n";
         }
 
-        if ($line =~ /HARDLINK $from_file denied by mod_procfs/) {
+        if ($line =~ /hardlink\@openssh.com $from_file denied by mod_procfs/) {
```
Castaglia, line 2422: This helps show that the "hardlink@openssh.com" SFTP extension is being properly guarded by mod_procfs _as_ an extension, rather than via custom internal command.

PR #2329, opened 2026-09-15
Eliminate a `memcpy(3)` used for all SFTP response packets by being a…

Castaglia:
… little more clever with buffer pointers.

Using flamegraphs locally did not indicate a major improvement, but did show that SFTP downloads were slightly faster.


contrib/mod_sftp/fxp.c
```diff
@@ -3664,20 +3664,39 @@ static struct fxp_packet *fxp_packet_read(uint32_t channel_id,
...
-  /* Use a buffer that's a little larger than the FX packet size */
-  buflen = bufsz = fxp->payload_sz + 32;
-  buf = ptr = palloc(fxp->pool, bufsz);
+  /* We know, a priori, that the payload pointer is actually 4 bytes past its
+   * actual start.
+   */
+  buflen = bufsz = fxp->payload_sz + 4;
+  buf = ptr = fxp->payload;
 
-  sftp_msg_write_data(&buf, &buflen, fxp->payload, fxp->payload_sz, TRUE);
```
Castaglia, line 3676: This `sftp_msg_write_data(..., TRUE)` wrote the 4-byte (`uint32_t`) header (value of `payload_sz`), then did a `memcpy()` of the data starting at `payload` for that length.

Now, we avoid that `memcpy()` entirely.

Issue #2330, opened 2026-09-15
Stack overflow possible via long PASS text in mod_radius module


Castaglia:
When building a RADIUS packet to send to the configured [`RadiusAuthServer`](http://www.proftpd.org/docs/contrib/mod_radius.html#RadiusAuthServer), the mod_radius module copies the password text, provided by the `PASS` FTP command, into a static fixed size buffer for computation.  However, the module does _not_ check whether the provided text will fit into the buffer before doing a `memcpy(3)`, leading to stack overflow for passwords that exceed the buffer size.

Thanks to Critizero Chen for reporting this issue.

Castaglia:
Fixed in master, and backported to the 1.3.9 branch.

PR #2331, opened 2026-09-15
Issue #2330: Check for passwords longer than our fixed-size buffer in…

Castaglia:
… mod_radius, and reject them.


contrib/mod_radius.c
```diff
@@ -270,9 +270,9 @@ static const char *trace_channel = "radius";
 /* Function prototypes. */
 static radius_attrib_t *radius_add_attrib(radius_packet_t *, unsigned char,
   const unsigned char *, size_t);
-static void radius_add_passwd(radius_packet_t *, unsigned char,
+static int radius_add_passwd(radius_packet_t *, unsigned char,
   const unsigned char *, unsigned char *, size_t);
-static void radius_build_packet(radius_packet_t *, const unsigned char *,
+static int radius_build_packet(radius_packet_t *, const unsigned char *,
```
Castaglia, line 275: We change these function signatures so that failures can be reported to the callers.

contrib/mod_radius.c
```diff
@@ -1868,10 +1868,10 @@ static void MD5_Final(unsigned char *, MD5_CTX *);
 
 /* Note: these MD5 routines are taken from RFC 1321 */
 
-#ifdef HAVE_MEMCPY
+#if defined(HAVE_MEMCPY)
```
Castaglia, line 1871: Opportunistic stylistic nit; no functional change.

PR #2335, opened 2026-09-17
The mod_ratio module was not properly requiring that a client be auth…

Castaglia:
…enticated in its command handling authorization checks, leading to extra (and unnecessary) computations being done.


contrib/mod_ratio.c
```diff
@@ -66,7 +66,9 @@
 
 #include "conf.h"
 
-int gotratuser,fileerr;
```
Castaglia, line 69: These variables should be file-scoped only, hence this opportunistic change to a `static int`.

contrib/mod_ratio.c
```diff
@@ -66,7 +66,9 @@
...
+static int ratio_engine = FALSE;
```
Castaglia, line 69: This is mostly a stylistic change, to make this module (not one of my favorites) more consistent, style-wise, with other modules.

contrib/mod_ratio.c
```diff
@@ -968,27 +1031,26 @@ static int ratio_sess_init(void) {
  */
 
 static cmdtable ratio_cmdtab[] = {
-  { PRE_CMD,  C_CWD,	G_NONE, pre_cmd, 	FALSE, FALSE },
-  { CMD,      C_CWD,	G_NONE, cmd_cwd, 	FALSE, FALSE },
-
-  { PRE_CMD,  C_LIST,	G_NONE, pre_cmd, 	FALSE, FALSE },
-  { POST_CMD, C_LIST,	G_NONE, ratio_post_cmd,	FALSE, FALSE },
+  { PRE_CMD,  C_CWD,	G_NONE, pre_cmd, 	TRUE, FALSE },
```
Castaglia, line 1034: These changes to `TRUE`, for the "requires auth?" field in these `cmdtable` rows, are the main change being done.  For well-behaved clients, there should be no net change.  This is a defensive change, to ensure that clients that send these commands prior to authentication do not cause extra work to be done by mod_ratio unnecessarily.

PR #2338, opened 2026-09-18
Fix the session user name tracking in mod_ratio to account for orderi…

Castaglia:
…ng of command handler phases.


contrib/mod_ratio.c
```diff
@@ -572,11 +572,6 @@ MODRET log_pass(cmd_rec *cmd) {
     return PR_DECLINED(cmd);
   }
 
-  if (session.anon_user != NULL) {
-    sstrncpy(g.user, session.anon_user, sizeof(g.user));
-  }
```
Castaglia, line 577: Not sure why this was done for anonymous users, but not regular users.

contrib/mod_ratio.c
```diff
@@ -892,21 +900,6 @@ MODRET post_stor(cmd_rec *cmd) {
...
-  if (ratio_engine == FALSE) {
-    return PR_DECLINED(cmd);
-  }
-
-  /* Make a copy of the provided username, but only once authentication has
-   * completed successfully.
-   */
-  if (!g.user[0]) {
-    sstrncpy(g.user, session.user, sizeof(g.user));
-  }
```
Castaglia, line 905: Now that this is done properly, in `post_cmd()`, and for anonymous users as well, we no longer need this command handler at all.  Nice.

PR #2339, opened 2026-09-18
If the "NoErrorIfAbsent" `ListOption` is configured, an FTP `NLST` co…

Castaglia:
…mmand might receive two 226 response lines, rather than the expected one line.

This does not cause a functional problem for most FTP clients, yet still is surprising and unnecessary behavior.


modules/mod_ls.c
```diff
@@ -3216,7 +3216,6 @@ MODRET ls_nlst(cmd_rec *cmd) {
             }
 
             session.sf_flags |= SF_ASCII_OVERRIDE;
-            pr_response_add(R_226, _("Transfer complete"));
             ls_done(cmd);
```
Castaglia, line 3219: These 226 responses are not necessary because the `ls_done()` function itself will add the same response.  It is this which is causing the unexpectedly doubled "226 Transfer complete" responses.

